PR #19, opened 2012-06-17
Added an extension so that IT Coverage is as modular as Coverage

stephanenicolas:
My goal is to be able to use a different coverage tool (other than the default jacoco) to run it tests.

On android projects, this is nessary : emma is the standard to run tests and it would be nice to be able to specify emma as the IT Coverage plugin. 

This fork makes it possible.
I just copied more or less what was done for normal coverage and added the same processing for IT Coverage.

It now works for me : if I add some 

		<sonar.core.codeItCoveragePlugin>emma</sonar.core.codeItCoveragePlugin>
		<sonar.emma.itReportPath>target/emma/</sonar.emma.itReportPath>

to properties in a maven pom, I can use emma as code coverage tool for integration tests.

It's my first fork, and the first time I work on sonar, please let me know if I can make it better.

Thanks, 
 Stéphane

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,14 @@
-Sonar Core
+Sonar Core for for better integration tests
 ==========
 
-This is the Core codebase of Sonar.
+This fork aims to generalize Coverage and ITCoverage. 
+Up to now Coverage is modular and can be implemented by many plugins like cobertura, clover, emma, jacoco, etc. 
+But IT Coverage can only be achived by Jacoco IT plugin. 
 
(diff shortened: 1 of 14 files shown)


henryju:
Execution of tests is no more supported in SQ. Only reuse report mode (ie parsing of existing reports) is supported and this is something to done by each language plugin. So maybe there is some contribution to be done on Java plugin side to support coverage reports coming from different tools but it should not have any impact on the platform.

PR #34, opened 2014-01-15
Webservice added for creating groups

gadton:
Sync with e.g. LDAP works well for users but not for groups. This service enables an external app to create groups known e.g. from LDAP. => 100% sync with LDAP possible


simonbrandhof:
Thanks for the contribution but web services are no more implemented in RoR but in pure Java. Nevertheless the JIRA ticket was created to track the feature.
https://jira.codehaus.org/browse/SONAR-6067

PR #43, opened 2014-09-22
[SONAR-4983] refactored security realms

stadler:
Hi

I implemented [SONAR-4983] (https://jira.codehaus.org/browse/SONAR-4983) and would be glad if someone of you could have a look at this change and merge it if you think it is fine.

I tested it together with [sonar-preauth] (https://github.com/stadler/sonar-preauth) and [sonar-ldap] (https://github.com/SonarCommunity/sonar-ldap) and it worked as expeced.

This change basically allows you to use different authenticators and user- and group-providers together. 
In my setup for example, I use preauth for authentication and ldap for the user and group details. The different authenticaors and user- and group providers can be defined in the sonar properties by three new properties: 
- `sonar.security.authenticators`: the class names of the classes that will be doing the authentication. E.g. org.sonar.plugins.preauth.PreAuthAuthenticator
- `sonar.security.userProviders`: the class names of the classes that will do the provisioning of the user details. E.g. org.sonar.plugins.ldap.LdapUsersProvider
- `sonar.security.groupProviders`: the class names of the classes that will do the provisioning of the group details. E.g. org.sonar.plugins.ldap.LdapGroupsProvider

Note that if a security realm is defined by using `sonar.security.realm`, then the defined security realm takes precedence.

Thank you for having a look at it and don't hesitate to comment if you have any questions.
   Jacques


stadler:
Hi.
The 5.0 release has passed and this feature wasn't pulled in.
Any plans on when you plan to do so?
Or are there some things missing to do so?
Thanks for the update.

julienlancelot:
Thanks for your work, unfortunately we won't apply it as we have work on another API (https://github.com/SonarSource/sonarqube/blob/master/sonar-plugin-api/src/main/java/org/sonar/api/server/authentication/IdentityProvider.java), which is full Java.
This new API still requires some work in order to be used by LDAP, but we won't maintain the old Authenticator.

PR #48, opened 2014-12-19
Changed all references to docs.codehaus.org to docs.sonarqube.org

gvsmirnov:
It is fairly inconvenient when I an error message that goes like

```
Not supported since v4.2. See http://docs.codehaus.org/display/SONAR/API+Changes
```

But end up on a non-existent page.


simonbrandhof:
Thanks for the contribution. Some links in the PR are not valid and the 4.5 LTS version will be fixed too, so I don't apply the PR as it. We're taking care of this problem.

PR #49, opened 2015-01-20
Auto-assign new issues

simonbrandhof:
@teryk please review this PR for guessing author and assignee of new issues https://jira.codehaus.org/browse/SONAR-5906
https://jira.codehaus.org/browse/SONAR-5960
Thanks


server/sonar-server/src/main/java/org/sonar/server/computation/AnalysisReportService.java
```diff
@@ -130,9 +133,10 @@ private DefaultIssue setFieldDiffs(DefaultIssue issue, String diffFields, Date a
     return issue;
   }
 
-  private DefaultIssue setComponentId(DefaultIssue issue, ReportComponent component) {
+  private DefaultIssue setComponent(DefaultIssue issue, @Nullable ReportComponent component) {
```
teryk: In most of the case, the component should not be null. The only case I can think of is in the case of deleted components. In the issues table, the component_id/component_uuid must not be null. Shouldn't we be more defensive ?

server/sonar-server/src/main/java/org/sonar/server/computation/issue/FinalIssues.java
```diff
@@ -37,4 +38,7 @@ public FinalIssues(TempFolder tempFolder, System2 system2) throws IOException {
     super(tempFolder.newFile("issues", ".dat"), system2);
   }
 
+  FinalIssues(File file, System2 system2) {
```
teryk: not convinced by the FinalIssues class name. Could IssuesCache be a better name ?

server/sonar-server/src/main/java/org/sonar/server/computation/issue/IssueComputation.java
```diff
@@ -51,16 +58,23 @@ public void afterReportProcessing() {
...
+    String scmAccount = issue.authorLogin();
+    if (scmAccount != null) {
+      issue.setAssignee(scmAccountCache.getNullable(scmAccount));
+    }
   }
 
   private void copyRuleTags(DefaultIssue issue) {
-    RuleDto rule = ruleCache.get(issue.ruleKey());
-    issue.setTags(Sets.union(rule.getTags(), rule.getSystemTags()));
+    RuleDto rule = ruleCache.getNullable(issue.ruleKey());
```
teryk: in which case the rule could be null ?

server/sonar-server/src/main/java/org/sonar/server/computation/issue/SourceLinesCache.java
```diff
@@ -0,0 +1,121 @@
...
+import java.io.Reader;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Cache of the lines of the currently processed file. Only a single file
+ * is kept in memory at a time. Moreover data is loaded on demand to avoid
+ * useless db trips.
+ * <p/>
+ * It assumes that db table FILE_SOURCES is up-to-date before using of this
```
teryk: -of —> before using this cache

server/sonar-server/src/main/java/org/sonar/server/util/cache/MemoryCache.java
```diff
@@ -54,12 +56,16 @@ public V getNullable(K key) {
...
     }
     return value;
   }
 
-  public Map<K, V> getAll(Iterable<K> keys) {
+  /**
+   * Get values associated with keys. All the requested keys are included
+   * in the Map result. Value is null if the key is not found in cache.
+   */
+  public Map<K, V> getAllNullable(Iterable<K> keys) {
```
teryk: Nullable means it could return null. getAllFoundInCache/getAll/getAllFound would be alternative candidates. WDYT ?

server/sonar-server/src/test/java/org/sonar/server/computation/issue/IssueComputationTest.java
```diff
@@ -0,0 +1,152 @@
...
+    issue.setAuthorLogin("charlie");
+    when(scmAccountCache.getNullable("charlie")).thenReturn("char.lie");
+
+    process();
+
+    assertThat(issue.assignee()).isNull();
+  }
+
+  private void process() {
+    IssueComputation computation = new IssueComputation(ruleCache, lineCache, scmAccountCache, finalIssues);
```
teryk: IssueComputation is the system under test. Why not put it as a field ?

server/sonar-server/src/main/java/org/sonar/server/computation/AnalysisReportService.java
simonbrandhof: exact. I kept this nullable parameter because it was done before :-) 
I drop it.

server/sonar-server/src/test/java/org/sonar/server/computation/issue/RuleCacheLoaderTest.java
```diff
@@ -0,0 +1,61 @@
...
+  public void load_by_key() throws Exception {
+    dbTester.prepareDbUnit(getClass(), "shared.xml");
+    DbClient dbClient = new DbClient(dbTester.database(), dbTester.myBatis(), new RuleDao());
+    RuleCacheLoader loader = new RuleCacheLoader(dbClient);
+
+    assertThat(loader.load(RuleKey.of("squid", "R001")).getName()).isEqualTo("Rule One");
+    assertThat(loader.load(RuleKey.of("squid", "MISSING"))).isNull();
+  }
+
+  @Test
```
teryk: two solutions: use @Test(expected:UnsupportedOperationException) or use https://github.com/Codearte/catch-exception to keep the arrange/act/assert test pattern

server/sonar-server/src/test/java/org/sonar/server/computation/issue/ScmAccountCacheLoaderTest.java
```diff
@@ -0,0 +1,60 @@
...
+    esTester.putDocuments("users", "user", getClass(), "user1.json");
+
+    UserIndex index = new UserIndex(esTester.client());
+    ScmAccountCacheLoader loader = new ScmAccountCacheLoader(index);
+
+    assertThat(loader.load("missing")).isNull();
+    assertThat(loader.load("jesuis@charlie.com")).isEqualTo("charlie");
+  }
+
+  @Test
```
teryk: two solutions to simplify: use @Test(expected:UnsupportedOperationException) or use https://github.com/Codearte/catch-exception to keep the arrange/act/assert test pattern

server/sonar-server/src/test/java/org/sonar/server/util/cache/MemoryCacheTest.java
```diff
@@ -62,30 +63,32 @@ public void get_throws_exception_if_not_exists() throws Exception {
     try {
       cache.get("not_exists");
       fail();
-    } catch (IllegalArgumentException e) {
+    } catch (NotFoundException e) {
```
teryk: use https://github.com/Codearte/catch-exception ?

sonar-core/src/main/java/org/sonar/core/source/db/FileSourceDao.java
```diff
@@ -47,6 +56,29 @@ public FileSourceDto select(String fileUuid) {
...
+    try {
+      pstmt = connection.prepareStatement("select data from file_sources where file_uuid = ?");
+      pstmt.setString(1, fileUuid);
+      rs = pstmt.executeQuery();
+      if (rs.next()) {
+        reader = rs.getCharacterStream(1);
+        function.apply(reader);
+      }
+    } catch (SQLException e) {
+      e.printStackTrace();
```
teryk: a log or something else should be done here

server/sonar-server/src/main/java/org/sonar/server/computation/issue/FinalIssues.java
simonbrandhof: ok for IssueCache

server/sonar-server/src/main/java/org/sonar/server/computation/issue/IssueComputation.java
simonbrandhof: it was null at a time in tests, but it's not the case anymore. And not a good reason. Fixed.

server/sonar-server/src/main/java/org/sonar/server/util/cache/MemoryCache.java
simonbrandhof: Go for getAll(). It's not only found in cache. Map contains all the keys, even if they are not found.

server/sonar-server/src/test/java/org/sonar/server/computation/issue/IssueComputationTest.java
simonbrandhof: I don't get the value.

server/sonar-server/src/test/java/org/sonar/server/computation/issue/RuleCacheLoaderTest.java
simonbrandhof: I don't like @Test(expected=xxx) as it does not check the initial location of exception.

sonar-core/src/main/java/org/sonar/core/source/db/FileSourceDao.java
simonbrandhof: good catch. How is it possible that I did that ??

server/sonar-server/src/test/java/org/sonar/server/computation/issue/IssueComputationTest.java
teryk: same pattern as most tests, easy to see where the IssueComputation is instanciated. process should just process

teryk:
LGTM !

PR #50, opened 2015-01-20
Improve performance of ES request on users

simonbrandhof:
Improve performance of ES request on users

Searching for users by SCM account does not require scoring, so filters must be used instead of queries.


simonbrandhof:
@julienlancelot could you please review this PR ? Thanks

PR #51, opened 2015-01-21
SONAR-5960 ES request fails if author is empty ("")

simonbrandhof:
SONAR-5960 ES request fails if author is empty ("")

 - do not allow blank author on issues
 - fix ES search

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/issue/SourceLinesCache.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/issue/SourceLinesCache.java
@@ -20,6 +20,7 @@
 package org.sonar.server.computation.issue;
 
 import com.google.common.base.Function;
+import com.google.common.base.Strings;
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.CSVParser;
 import org.apache.commons.csv.CSVRecord;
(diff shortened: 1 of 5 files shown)


server/sonar-server/src/main/java/org/sonar/server/computation/issue/SourceLinesCache.java
```diff
@@ -69,7 +70,8 @@ void init(String fileUuid) {
   public String lineAuthor(int lineId) {
     loadIfNeeded();
     if (lineId <= authors.size()) {
-      return authors.get(lineId - 1);
+      String author = authors.get(lineId - 1);
```
teryk: Why go the null way when we can return an empty string ? No NPE, we can do a var.isEmpty(), lass things to worry about.

simonbrandhof: We should go deeper in this discussion with the whole team. Currently null is used in most of our models for unset fields. Empty string currently means "information exists and value is blank". That's not the case here. Issue has no author.
Changing that would require to update the whole stack.

teryk: Fine. LGTM !

PR #52, opened 2015-01-21
issues - add component and project uuids to issues table

teryk:
two new columns: component_uuid and project_uuid
the PR contains migration scripts and insert


teryk:
@simonbrandhof to review
component_uuid and project_uuid are now filled at insertion (not possible to update this columns)

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v51/FeedIssueComponentUuids.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Override
+  public void execute(Context context) throws SQLException {
+    MassUpdate update = context.prepareMassUpdate().rowPluralName("issues");
+    update.select(
+      "SELECT c.uuid, c.project_uuid, i.id " +
+        "FROM issues i " +
+        "INNER JOIN projects c ON i.component_id=c.id " +
+        "WHERE i.component_uuid is null");
```
simonbrandhof: There are two strategies when optimizing reentrancy of migration:
- restrict input to non-updated rows in SQL request -> "where i.component_uuid is null"
- full scan table and post-filter in Java (remove "where i.component_uuid is null")

The first option can be faster if there's an index on the column, else execution time can be much slower than full scan. On the other hand, volume of response on the db wire is still smaller with first option.

All that to say: there's an index on ISSUES.COMPONENT_ID, so it sounds good !
But... this index is missing on the new column ISSUES.COMPONENT_UUID. Is it required regarding to existing SQL requests ?

simonbrandhof:
About IssueMapper.xml:
<sql id="issueColumns"> of must be updated in order to replace:
- root.uuid with i.project_uuid
- p.uuid with i.component_uuid

Requests "update" and "updateIfBeforeSelectedDate" must set project_uuid.

SELECT requests must use i.component_uuid and i.project_uuid.

simonbrandhof:
else LGTM

simonbrandhof:
LGTM

teryk:
merge manually

PR #54, opened 2015-01-23
SONAR-5906 log warning when multiple users share the same SCM account

simonbrandhof:
SONAR-5906 log warning when multiple users share the same SCM account, so that issue can't be assigned to a single person.


server/sonar-server/src/main/java/org/sonar/server/computation/issue/ScmAccountCacheLoader.java
```diff
@@ -19,28 +19,53 @@
...
+      return users.get(0).login();
+    }
+    Collection<String> logins = Collections2.transform(users, new Function<UserDoc, String>() {
+      @Override
+      public String apply(UserDoc input) {
+        return input.login();
+      }
+    });
+    log.warn(String.format("Multiple users share the SCM account '%s': %s", scmAccount, Joiner.on(", ").join(logins)));
+    return null;
```
teryk: we should not return null if there's multiple login, should we ? what is the points of the logins Collection then ?

server/sonar-server/src/main/java/org/sonar/server/user/index/UserIndex.java
```diff
@@ -80,4 +83,26 @@ public UserDoc getByLogin(String login) {
     return userDoc;
   }
 
+  /**
+   * Returns the users (at most 3) who are associated to the given SCM account. This method can be used
+   * to detect user conflicts. It never returns null.
```
teryk: don't need to indicate that it never returns null, the @CheckForNull is done for that. More generally, your method name is clear enough and I would not use any comment. They are useful in the case of a public API.

server/sonar-server/src/main/java/org/sonar/server/computation/issue/ScmAccountCacheLoader.java
simonbrandhof: this is the CacheLoader, not the Cache. It's called once for a given key.

server/sonar-server/src/main/java/org/sonar/server/computation/issue/ScmAccountCacheLoader.java
```diff
@@ -19,28 +19,53 @@
...
+public class ScmAccountCacheLoader implements CacheLoader<String, String> {
 
+  private final Logger log;
   private final UserIndex index;
 
   public ScmAccountCacheLoader(UserIndex index) {
+    this(index, LoggerFactory.getLogger(ScmAccountCacheLoader.class));
+  }
+
+  public ScmAccountCacheLoader(UserIndex index, Logger log) {
```
teryk: the visibility should be package-protected and add the annotation that it's for testing purpose

server/sonar-server/src/main/java/org/sonar/server/computation/issue/ScmAccountCacheLoader.java
teryk: I've seen your test, so it should return null. In that case, delete the logins collection. You could simplify checking if the size if 1, otherwise returning null, and a warning if size > 1

server/sonar-server/src/test/java/org/sonar/server/user/index/UserIndexTest.java
```diff
@@ -112,4 +112,17 @@ public void get_nullable_by_scm_account_return_null_when_two_users_have_same_ema
     assertThat(index.getNullableByScmAccount("user1@mail.com")).isNull();
   }
 
+  @Test
```
teryk: should you test when there's more than 3 accounts, that it returns only three accounts ?

teryk:
LGTM @simonbrandhof

PR #55, opened 2015-01-23
Keep compatibility with issue-assign and cockpit plugins

simonbrandhof:
Keep compatibility with issue-assign and cockpit plugins

Do not override author and assignee fields of new issues


teryk:
LGTM

PR #56, opened 2015-01-26
5912 delete component ids from the issues table

teryk:
all tests pass, IT concerning purge pass.


julienlancelot:
Just a simple comment, "SONAR-" is missing at the begin of your commit, should be "SONAR-5912 delete..."

teryk:
yep, I should not put a jira number in fact, as it does not concern a jira ticket directly.

sonar-core/src/main/resources/org/sonar/core/persistence/schema-h2.ddl
```diff
@@ -442,9 +442,7 @@ CREATE TABLE "GRAPHS" (
 CREATE TABLE "ISSUES" (
   "ID" BIGINT NOT NULL GENERATED BY DEFAULT AS IDENTITY (START WITH 1, INCREMENT BY 1),
   "KEE" VARCHAR(50) UNIQUE NOT NULL,
-  "COMPONENT_ID" INTEGER NOT NULL,
-  "COMPONENT_UUID" VARCHAR(50),
-  "ROOT_COMPONENT_ID" INTEGER,
+  "COMPONENT_UUID" VARCHAR(50) NOT NULL,
   "PROJECT_UUID" VARCHAR(50),
```
simonbrandhof: PROJECT_UUID must be NOT NULL too

simonbrandhof:
In IssueMapper.xml, don't you think that the SELECT statements should return i.component_uuid and i.project_uuid instead of the joined columns p.uuid and root.uuid ? (see <sql id="issueColumns">). I hope these joins on PROJECTS will be dropped soon, so using ISSUES columns is better when possible.

simonbrandhof:
LGTM if my 2 comments are applied.

PR #57, opened 2015-01-26
SONAR-6096 When a new issue is at file level, automatically assign it to the latest committer

simonbrandhof:
SONAR-6096 When a new issue is at file level, automatically assign it to the latest committer


teryk:
LGTM

PR #58, opened 2015-01-28
Use protocol buffers for format of batch output report 

simonbrandhof:


dbolkensteyn:
Please organize a BBL to present this @simonbrandhof :)

simonbrandhof:
@dbolkensteyn this is already planned !

server/sonar-server/src/main/java/org/sonar/server/computation/ComputationContext.java
```diff
@@ -20,31 +20,27 @@
...
-  private Date analysisDate;
+  private final BatchOutputReader reportReader;
+
+  /**
+   * Cache of analysis date as it can be accessed several times
+   */
+  private Date analysisDate = null;
 
-  public ComputationContext(AnalysisReportDto reportDto, ComponentDto project, File reportDir) {
+  public ComputationContext(AnalysisReportDto reportDto, ComponentDto project, BatchOutputReader reportReader) {
```
teryk: why call it BatchOutputReader ? Shouldn't it be better to call it ReportReader or AnalysisReportReader ? It's not important it's from the batch, what is important is that it reads a report.

simonbrandhof: Because protocol buffers will also be applied to batch input: BatchInputReader, BatchInputWriter

teryk: how could it be in json ? Are we talking about another batch report than the computation one ?
I still think that ReportOutputReader or AnalysisReportOutputReader are better names. I don't think it's useful to tell it comes from the batch. WDYT ?

teryk:
LGTM

PR #59, opened 2015-01-28
SONAR-6087 Purge removed Views and Sub-Views Index on each Views analysis

julienlancelot:
SONAR-6087 Purge removed Views and Sub-Views Index on each Views analysis


server/sonar-server/src/main/java/org/sonar/server/computation/step/PurgeRemovedViewsStep.java
```diff
@@ -0,0 +1,85 @@
...
+
+import javax.annotation.Nullable;
+
+import java.util.List;
+import java.util.Set;
+
+import static com.google.common.collect.Sets.newHashSet;
+
+/**
+ * This step will remove every Views and Sub Views from the index that do not exists in the db.
```
simonbrandhof: I would say: "This step removes from index the views and sub-views that do not exist in db."

sonar-core/src/main/resources/org/sonar/core/component/db/ComponentMapper.xml
```diff
@@ -156,7 +156,7 @@
     </where>
   </select>
 
-  <select id="selectAllViewsAndSubViews" resultType="hashmap">
+  <select id="selectAllViewsAndSubViews" resultType="UuidWithProjectUuid">
```
simonbrandhof: The method name "selectAllViewsAndSubViews" is clear, it loads views, so it should not accept qualifier parameters. Two options:
- hardcode values of qualifiers in SQL request
- declare a more generic method like selectUuidsForQualifiers(String... qualifiers)

server/sonar-server/src/test/java/org/sonar/server/computation/step/PurgeRemovedViewsStepTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+    ComponentDto view = ComponentTesting.newProjectDto("ABCD").setQualifier(Qualifiers.VIEW);
+    ComponentDto subView = ComponentTesting.newModuleDto(view, "BCDE").setQualifier(Qualifiers.SUBVIEW);
+    dbClient.componentDao().insert(session, view, subView);
+    session.commit();
+
+    step.execute(context);
+
+    List<SearchHit> results = esTester.getDocuments(ViewIndexDefinition.INDEX, ViewIndexDefinition.TYPE_VIEW);
+    List<String> viewUuids = newArrayList(Iterables.transform(results, new Function<SearchHit, String>() {
```
simonbrandhof: This function is helpful and must be moved to EsTester:
```java
public <T> Iterable<T> getDocumentFields(String indexName, String typeName, final String fieldName) {
    return Iterables.transform(getDocuments(indexName, typeName), new Function<SearchHit, T>() {
      @Override
      public T apply(SearchHit input) {
        return (T)input.sourceAsMap().get(fieldName);
      }
    });
  }
```

Test would be simply: 
```java
Iterable<String> viewUuids = esTester.getDocumentFields(ViewIndexDefinition.INDEX, ViewIndexDefinition.TYPE_VIEW, ViewIndexDefinition.FIELD_UUID);
assertThat(viewUuids).containsOnly("ABCD", "BCDE");
```

sonar-core/src/main/resources/org/sonar/core/component/db/ComponentMapper.xml
julienlancelot: Indeed, the generic option seems good to me. 

By the way, searching in the table projects with only filters on qualifiers should require an index on the projects.qualifier column, don't you think (There's an index on snapshots.qualifier) ?

julienlancelot: In fact no need to add an index on projects.qualifier : as the goal is to remove all views from the index that not exists in db, it would be better to search views from the list of view uuid found in the index.

sonar-core/src/main/resources/org/sonar/core/component/db/ComponentMapper.xml
```diff
@@ -156,7 +156,7 @@
...
     SELECT v.uuid as "uuid", v.project_uuid as "projectUuid" FROM projects v
     <where>
       v.enabled=${_true} AND v.qualifier=#{viewQualifier} OR v.qualifier=#{subViewQualifier}
```
simonbrandhof: Ambiguity with precedence of OR/AND. I would put OR condition into parenthesis.

julienlancelot: Indeed !

simonbrandhof:
LGTM

PR #60, opened 2015-01-29
SONAR-6113-notifications

simonbrandhof:
Optimization - do not process issue notifications if there aren't any subscribers


server/sonar-server/src/main/java/org/sonar/server/platform/ServerComponents.java
```diff
@@ -680,7 +679,6 @@ void startLevel4Components(ComponentContainer pico) {
     pico.addSingleton(IssueService.class);
     pico.addSingleton(IssueActionsWriter.class);
     pico.addSingleton(IssueQueryService.class);
-    pico.addSingleton(IssueNotifications.class);
```
teryk: the notifications are used only in the context of computation stack ? Otherwise, I don't understand why you remove this component

simonbrandhof: This class was indeed in org.sonar.server.computation.issue -> specific to computation

server/sonar-server/src/test/java/org/sonar/server/issue/notification/IssueNotificationsTest.java
```diff
@@ -17,133 +17,152 @@
...
-      .setMessage("the message")
-      .setKey("ABCDE")
-      .setComponentKey("struts:Action")
-      .setProjectKey("struts");
-    issueNotifications.sendChanges(issue, "charlie", null, new Project("struts"), null, null, false);
-
-    Mockito.verifyZeroInteractions(manager);
-  }
-}
+///*
```
teryk: no need for tests ? what is its current responsability ?

teryk:
LGTM

PR #61, opened 2015-01-29
SONAR-6089 Allow to search Issues by Views / Sub-Views

julienlancelot:


jblievremont:
I would rebase  f1c1999 and 7b19e10 as fixups into 5d2b298, else LGTM!

PR #62, opened 2015-01-30
SONAR-5953 Fall back to backwards compatible API for issues search WS

jblievremont:
Usage of parameters components and componentRoot(Uuid)s must be kept compatible with previous versions.

OTOH, componentKeys, componentUuids and onComponentOnly are introduced for cleaner control over behavior and scope of the query.


jblievremont:
Addressed through refactoring of IssueQueryService

PR #67, opened 2015-02-03
Snapshot migration from timestamp to long

teryk:
ci on every database ok
migration ok for postgre sql
it compute ok


server/sonar-server/src/main/java/org/sonar/server/db/migrations/v51/FeedSnapshotsLongDates.java
```diff
@@ -0,0 +1,58 @@
...
+  public FeedSnapshotsLongDates(Database db) {
+    super(db);
+  }
+
+  @Override
+  public void execute(Context context) throws SQLException {
+    MassUpdate massUpdate = context.prepareMassUpdate();
+    massUpdate.select("SELECT s.created_at, s.build_date, s.period1_date, s.period2_date, s.period3_date, s.period4_date, s.period5_date, s.id FROM snapshots s WHERE created_at_ms IS NULL");
+    massUpdate.update("UPDATE snapshots SET created_at_ms=?, build_date_ms=?, period1_date_ms=?, period2_date_ms=?, period3_date_ms=?, period4_date_ms=?, period5_date_ms=? WHERE id=?");
+    massUpdate.rowPluralName("snapshosts");
```
simonbrandhof: Typo on "snapshots" (what a valuable feedback !)

sonar-plugin-api/src/main/java/org/sonar/api/utils/DateUtils.java
```diff
@@ -60,10 +60,16 @@ public static String formatDateTimeNullSafe(@Nullable Date date) {
...
+  @CheckForNull
   public static Date timeToDate(@Nullable Long time) {
     return time == null ? null : new Date(time);
   }
 
+  @CheckForNull
+  public static Long dateToTime(@Nullable Date date) {
+    return date == null ? null : date.getTime();
+  }
+
```
simonbrandhof: IMO method names are misleading because "time" can also be associated to java objects (java.util.Date, ...). I would name them "dateToLong" and "longToDate"

plugins/sonar-core-plugin/src/main/resources/org/sonar/plugins/core/widgets/timeline.html.erb
```diff
@@ -53,9 +53,7 @@
          metric_data_map[metric_id].each() do |metric_data|
            # for every metric value, we need to check that the corresponding snapshot has values for each metric
            if metric_count_per_snapshot_id[metric_data[:sid]]==total_number_of_metrics
-             m_date = metric_data[:date]
-             # Only Oracle returns a Time object, so let's parse this string if it's not a Time instance
-             m_date = Time.parse(metric_data[:date]) unless m_date.is_a? Time
```
julienlancelot: metric_data[:date] is a long because metric_data_map is populated by TrendsChart.time_machine_measures (where we can see that the date is coming from snapshots.created_at).
When snapshots.created_at was a DateTime, Oracle was returning a Time object instead of a Date object, but as now this column is a long I think this 'hack' is useless.

sonar-core/src/test/java/org/sonar/core/timemachine/PeriodsTest.java
```diff
@@ -63,7 +63,7 @@ public void before() {
   @Test
   public void label_of_duration_in_days() {
     when(snapshot.getPeriodMode(periodIndex)).thenReturn(CoreProperties.TIMEMACHINE_MODE_DAYS);
-    when(snapshot.getPeriodDate(periodIndex)).thenReturn(new Date());
+    when(snapshot.getPeriodDate(periodIndex)).thenReturn(new Date().getTime());
```
simonbrandhof: new Date().getTime() is System.currentTimeMillis()

plugins/sonar-core-plugin/src/main/resources/org/sonar/plugins/core/widgets/timeline.html.erb
simonbrandhof: Ok thanks @julienlancelot

server/sonar-web/src/main/webapp/WEB-INF/app/models/event.rb
```diff
@@ -63,6 +63,6 @@ def self.already_exists(snapshot_id, event_name, event_category)
   # TODO: Remove this code when everything has been checked on the Event handling, both on the UI and the WS API
   #
   def populate_snapshot
-    self.snapshot=Snapshot.snapshot_by_date(resource_id, event_date) unless self.snapshot
+    self.snapshot=Snapshot.snapshot_by_date(resource_id, event_date.to_i*1000) unless self.snapshot
```
simonbrandhof: Method snapshot_by_date still requires a DateTime, so this change must be reverted.

sonar-batch/src/main/java/org/sonar/batch/ProjectConfigurator.java
```diff
@@ -94,11 +95,11 @@ private void checkCurrentAnalysisIsTheLatestOne(String projectKey, Date analysis
       ResourceModel persistedProject = databaseSession.getSingleResult(ResourceModel.class, "key", projectKey, "enabled", true);
       if (persistedProject != null) {
         Snapshot lastSnapshot = databaseSession.getSingleResult(Snapshot.class, "resourceId", persistedProject.getId(), "last", true);
-        if (lastSnapshot != null && !lastSnapshot.getCreatedAt().before(analysisDate)) {
+        if (lastSnapshot != null && !timeToDate(lastSnapshot.getCreatedAt()).before(analysisDate)) {
```
simonbrandhof: instead of building a new Date object, you can compare longs:
lastSnapshot.getCreatedAt()>=analysisDate.getTime()

server/sonar-web/src/main/webapp/WEB-INF/app/models/snapshot.rb
```diff
@@ -36,6 +36,39 @@ class Snapshot < ActiveRecord::Base
 
   STATUS_UNPROCESSED = 'U'
   STATUS_PROCESSED = 'P'
+  
+  def created_at
+    long_to_date(:created_at)
+  end
+  
+  def build_date
+    long_to_date(:buil_date)
```
simonbrandhof: Error, bad field name: long_to_date(:build_date)

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v51/FeedSnapshotsLongDates.java
teryk: what is the point of this rowPluralName ? I update id

simonbrandhof: this is the label used in logs when displaying progress

teryk:
I will revert all the files that have the EOF modification when I'll merge

PR #68, opened 2015-02-03
SONAR-5946 SONAR-6027 better handling of ajax errors

stas-vilchik:


jblievremont:
Authentication/Authorization errors should be handled separately, but as a first simplifying step, LGTM!

PR #69, opened 2015-02-03
Stop storing distribution of issue-related measures by rule

julienlancelot:
SONAR-6044 Stop storing distribution of issue-related measures by rule


julienlancelot:
@teryk I've applied your feedback, please review it !

teryk:
LGTM

PR #70, opened 2015-02-03
SONAR-6137 Show distribution of issues by creation date

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/search/Result.java
```diff
@@ -75,31 +72,52 @@ public Result(@Nullable BaseIndex<K, ?, ?> index, SearchResponse response) {
 
   private void processAggregation(Aggregation aggregation) {
```
julienlancelot: I'm not sure that all aggregation stuff should be put in this class : why not having a kind of helper class that will take a SearchResponse in parameter and add process aggregation on it ?

julienlancelot:
I think it would be better for P/R to be created from a branch in the SonarQube git repo, it will be more easy to test it locally

julienlancelot:
It's not related to this P/R, but IssueQueryService become too complex : if nothing is done is will become very hard to maintain. And its unit tests are covering only a small part of it (it's tested by MT, but we should not count only on it).

server/sonar-server/src/main/java/org/sonar/server/search/Result.java
julienlancelot: And separating aggregation stuff will allow to have Unit Test on it, as for the moment there's not UT on Result.

server/sonar-server/src/test/java/org/sonar/server/issue/index/IssueIndexMediumTest.java
```diff
@@ -645,6 +652,76 @@ public void filter_by_created_at() throws Exception {
   }
 
   @Test
+  public void facet_on_created_at() throws Exception {
```
julienlancelot: Different test cases should be created, instead of testing all use cases in a single test

server/sonar-server/src/main/java/org/sonar/server/search/Result.java
jblievremont: I'm not fond of creating helper classes that manipulate another class's internals. What could be done is to define a Facets class, embedded into Result<?>, which would be responsible for processing ES Aggregations and holding facet values.

WDYT ?

julienlancelot: In fact your right, I thought SearchResponse could be used externally of the Result class.
Ok for an internal class, and maybe we should think of another name for this classe, because it's main purpose is to decorate facets on a search query.

julienlancelot:
LGTM

PR #76, opened 2015-02-07
SONAR-5960 remove CSV from db USERS.SCM_ACCOUNTS

simonbrandhof:
Simplify format of USERS.SCM_ACCOUNTS by replacing CVS by pipe-separated values

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/db/migrations/v51/CopyScmAccountsFromAuthorsToUsers.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/db/migrations/v51/CopyScmAccountsFromAuthorsToUsers.java
@@ -25,7 +25,6 @@ import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
 import org.sonar.api.utils.System2;
 import org.sonar.core.persistence.Database;
-import org.sonar.core.user.UserDto;
 import org.sonar.server.db.migrations.BaseDataChange;
 import org.sonar.server.db.migrations.Select;
 import org.sonar.server.db.migrations.Upsert;
@@ -42,6 +41,8 @@ import java.util.concurrent.atomic.AtomicLong;
 import static com.google.common.collect.Lists.newArrayList;
 
 public class CopyScmAccountsFromAuthorsToUsers extends BaseDataChange {
+
+  private static final char SCM_ACCOUNTS_SEPARATOR = '\n';
 
   private final System2 system;
   private final AtomicLong counter = new AtomicLong(0L);
@@ -136,20 +137,20 @@ public class CopyScmAccountsFromAuthorsToUsers extends BaseDataChange {
     if (scmAccounts.isEmpty()) {
       return null;
     }
-    return UserDto.SCM_ACCOUNTS_SEPARATOR + Joiner.on(UserDto.SCM_ACCOUNTS_SEPARATOR).join(scmAccounts) + UserDto.SCM_ACCOUNTS_SEPARATOR;
+    return SCM_ACCOUNTS_SEPARATOR + Joiner.on(SCM_ACCOUNTS_SEPARATOR).join(scmAccounts) + SCM_ACCOUNTS_SEPARATOR;
   }
 
   private static class User {
     Long id;
     String login;
     String email;
-    String scmAcounts;
+    String scmAccounts;
 
-    User(Long id, String login, String email, String scmAcounts) {
+    User(Long id, String login, String email, String scmAccounts) {
       this.id = id;
       this.login = login;
       this.email = email;
-      this.scmAcounts = scmAcounts;
+      this.scmAccounts = scmAccounts;
     }
   }
 }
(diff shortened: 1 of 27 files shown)


server/sonar-server/src/main/java/org/sonar/server/user/UserUpdater.java
```diff
@@ -151,7 +151,7 @@ private UserDto createNewUserDto(DbSession dbSession, NewUser newUser) {
     List<String> scmAccounts = sanitizeScmAccounts(newUser.scmAccounts());
     if (scmAccounts != null && !scmAccounts.isEmpty()) {
       validateScmAccounts(dbSession, scmAccounts, login, email, null, messages);
-      userDto.setScmAccounts(convertScmAccountsToCsv(scmAccounts));
+      userDto.setScmAccounts(scmAccounts);
```
julienlancelot: convertScmAccountsToCsv should be removed as it's now useless

sonar-core/src/main/java/org/sonar/core/user/UserDto.java
```diff
@@ -88,11 +97,39 @@ public String getScmAccounts() {
...
-  public UserDto setScmAccounts(@Nullable String scmAccounts) {
-    this.scmAccounts = scmAccounts;
+  public List<String> getScmAccountsAsList() {
+    return decodeScmAccounts(scmAccounts);
+  }
+
+  /**
+   * List of SCM accounts separated by '|'
+   */
+  public UserDto setScmAccounts(@Nullable String s) {
```
julienlancelot: I think this method should not be public. It's only used to test encodeScmAccounts(), in other use case setScmAccounts(@Nullable List list) should be used.

sonar-core/src/main/java/org/sonar/core/user/UserDto.java
```diff
@@ -88,11 +97,39 @@ public String getScmAccounts() {
     return scmAccounts;
   }
 
-  public UserDto setScmAccounts(@Nullable String scmAccounts) {
-    this.scmAccounts = scmAccounts;
+  public List<String> getScmAccountsAsList() {
```
julienlancelot: This method should replace getScmAccounts(), so getScmAccounts() should not be public as it's only used to test encodeScmAccounts().

sonar-core/src/main/java/org/sonar/core/user/UserDto.java
```diff
@@ -19,13 +19,22 @@
...
 import javax.annotation.Nullable;
 
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * @since 3.2
  */
 public class UserDto {
+  private static final char SCM_ACCOUNTS_SEPARATOR = '|';
```
julienlancelot: As discussed, it should be replaced by \n

sonar-core/src/main/java/org/sonar/core/user/UserDto.java
simonbrandhof: Not possible because of mybatis.

julienlancelot: It seems that Mybatis do not need getters / setters : see "... MyBatis also supports private properties and private JavaBeans properties..." from http://mybatis.github.io/mybatis-3/sqlmap-xml.html

simonbrandhof: Ok, even if you're right about MyBatis, I think that the raw data should always be accessible from a DTO.

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v51/CopyScmAccountsFromAuthorsToUsers.java
```diff
@@ -130,19 +131,12 @@ public void handle(Select.Row row) throws SQLException {
...
-    csv.values(list.toArray(new String[size]));
-    csv.close();
-    // Remove useless line break added by CsvWriter at this end of the text
-    return CharMatcher.JAVA_ISO_CONTROL.removeFrom(writer.toString());
+  @CheckForNull
+  private static String encodeScmAccounts(List<String> scmAccounts) {
+    if (scmAccounts.isEmpty()) {
+      return null;
+    }
+    return UserDto.SCM_ACCOUNTS_SEPARATOR + Joiner.on(UserDto.SCM_ACCOUNTS_SEPARATOR).join(scmAccounts) + UserDto.SCM_ACCOUNTS_SEPARATOR;
```
julienlancelot: Are you sure it's a good idea to use a constant from UserDto ? If tomorrow we change the format and we update this constant, this migration won't work anymore.

sonar-core/src/main/java/org/sonar/core/user/UserDto.java
julienlancelot: Indeed, in such case maybe we should call this methods getRawXXX and setRawXXX ? It would make it more easy to understand.

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v51/CopyScmAccountsFromAuthorsToUsers.java
simonbrandhof: Correct. It's fixed

julienlancelot:
LGTM

PR #78, opened 2015-02-10
SONAR-5183 timezones - semaphores migration

teryk:


server/sonar-server/src/test/java/org/sonar/server/db/migrations/v51/FeedSemaphoresLongDatesTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+  @Test
+  public void take_now_if_date_in_the_future() throws Exception {
+    DatabaseMigration migration = newMigration(mock(System2.class));
+
+    migration.execute();
+
+    int count = db
+      .countSql("select count(*) from semaphores where " +
+        "created_at_ms = 0");
```
simonbrandhof: Using 0 for now() is misleading. It could be a bug for example if System2#now() is not called. I would mock System2#now() and return an hardcoded value.

simonbrandhof:
LGTM

PR #80, opened 2015-02-10
Feature/refactoring on IssueQueryService

jblievremont:
- First commit gets rid of the "contextualized" boolean, which only introduced complexity that can be handled on UI side
- Second commit moves some code to make IssueQueryService a bit more readable and testable

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/issue/IssueQueryService.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/issue/IssueQueryService.java
@@ -204,16 +204,12 @@ public class IssueQueryService implements ServerComponent {
     builder.onComponentOnly(effectiveOnComponentOnly);
 
     if (allComponentUuids.isEmpty()) {
-      builder.setContextualized(false);
       addComponentsBelowView(builder, session, authors, projects, projectUuids, moduleUuids, directories, fileUuids);
     } else {
       if (effectiveOnComponentOnly) {
-        builder.setContextualized(false);
         builder.componentUuids(allComponentUuids);
         return;
       }
-
-      builder.setContextualized(true);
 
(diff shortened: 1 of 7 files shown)


server/sonar-server/src/main/java/org/sonar/server/issue/IssueQueryService.java
```diff
@@ -174,127 +183,27 @@ public IssueQuery createFromRequest(Request request) {
     }
   }
 
-  private void addComponentParameters(IssueQuery.Builder builder, DbSession session,
-    @Nullable Boolean onComponentOnly,
+  private boolean mergeComponentParameters(DbSession session, Boolean onComponentOnly,
```
julienlancelot: As the goal of this method is to manage backward compatibility, please remove its name to make it clear what is the goal of this method

server/sonar-server/src/main/java/org/sonar/server/issue/IssueQueryService.java
```diff
@@ -174,127 +183,27 @@ public IssueQuery createFromRequest(Request request) {
...
-  private boolean mergeComponentParameters(DbSession session, Boolean onComponentOnly,
-    Collection<String> components,
-    Collection<String> componentUuids,
-    Collection<String> componentKeys,
-    Collection<String> componentRootUuids,
-    Collection<String> componentRoots,
     Set<String> allComponentUuids) {
     boolean effectiveOnComponentOnly = false;
 
+    failOnIncompatibleParameters(components, componentUuids, componentKeys, componentRootUuids, componentRoots);
```
julienlancelot: This method should be removed : it adds 10 lines, but it can be replaced by only 3 lines...

julienlancelot:
After applying my 2 feedbacks, LGTM.

PR #81, opened 2015-02-10
SONAR-6156 The search of projects on contextualized Issues page should support views

julienlancelot:


sonar-core/src/main/resources/org/sonar/core/user/AuthorizationMapper.xml
```diff
@@ -30,6 +30,33 @@
         <foreach collection="componentKeys" open="(" close=")" item="element" index="index" separator=" or " >p.kee=#{element}</foreach>
   </select>
 
+  <select id="keepAuthorizedProjectIdsForUser" parameterType="map" resultType="long">
```
simonbrandhof: it's a pity that these sql requests must be duplicated but ok.

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchAction.java
```diff
@@ -0,0 +1,135 @@
...
+
+  private final DbClient dbClient;
+
+  public SearchAction(DbClient dbClient) {
+    this.dbClient = dbClient;
+  }
+
+  void define(WebService.NewController controller) {
+    WebService.NewAction action = controller.createAction("search")
+      .setDescription("Search for projects that are containing in a view or a sub view")
```
simonbrandhof: Should be renamed: "Search for components. Currently limited to projects in a view or a sub-view"

simonbrandhof:
LGTM

PR #82, opened 2015-02-10
SONAR-5183 timezones - project_measures and manual_measures migrations

teryk:
the graphs table hasn't been migrated as the timestamp columns are not used at all. At insertion, they are set at current_timestamp.
I haven't found any other side effects than the ones listed here.


simonbrandhof:
LGTM

PR #89, opened 2015-02-14
Speed-up build

simonbrandhof:
Benchmarks on my box:

1. BUILD WITH TESTS 
mvn clean install
shortcut: build.sh
08:33 min

2. PARALLEL BUILD WITH PARALLEL TESTS, NO JS TESTS
Note: JS tests are disabled because of not stable
mvn clean install -T0.5C -DforkCount=0.5C -DskipWebTests=true
shortcut: quick-build.sh -DskipWebTests=true
03:07 min
The hotspots are modules server (1:56 min), web (49 sec) and batch (34 sec)

3. BUILD WITHOUT TESTS
mvn clean install -DskipTests
01:26 min

4. PARALLEL BUILD WITHOUT TESTS
mvn clean install -T0.5C -DforkCount=0.5C -DskipTests=true
shortcut: unsafe-build.sh
50 sec


quick-build.sh
```diff
@@ -1,41 +1,6 @@
...
-echo '               ARE U CRAZY ?'
-echo ''
-echo ''
-echo ''
-
-mvn clean install -e -DskipTests -Pdev $*
+# Parallel executions of maven modules and tests.
+# Half of CPU core are used in to keep other half for OS and other programs.
+# As long as web JS tests are not stable, it's recommended to execute: quick-build.sh -DskipWebTests=true
+mvn clean install -T0.5C -DforkCount=0.5C $*
```
bellingard: @simonbrandhof Is that expected to not have "-DskipTests -Pdev" anymore? If yes, then this is no more a "quick build".

simonbrandhof: quick-build.sh was unsafe-build.sh.
But in fact a build should always be fast, so I renamed quick-build.sh to build.sh and dropped unsafe-build.sh

henryju:
LGTM

PR #90, opened 2015-02-16
Enhance profiling of "delete by query" and "bulk" ES requests

jblievremont:


julienlancelot:
After applying one feedback, LGTM

PR #91, opened 2015-02-16
feature/logging

simonbrandhof:
SONAR-5700 new logging API. Enjoy org.sonar.api.utils.log.Logger and LogTester

SONAR-6057 customization of log rotation, including sonar.log and access.log

Note: still needs to sanitize logs in computation stack


simonbrandhof:
Note: a lot of files have been touched because of removal of SLF4J loggers.

The interesting classes are:
- org.sonar.api.utils.log.*
- org.sonar.process.LogbackHelper
- org.sonar.application.AppLogging
- org.sonar.search.SearchLogging
- org.sonar.server.app.WebLogging and TomcatAccessLog

sonar-plugin-api/src/main/java/org/sonar/api/utils/log/LogInterceptor.java
```diff
@@ -0,0 +1,35 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.api.utils.log;
+
+abstract class LogInterceptor {
```
henryju: If my understanding is correct this is the way to intercept logs in case SQ is embedded (ie SQ Runner API). If true then I think it would be better to distinguish log levels. For example in Eclipse we can configure the console to focus to front in case of error message.

sonar-plugin-api/src/test/java/org/sonar/api/utils/log/NullInterceptorTest.java
```diff
@@ -0,0 +1,37 @@
...
+package org.sonar.api.utils.log;
+
+import org.junit.Test;
+
+import static org.mockito.Mockito.mock;
+
+public class NullInterceptorTest {
+
+  @Test
+  public void do_nothing() throws Exception {
```
teryk: how does this test checks that it does nothing ?

sonar-plugin-api/src/main/java/org/sonar/api/utils/log/LogTester.java
```diff
@@ -0,0 +1,100 @@
...
+
+import java.util.List;
+
+/**
+ * <b>For tests only</b>
+ * <p/>
+ * This JUnit rule allows to configure and access logs in tests. By default
+ * debug logs are enabled.
+ * <p/>
+ * Warning - not compatible with parallel execution of tests.
```
teryk: how

sonar-plugin-api/src/main/java/org/sonar/api/utils/log/ConsoleLogger.java
```diff
@@ -0,0 +1,148 @@
...
+ * Slow implementation based on {@link java.lang.System#out}. It is not production-ready and it must be used
+ * only for the tests that do not have logback dependency.
+ * <p/>Implementation of message patterns is naive. It does not support escaped '{' and '}'
+ * arguments.
+ */
+class ConsoleLogger extends BaseLogger {
+
+  private final PrintStream stream;
+
+  ConsoleLogger(String unusedName) {
```
teryk: unused so useless ?

sonar-application/src/main/java/org/sonar/application/AppLogging.java
```diff
@@ -0,0 +1,102 @@
...
+      // used by SonarSource testing environment
+      copyGobblerToConsole();
+    }
+    return ctx;
+  }
+
+  /**
+   * Enable the copy in console of the logs written in logs/sonar.log
+   */
+  private void copyGobblerToConsole() {
```
teryk: why the gobbler name ?

teryk:
What's the work left on the computation part ?

sonar-plugin-api/src/test/java/org/sonar/api/utils/log/NullInterceptorTest.java
teryk: rename to do_not_throws_exception and I'm good

sonar-plugin-api/src/main/java/org/sonar/api/utils/log/LogTester.java
simonbrandhof: Parallel execution of tests is already limited to multiple JVM forks. In-process parallel execution is not supported because of H2 and Elasticsearch. So it does not change anything.

sonar-plugin-api/src/main/java/org/sonar/api/utils/log/ConsoleLogger.java
simonbrandhof: Stupid I am. Name is mandatory in Loggers#newInstance(String) but not in BaseLogger. 
I fix it.

sonar-application/src/main/java/org/sonar/application/AppLogging.java
simonbrandhof: the bootstrap process of sonar-application forks ES and Web processes and forwards all their outputs to file. Gobbler is the name for process output.

teryk: Ok

PR #95, opened 2015-02-17
SONAR-6009 Each time the batch call the project referentials WS, a E/S query is executed for each rule

julienlancelot:
SONAR-6009 Each time the batch call the project referentials WS, a E/S query is executed for each rule

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/batch/ProjectRepositoryLoader.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/batch/ProjectRepositoryLoader.java
@@ -20,6 +20,7 @@
 
 package org.sonar.server.batch;
 
+import com.google.common.base.Function;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
@@ -67,7 +68,7 @@ public class ProjectRepositoryLoader implements ServerComponent {
   private final Languages languages;
 
   public ProjectRepositoryLoader(DbClient dbClient, QProfileFactory qProfileFactory, QProfileLoader qProfileLoader, RuleService ruleService,
-                                 Languages languages) {
+    Languages languages) {
     this.dbClient = dbClient;
     this.qProfileFactory = qProfileFactory;
     this.qProfileLoader = qProfileLoader;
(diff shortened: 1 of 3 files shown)


server/sonar-server/src/main/java/org/sonar/server/batch/ProjectRepositoryLoader.java
```diff
@@ -88,8 +89,9 @@ public ProjectRepositories load(ProjectRepositoryQuery query) {
       if (module != null) {
         if (query.isPreview()) {
           // Scan permission is enough to analyze all projects but preview permission is limited to projects user can access
-          UserSession.get().checkComponentPermission(UserRole.USER, query.getModuleKey(),
-            "You're not authorized to access to project '" + module.name() + "', please contact your SonarQube administrator.");
+          if (!UserSession.get().hasProjectPermissionByUuid(UserRole.USER, module.projectUuid())) {
```
jblievremont: Just to understand here, what was the issue with ``checkComponentPermission`` vs ``hasProjectPermission``? Slower request?

server/sonar-server/src/main/java/org/sonar/server/batch/ProjectRepositoryLoader.java
```diff
@@ -224,12 +226,13 @@ private QualityProfileDto getProfile(String languageKey, @Nullable String projec
 
   private void addActiveRules(ProjectRepositories ref) {
     for (org.sonar.batch.protocol.input.QProfile qProfile : ref.qProfiles()) {
-      for (ActiveRule activeRule : qProfileLoader.findActiveRulesByProfile(qProfile.key())) {
-        Rule rule = ruleService.getNonNullByKey(activeRule.key().ruleKey());
+      Map<RuleKey, ActiveRule> activeRules = activeRuleByRuleKey(qProfileLoader.findActiveRulesByProfile(qProfile.key()));
+      for (Rule rule : ruleService.search(new RuleQuery().setQProfileKey(qProfile.key()).setActivation(true), new QueryContext()).getHits()) {
```
jblievremont: Nice case of memory/resource tradeoff :+1:

jblievremont:
LGTM!

server/sonar-server/src/main/java/org/sonar/server/batch/ProjectRepositoryLoader.java
julienlancelot: @jblievremont Indeed, checkComponentPermission will first search the project of the component, then check permission on the found project.
But direct check on hasProjectPermissionByUuid or checkProjectUuidPermission will only check permission on the project.

To sum up, there's one less SQL query when using XXXProjectUuisPermission !

server/sonar-server/src/main/java/org/sonar/server/batch/ProjectRepositoryLoader.java
julienlancelot: :)

PR #96, opened 2015-02-18
SONAR-5885 Fix mapping of htmlDesc attribute to fetch correct fields in ES

jblievremont:
The issue here was that a simple ``Mapper`` does not declare which document fields the WS field maps to, whereas an ``IndexMapper`` allows this.


simonbrandhof:
LGTM

PR #97, opened 2015-02-18
SONAR-5659 Issues link parameters disappear when fail to log

julienlancelot:
SONAR-5659 Issues link parameters disappear when fail to log


stas-vilchik:
LGTM

PR #98, opened 2015-02-18
Fix issue date filters

jblievremont:
Up to 4.5.x, ``createdAfter``/``createdBefore`` used strict comparison, excluding the bounds. When we migrated to ES in 5.0, those filters became inclusive, which poses an issue with differential periods. The first commit of this PR reverts to exclusive behavior.

The second commit avoids 500 error when ``createdAfter == createdBefore`` (there is a check that createdAfter.before(createdBefore) to avoid issues with the date histogram aggregation on creation date).

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
@@ -318,13 +318,13 @@ public class IssueIndex extends BaseIndex {
     if (createdAfter != null) {
       filters.put("__createdAfter", FilterBuilders
         .rangeFilter(IssueIndexDefinition.FIELD_ISSUE_FUNC_CREATED_AT)
-        .gte(createdAfter)
+        .gt(createdAfter)
         .cache(false));
     }
     if (createdBefore != null) {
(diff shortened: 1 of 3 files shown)


server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
```diff
@@ -336,7 +336,7 @@ private void addDatesFilter(Map<String, FilterBuilder> filters, IssueQuery query
   private void validateCreationDateBounds(Date createdBefore, Date createdAfter) {
     Preconditions.checkArgument(createdAfter == null || createdAfter.before(system.newDate()),
       "Start bound cannot be in the future");
-    Preconditions.checkArgument(createdAfter == null || createdBefore == null || createdAfter.before(createdBefore),
+    Preconditions.checkArgument(createdAfter == null || createdAfter.equals(createdBefore) || createdBefore == null || createdAfter.before(createdBefore),
       "Start bound cannot be larger than end bound");
```
teryk: larger or equals

teryk:
if you update the message in the case dates are equals
  return LGTM;
else
  Thread.sleep(Long.MAX_VALUE);

PR #101, opened 2015-02-18
SONAR-6174 Ignore any exception during file_sources migration when  pars...

henryju:
...ing duplication data

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/db/migrations/Select.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/db/migrations/Select.java
@@ -22,6 +22,7 @@ package org.sonar.server.db.migrations;
 import javax.annotation.CheckForNull;
 
 import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.util.Date;
@@ -104,6 +105,25 @@ public interface Select extends SqlStatement<Select> {
 
     public byte[] getBytes(int columnIndex) throws SQLException {
       return rs.getBytes(columnIndex);
+    }
+
+    @Override
+    public String toString() {
+      try {
+        ResultSetMetaData rsMetaData = rs.getMetaData();
+        StringBuilder sb = new StringBuilder();
+        for (int i = 1; i <= rsMetaData.getColumnCount(); i++) {
+          if (i > 1) {
+            sb.append(",");
+          }
+          sb.append(rsMetaData.getColumnLabel(i).toLowerCase());
+          sb.append("=");
+          sb.append(rs.getObject(i));
+        }
+        return sb.toString();
+      } catch (Exception e) {
+        return "Unavailable: " + e.getMessage();
+      }
     }
   }
 
--- a/server/sonar-server/src/main/java/org/sonar/server/db/migrations/SelectImpl.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/db/migrations/SelectImpl.java
@@ -45,6 +45,8 @@ class SelectImpl extends BaseSqlStatement<Select> implements Select {
         rows.add(reader.read(row));
       }
       return rows;
+    } catch (Exception e) {
+      throw newExceptionWithRowDetails(row, e);
     } finally {
       DbUtils.closeQuietly(rs);
       close();
@@ -60,6 +62,8 @@ class SelectImpl extends BaseSqlStatement<Select> implements Select {
         return reader.read(row);
       }
       return null;
+    } catch (Exception e) {
+      throw newExceptionWithRowDetails(row, e);
     } finally {
       DbUtils.closeQuietly(rs);
       close();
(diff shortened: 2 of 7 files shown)


server/sonar-server/src/main/java/org/sonar/server/db/migrations/Select.java
```diff
@@ -105,6 +106,25 @@ public Date getDate(int columnIndex) throws SQLException {
...
+          if (i > 1) {
+            sb.append(",");
+          }
+          sb.append(rsMetaData.getColumnLabel(i).toLowerCase());
+          sb.append("=");
+          sb.append(rs.getObject(i));
+        }
+        return sb.toString();
+      } catch (Exception e) {
+        return "Unavailable: " + e.getMessage();
```
julienlancelot: Why not throw an exception in such case ?

server/sonar-server/src/test/java/org/sonar/server/db/migrations/MassUpdateTest.java
```diff
@@ -0,0 +1,147 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class MassUpdateTest {
+
+  @Rule
+  public ExpectedException thrown = ExpectedException.none();
+
+  @ClassRule
+  public static DbTester db = new DbTester().schema(FeedFileSourcesTest.class, "schema.sql");
```
julienlancelot: The class is wrong, it should be MassUpdateTest.class

server/sonar-server/src/test/java/org/sonar/server/db/migrations/MassUpdateTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+  @Before
+  public void setUp() throws Exception {
+    db.executeUpdateSql("truncate table metrics");
+    db.executeUpdateSql("truncate table snapshots");
+    db.executeUpdateSql("truncate table snapshot_sources");
+    db.executeUpdateSql("truncate table projects");
+    db.executeUpdateSql("truncate table project_measures");
+    db.executeUpdateSql("truncate table file_sources");
+
```
julienlancelot: Why this test is using so many tabled ? Only one should be enough

server/sonar-server/src/test/java/org/sonar/server/db/migrations/MassUpdateTest.java
```diff
@@ -0,0 +1,147 @@
...
+    system = mock(System2.class);
+    Date now = DateUtils.parseDateTime("2014-11-17T16:27:00+0100");
+    when(system.now()).thenReturn(now.getTime());
+    migration = new FakeMigration(db.database());
+  }
+
+  @Test
+  public void testDisplayOfRowDataInCaseOfUnexpectedError() throws Exception {
```
julienlancelot: Please use the lower case with '_' convention

server/sonar-server/src/main/java/org/sonar/server/db/migrations/MassUpdate.java
```diff
@@ -28,6 +30,8 @@
 
 public class MassUpdate {
 
+  private static final Logger LOG = LoggerFactory.getLogger(MassUpdate.class);
```
julienlancelot: This LOG is never used

server/sonar-server/src/test/java/org/sonar/server/db/migrations/MassUpdateTest.java
```diff
@@ -0,0 +1,147 @@
...
+      "ON s.root_project_id = p.id " +
+      "JOIN projects f " +
+      "ON s.project_id = f.id " +
+      "LEFT JOIN file_sources fs " +
+      "ON fs.file_uuid = f.uuid " +
+      "WHERE " +
+      "f.enabled = ? " +
+      "AND f.scope = 'FIL' " +
+      "AND p.scope = 'PRJ' AND p.qualifier = 'TRK' " +
+      "AND fs.file_uuid IS NULL";
```
julienlancelot: The select should be more simple, it's too complex here

server/sonar-server/src/main/java/org/sonar/server/db/migrations/MassUpdate.java
```diff
@@ -76,8 +80,12 @@ public void execute(final Handler handler) throws SQLException {
...
         public void handle(Select.Row row) throws SQLException {
-          if (handler.handle(row, update)) {
-            update.addBatch();
+          try {
+            if (handler.handle(row, update)) {
+              update.addBatch();
+            }
+          } catch (Exception e) {
+            throw new IllegalStateException("Error during processing of row: [" + row + "]", e);
           }
```
julienlancelot: This try catch should be moved in each method of SelectImpl (list, get and scroll around reader.read(row) and handler.handle(row)), in order to manage more exceptions.

server/sonar-server/src/main/java/org/sonar/server/db/migrations/Select.java
henryju: Because I'm using row.toString() in MassUpdate to provide more details during an error. So this exception would hide the original one. As a general rule I think a toString method should never throw an exception. It's a pain during debugging in IDEs.

julienlancelot:
LGTM !

PR #102, opened 2015-02-18
SONAR-5711 Send validation messages to client all at once

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/exceptions/Errors.java
```diff
@@ -63,6 +63,12 @@ public boolean isEmpty() {
     return messages.isEmpty();
   }
 
+  public void checkMessages() {
+    if (!isEmpty()) {
+      throw new BadRequestException(this);
+    }
+  }
+
```
julienlancelot: I do not think it's the responsibility of this class to do that. So please move this method where you use it (in RuleCreator).

julienlancelot:
Move the Errors.checkMessage() to RuleCreator and it LGTM.

PR #103, opened 2015-02-18
adding new property to timeline widget

laurents59:
in order to draw 0 values where metric is undefined (but others are).
default value is false (activate manually in order to use)


teryk:
hi thanks for your pull request. Jira ticket created: https://jira.codehaus.org/browse/SONAR-6219

PR #104, opened 2015-02-18
Technical monitoring of web server - SONAR-4318 SONAR-5936

teryk:


teryk:
obviously, it lacks tests at the moment. It works on my box, both json and MBeans accessed via VisualVM

PR #106, opened 2015-02-19
SONAR-6046 When calling a web service with some wrong data, the generated exception does not contains the called URL

julienlancelot:
SONAR-6046 When calling a web service with some wrong data, the generated exception does not contains the called URL

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/ws/ServletRequest.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/ws/ServletRequest.java
@@ -67,4 +67,14 @@ public class ServletRequest extends ValidatingRequest {
     }
     return null;
   }
+
+  @Override
+  public String toString() {
+    StringBuffer url = source.getRequestURL();
+    String query = source.getQueryString();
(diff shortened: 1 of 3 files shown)


server/sonar-server/src/main/java/org/sonar/server/ws/ServletRequest.java
```diff
@@ -67,4 +67,14 @@ protected InputStream readInputStreamParam(String key) {
     }
     return null;
   }
+
+  @Override
+  public String toString() {
+    StringBuffer url = source.getRequestURL();
+    String query = source.getQueryString();
```
jblievremont: This will not work with POST requests and post-data parameters (but then POST requests can contain pretty big parameters).

julienlancelot: Indeed, but as you say POST params can be pretty big, so we'll do nothing more for the moment for POST.

jblievremont:
LGTM!

PR #110, opened 2015-02-19
SONAR-6011 Restoring built-in profiles does not update existing rule parameters

julienlancelot:
SONAR-6011 Restoring built-in profiles does not update existing rule parameters


simonbrandhof:
LGTM

PR #111, opened 2015-02-19
Add IPC with shared memory instead of file creation.

ehartmann:


server/sonar-process/src/main/java/org/sonar/process/ProcessCommands.java
```diff
@@ -42,75 +43,71 @@
...
+  private final RandomAccessFile sharedmemory;
+  /**
+   * The ByteBuffer will contains :
+   * <ul>
+   *   <li>First byte will contains 0x00 until stop command is issued = 0xFF</li>
+   *   <li>Then each 9 bytes will contains first the status of process (0x01 : READY) then a long (8 bytes) with the latest ping</li>
+   * </ul>
+   */
+  final MappedByteBuffer mappedByteBuffer;
+  private static final Integer MAX_SHARED_MEMORY = 1 + 9 * 10; // We this shared memory we can handle up to 10 processes (91 bytes)
```
simonbrandhof: typo: "we this shared memory" --> "with this shared memory" ?

server/sonar-process/src/main/java/org/sonar/process/ProcessCommands.java
```diff
@@ -42,75 +43,71 @@
  */
 public class ProcessCommands {
 
-  private final File readyFile, stopFile;
+  private final RandomAccessFile sharedmemory;
```
simonbrandhof: This field can be moved as local variable

server/sonar-process/src/main/java/org/sonar/process/ProcessCommands.java
simonbrandhof: Integer can be replaced by int

server/sonar-process/src/main/java/org/sonar/process/StopWatcher.java
```diff
@@ -45,23 +45,19 @@ public StopWatcher(ProcessCommands commands, Stoppable stoppable) {
...
   @Override
   public void run() {
-    commands.prepare();
-    try {
-      while (watching) {
-        if (commands.askedForStop()) {
-          LoggerFactory.getLogger(getClass()).info("Stopping process");
-          stoppable.stopAsync();
+    while (watching) {
+      commands.ping();
```
simonbrandhof: why does it ping when asking for stop ?

sonar-application/src/main/java/org/sonar/application/App.java
```diff
@@ -56,7 +56,7 @@ public void start(Props props) {
     if (props.valueAsBoolean(ProcessConstants.ENABLE_STOP_COMMAND, false)) {
       // stop application when file <temp>/app.stop is created
       File tempDir = props.nonNullValueAsFile(ProcessConstants.PATH_TEMP);
-      ProcessCommands commands = new ProcessCommands(tempDir, "app");
+      ProcessCommands commands = new ProcessCommands(tempDir, 1);
```
simonbrandhof: Index is an internal information used only by process monitor. I think indices should be hidden from caller and should automatically generated and incremented in Monitor#start().

server/sonar-process/src/main/java/org/sonar/process/StopWatcher.java
ehartmann: If my understanding is correct, it's the thread that monitor if there is an askedForStop every 500ms. I though it's the best way to both monitor the flag and update the field containing the last ping command. I think that we do not need another thread just for pinging.

server/sonar-process/src/main/java/org/sonar/process/ProcessCommands.java
ehartmann: Exact, I'll update the code.

server/sonar-process/src/main/java/org/sonar/process/ProcessCommands.java
ehartmann: Exact, I'll update the code.

server/sonar-process/src/main/java/org/sonar/process/StopWatcher.java
simonbrandhof: Two options:
* rename the class as it does not do only stop watching but also pinging
* remove ping() as it's still not used for autokill
I'm for option 2.

server/sonar-process/src/main/java/org/sonar/process/ProcessCommands.java
```diff
@@ -42,75 +45,99 @@
...
+      return -1;
+    }
   }
 
   /**
    * To be executed by monitor process to ask for child process termination
    */
   public void askForStop() {
-    createFile(stopFile);
+    mappedByteBuffer.put(offset() + 1, STOP);
```
simonbrandhof: you should document why you add 1

server/sonar-process/src/main/java/org/sonar/process/ProcessCommands.java
```diff
@@ -42,75 +45,99 @@
...
   }
 
   public boolean askedForStop() {
-    return stopFile.exists();
+    return mappedByteBuffer.get(offset() + 1) == STOP;
```
simonbrandhof: you should document why you add 1

server/sonar-process/src/main/java/org/sonar/process/ProcessCommands.java
ehartmann: Done

server/sonar-process/src/main/java/org/sonar/process/ProcessCommands.java
ehartmann: Done

PR #112, opened 2015-02-20
speed up uglify js step

stas-vilchik:
The longest step in `grunt build` is `uglify`. I replaced it by `uglify_parallel` that runs tasks in parallel.
Also I enabled `jit-grunt` that enables lazy loading of grunt plugins.

On my machine I get 100% boost (7s vs 14s).


simonbrandhof:
Fail on:
[INFO] jit-grunt: Plugin for the "express" task not found.
[INFO] If you have installed the plugin already, please setting the static mapping.
[INFO] See https://github.com/shootaroo/jit-grunt#static-mappings

simonbrandhof:
LGTM

PR #115, opened 2015-02-24
SONAR-6012 Use protobuf to load previous issues

henryju:


sonar-batch-protocol/src/main/protobuf/batch_input.proto
```diff
@@ -0,0 +1,56 @@
...
+    optional string rule_repository = 3;
+    optional string rule_key = 4;
+    optional int32 line = 5;
+    optional string msg = 6;
+    optional Severity severity = 7;
+    optional bool manual_severity = 8;
+    optional string resolution = 9;
+    optional string status = 10;
+    optional string checksum = 11;
+    optional string assigneeLogin = 12;
```
simonbrandhof: should be renamed assignee_login

sonar-batch-protocol/src/main/protobuf/batch_input.proto
```diff
@@ -0,0 +1,56 @@
...
+    are updated and committed for each change (see src/main/gen-java).
+*/
+
+
+import "constants.proto";
+
+option java_package = "org.sonar.batch.protocol.input";
+option optimize_for = SPEED;
+
+message PreviousIssue {
```
simonbrandhof: Name is not completely exact. It should not be restricted to a time dimension. It's the issue as it exists now in the central repository. It will become the "previous" issue only if the batch report will be sent and integrated.
The name should highlight the fact that it's the reference data used for comparison of results (issue tracking). I suggest names like "BaseIssue", "ReferenceIssue", "RemoteIssue" or "ServerIssue"

henryju: OK for ServerIssue

server/sonar-server/src/main/java/org/sonar/server/batch/IssuesAction.java
```diff
@@ -72,14 +69,13 @@ public void handle(Request request, Response response) throws Exception {
     UserSession.get().checkGlobalPermission(GlobalPermissions.PREVIEW_EXECUTION);
     final String moduleKey = request.mandatoryParam(PARAM_KEY);
 
-    response.stream().setMediaType(MimeTypes.JSON);
-    PreviousIssueHelper previousIssueHelper = PreviousIssueHelper.create(new OutputStreamWriter(response.stream().output(), Charsets.UTF_8));
+    response.stream().setMediaType(MimeTypes.DEFAULT);
```
simonbrandhof: you could add "application/x-protobuf" to MimeTypes class.

server/sonar-server/src/main/java/org/sonar/server/batch/IssuesAction.java
```diff
@@ -87,43 +83,52 @@ public void handle(Request request, Response response) throws Exception {
...
+        issueBuilder.setResolution(resolution);
+      }
+      issueBuilder.setCreationDate(batchIssueDto.getCreationTime());
+      try {
+        issueBuilder.build().writeDelimitedTo(out);
+      } catch (IOException e) {
+        throw new IllegalStateException("Unable to serialize issue", e);
       }
-      return null;
+      issueBuilder.clear();
```
simonbrandhof: nice !

sonar-plugin-api/src/main/java/org/sonar/api/batch/sensor/highlighting/internal/DefaultHighlighting.java
```diff
@@ -47,8 +48,7 @@ public DefaultHighlighting(@Nullable SensorStorage storage) {
     super(storage);
     syntaxHighlightingRuleSet = Sets.newTreeSet(new Ordering<SyntaxHighlightingRule>() {
       @Override
-      public int compare(@Nullable SyntaxHighlightingRule left,
-        @Nullable SyntaxHighlightingRule right) {
+      public int compare(@Nonnull SyntaxHighlightingRule left, @Nonnull SyntaxHighlightingRule right) {
```
simonbrandhof: All parameters are non null by default (see package_info.java), so no need for annotations @Nonnull.

henryju: Is it not the workaround to override @Nullable annotation declared in interface Ordering?

simonbrandhof: Contract is inherited and constraints can't be relaxed. It would break Liskov substitution principle. I had the same pb with Guava Function. The solution was to create class NonNullInputFunction (see in sonar-core)

PR #116, opened 2015-02-24
Allow to set a default assignee on issues - SONAR-6154

teryk:


sonar-plugin-api/src/main/java/org/sonar/api/config/Category.java
```diff
@@ -26,7 +26,7 @@
 /**
  * @since 3.7
  */
-public class Category {
+public class  Category {
```
simonbrandhof: unexpected change

sonar-plugin-api/src/main/java/org/sonar/api/CoreProperties.java
```diff
@@ -564,4 +569,9 @@
    */
   String IMPORT_UNKNOWN_FILES_KEY = "sonar.import_unknown_files";
 
+  /**
+   * @since 5.1
+   */
+  String DEFAULT_ISSUE_ASSIGNEE = "sonar.assign.default";
```
simonbrandhof: the last part "default" is misleading. Is it a boolean or a string ?
I would rename sonar.assign.defaultLogin (is it a login ?)

simonbrandhof: moreover it the property is in the category "issues", maybe property should be renamed "sonar.issues.defaultAssigneeLogin"

sonar-plugin-api/src/main/java/org/sonar/api/config/Category.java
teryk: thanks taken into account

sonar-plugin-api/src/main/java/org/sonar/api/CoreProperties.java
teryk: it's a login, you're right. updated

server/sonar-server/src/main/java/org/sonar/server/computation/ComputationContext.java
```diff
@@ -50,11 +53,20 @@ public ComponentDto getProject() {
   }
 
   public BatchReport.Metadata getReportMetadata() {
-    Preconditions.checkState(reportMetadata != null, "Report metadata is available after execution of " + ParseReportStep.class);
+    checkState(reportMetadata != null, "Report metadata is available after execution of " + ParseReportStep.class);
```
jblievremont: Shouldn't the message read "Report metadata is not available"?

server/sonar-server/src/test/java/org/sonar/server/computation/ComputationServiceTest.java
```diff
@@ -66,7 +68,7 @@ public void process_project() throws Exception {
     when(steps.orderedSteps()).thenReturn(Arrays.asList(projectStep1, projectStep2, viewStep));
 
     // load report from db and parse it
-    ComputationService sut = new ComputationService(dbClient, steps, activityService);
+    ComputationService sut = new ComputationService(dbClient, steps, activityService, mock(ProjectSettingsFactory.class, Mockito.RETURNS_DEEP_STUBS));
```
jblievremont: Is RETURNS_DEEP_STUBS necessary here? The mocked ProjectSettingsFactory should return a true Settings object.

server/sonar-server/src/main/java/org/sonar/server/computation/ComputationContext.java
teryk: nope, the reportMetadata is set during the parse report step. If we try to access this data before, it throws a exception

jblievremont: OK then I would reword it as "Report metadata is available only after execution of ".

server/sonar-server/src/test/java/org/sonar/server/computation/step/ParseReportStepTest.java
```diff
@@ -63,22 +70,53 @@ public void setUp() throws Exception {
...
+  public void default_assignee_is_null_when_no_property() throws Exception {
+    sut = new ParseReportStep(mock(IssueComputation.class), mock(DbClient.class, Mockito.RETURNS_DEEP_STUBS), mock(TempFolder.class));
+
+    String assignee = sut.defaultAssignee(null);
+
+    assertThat(assignee).isNull();
+  }
+
+  @Test
+  public void default_assignee_is_null_when_not_found_in_database() throws Exception {
```
jblievremont: With the new logging API, you could check that the INFO log is printed.

server/sonar-server/src/main/java/org/sonar/server/computation/issue/IssueComputation.java
```diff
@@ -47,13 +49,13 @@ public IssueComputation(RuleCache ruleCache, SourceLinesCache linesCache, ScmAcc
     this.diskIssuesAppender = issueCache.newAppender();
   }
 
-  public void processComponentIssues(ComputationContext context, String componentUuid, Iterable<BatchReport.Issue> issues) {
+  public void processComponentIssues(ComputationContext context, String componentUuid, Iterable<BatchReport.Issue> issues, @Nullable String defaultAssignee) {
```
simonbrandhof: Adding such nullable parameters should be avoided. With such pattern in a few weeks this method will have more and more nullable params.

server/sonar-server/src/main/java/org/sonar/server/computation/step/ParseReportStep.java
```diff
@@ -69,7 +77,7 @@ public void execute(ComputationContext context) {
 
       // and parse!
       int rootComponentRef = reportMetadata.getRootComponentRef();
-      recursivelyProcessComponent(reader, context, rootComponentRef);
+      recursivelyProcessComponent(reader, context, rootComponentRef, defaultAssignee(context.getProjectSettings().getString(DEFAULT_ISSUE_ASSIGNEE)));
```
simonbrandhof: it's not the responsibility of ParseReportStep to load information required by IssueComputation

server/sonar-server/src/test/java/org/sonar/server/computation/ComputationServiceTest.java
teryk: ok, no need for RETURNS_DEEP_STUBS, one fairy saved. I check that the newProjectSettings method is called

server/sonar-server/src/main/java/org/sonar/server/computation/step/ParseReportStep.java
teryk: I'm aware of the problem but i want to avoid loading the same information for each component. An alternative is to load the information in the ComputationContext.

simonbrandhof: You mean you want to cache the result of context.getProjectSettings().getString(DEFAULT_ISSUE_ASSIGNEE) ?

simonbrandhof: Moreover calls to default assignee must not be frequent. And even if they are, I don't expect a big difference between calls to cached variable versus context.getProjectSettings().getString(DEFAULT_ISSUE_ASSIGNEE).
Cache over cache...

teryk: I mean I check in the database if the login exists. I don't want to assign it to an issue if the login doesn't exist in the database.

server/sonar-server/src/main/java/org/sonar/server/computation/issue/IssueComputation.java
```diff
@@ -20,40 +20,54 @@
...
     this.ruleCache = ruleCache;
     this.linesCache = linesCache;
     this.scmAccountCache = scmAccountCache;
+    this.userIndex = userIndex;
     this.diskIssuesAppender = issueCache.newAppender();
   }
 
   public void processComponentIssues(ComputationContext context, String componentUuid, Iterable<BatchReport.Issue> issues) {
     linesCache.init(componentUuid);
+    String defaultAssignee = defaultAssignee(context.getProjectSettings().getString(CoreProperties.DEFAULT_ISSUE_ASSIGNEE));
```
simonbrandhof: I think it's a bug. The local variable defaultAssignee conflicts with the field with same name.

server/sonar-server/src/main/java/org/sonar/server/computation/issue/IssueComputation.java
```diff
@@ -20,40 +20,54 @@
...
     for (BatchReport.Issue reportIssue : issues) {
       DefaultIssue issue = toDefaultIssue(context, componentUuid, reportIssue);
       if (issue.isNew()) {
         guessAuthor(issue);
-        autoAssign(issue);
+        autoAssign(issue, defaultAssignee);
```
simonbrandhof: why the state field "defaultAssignee" is not used ?

PR #118, opened 2015-02-25
enable web tests coverage report

stas-vilchik:
* `mvn clean install -DskipWebTests` without tests
* `mvn clean install` standard build with tests but without coverage
* `mvn clean install -Panalysis` with tests and with coverage


server/sonar-web/Gruntfile.coffee
```diff
@@ -336,6 +342,14 @@ module.exports = (grunt) ->
           parallel: true
           port: expressPort
         src: ['<%= pkg.sources %>js/tests/e2e/tests/**/*.js']
+      testCoverage:
```
jblievremont: Can it work with "parallel: true"?

stas-vilchik:
some measures on my box (for sonar-web module only):
* `mvn clean install` Total time: 01:26 min
* `mvn clean install -DskipWebTests` Total time: 21.949 s
* `mvn clean install -Panalysis	` Total time: 05:05 min

server/sonar-web/Gruntfile.coffee
stas-vilchik: @jblievremont Well, with `parallel: true` it randomly fails on tests with browser popup windows. I didn't find the reason yet.

jblievremont:
LGTM. Too bad it does not include CoffeeScript code, but we have a plan for this :)

jblievremont:
For even more consistency, I would move the resources from ``src/test`` to ``src/test/js`` and eventually create a ``src/test/js/specs`` for the current contents of ``src/test/js``. This way we would have:

* src/test/js
  * json
  * lib.js
  * server-coverage.js
  * server.js
  * views
  * specs
    * xxx-spec.js
    * yyy-spec.js
    * ...

PR #119, opened 2015-02-25
feature/server/ce-data

simonbrandhof:
See comment of commit


simonbrandhof:
Branch was renamed feature/server/ce_data

PR #121, opened 2015-02-27
SONAR-6222 SQALE Rating is badly render in the timeline widget

julienlancelot:
SONAR-6222 SQALE Rating is badly render in the timeline widget


teryk:
LGTM

PR #124, opened 2015-02-28
Added sonar property `sonar.search.host`

schnatterer:
In certain environments (e.g. OpenShift) you cannot bind to 127.0.0.1 or you can only bind to certain port ranges.
That's why we have the `sonar.web.host`, `sonar.web.port`, `sonar.search.httpPort` and `sonar.search.port` properties.
What's missing is a `sonar.search.host`. I don't see any other way to run the recent versions of SonarQube on such environments.

See also my question on [stackoverflow](http://stackoverflow.com/questions/28594063/sonarqube-change-the-elastic-search-client-host).

### DIFF
--- a/server/sonar-process/src/main/java/org/sonar/process/ProcessConstants.java
+++ b/server/sonar-process/src/main/java/org/sonar/process/ProcessConstants.java
@@ -48,12 +48,12 @@ public interface ProcessConstants {
   String PATH_TEMP = "sonar.path.temp";
   String PATH_WEB = "sonar.path.web";
 
+  String SEARCH_HOST = "sonar.search.host";
   String SEARCH_PORT = "sonar.search.port";
   String SEARCH_JAVA_OPTS = "sonar.search.javaOpts";
   String SEARCH_JAVA_ADDITIONAL_OPTS = "sonar.search.javaAdditionalOpts";
   String SEARCH_TYPE = "sonar.search.type";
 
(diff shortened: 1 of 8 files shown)


simonbrandhof:
Thanks for the contribution. I think we should have a single property to declare ip address. There's no need to have different values for Elasticsearch and web server.
Options are:
* reuse sonar.web.host
* replace sonar.web.host by sonar.host (or sonar.address, which is less misleading IMO).
WDYT ?

schnatterer:
Hi @simonbrandhof,
 thanks for your input. That's a good point, actually. Keeps it simpler. 
On the other hand, configurations like "listen to any address for incoming web requests but use only a specific address for binding the search server" will not be possible when both addresses are read from the same property.


Either way, I'd prefer reusing `sonar.web.host`, though. Keeps backward compatibility.

simonbrandhof:
Ok. So we will go on with new property "sonar.search.host". Could you complete the pull request with the related documentation in conf/sonar.properties please ?

schnatterer:
Alright, I reverted to my initial commit. It contains a commented out line for `sonar.search.host` as well as a short description in `sonar-application/src/main/assembly/conf/sonar.properties`. I hope that's what you meant.

simonbrandhof:
FYI https://jira.codehaus.org/browse/SONAR-6293

PR #126, opened 2015-03-02
SONAR-5884 Replace overall coverage with UT coverage, add IT coverage info

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/source/ws/LinesAction.java
```diff
@@ -115,9 +121,12 @@ private void writeSource(List<SourceLineDoc> lines, int from, JsonWriter json) {
         .prop("scmRevision", line.scmRevision());
       Date scmDate = line.scmDate();
       json.prop("scmDate", scmDate == null ? null : DateUtils.formatDateTime(scmDate));
-      json.prop("lineHits", line.overallLineHits())
-        .prop("conditions", line.overallConditions())
-        .prop("coveredConditions", line.overallCoveredConditions());
+      json.prop("lineHits", line.utLineHits())
+        .prop("conditions", line.utConditions())
+        .prop("coveredConditions", line.utCoveredConditions())
+        .prop("itLineHits", line.itLineHits())
```
julienlancelot: why not use "utConditions" and "utCoveredConditions" ? It would made more clear what is returned

julienlancelot:
LGTM

server/sonar-server/src/main/java/org/sonar/server/source/ws/LinesAction.java
stas-vilchik: for the same reason why we use "coverage" and "it coverage" everywhere

julienlancelot: But as different types of coverage exists (ut, it, overall), it seems important to me to know what are returned in the WS, it will be easier to debug (what 'coverage' is ? ut coverage ? it coverage ? overall coverage ? another coverage ?).

PR #127, opened 2015-03-02
Add the ability to search for rules by profiles in RubyRuleService - Needed to fix SQALE-269

julienlancelot:
Add search the ability to search for rules by profiles in RubyRuleService
Needed to fix SQALE-269

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/rule/RubyRuleService.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/rule/RubyRuleService.java
@@ -83,6 +83,11 @@ public class RubyRuleService implements ServerComponent, Startable {
     query.setDebtCharacteristics(RubyUtils.toStrings(params.get("debtCharacteristics")));
     query.setHasDebtCharacteristic(RubyUtils.toBoolean(params.get("hasDebtCharacteristic")));
     query.setSortField(RuleNormalizer.RuleField.NAME);
+    String profile = Strings.emptyToNull((String) params.get("profile"));
+    if (profile != null) {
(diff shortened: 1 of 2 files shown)


server/sonar-server/src/main/java/org/sonar/server/rule/RubyRuleService.java
```diff
@@ -83,6 +83,11 @@ public Rule findByKey(String ruleKey) {
     query.setDebtCharacteristics(RubyUtils.toStrings(params.get("debtCharacteristics")));
     query.setHasDebtCharacteristic(RubyUtils.toBoolean(params.get("hasDebtCharacteristic")));
     query.setSortField(RuleNormalizer.RuleField.NAME);
+    String profile = Strings.emptyToNull((String) params.get("profile"));
+    if (profile != null) {
```
teryk: why not use !Strings.isNullOrEmpty(profile) instead ?

teryk:
LGTM

server/sonar-server/src/main/java/org/sonar/server/rule/RubyRuleService.java
julienlancelot: Because rails can send an empty string, and I don't want add the profile criteria in this case.
Or am I missing something ?

teryk: instead of using emptyToNull you could have used in the test !Strings.isNullOrEmpty(profile). It's really a matter of personal taste, so not a big deal

PR #129, opened 2015-03-03
SONAR-6238 Updating only the name of a rule is not taking in account when synchronizing rules at startup

julienlancelot:
SONAR-6238 Updating only the name of a rule is not taking in account when synchronizing rules at startup

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
@@ -51,12 +51,8 @@ import org.sonar.server.startup.RegisterDebtModel;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
-import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
```diff
@@ -228,7 +224,9 @@ private boolean mergeRule(RulesDefinition.Rule def, RuleDto dto) {
       dto.setName(def.name());
       changed = true;
     }
-    changed = mergeDescription(def, dto);
```
jblievremont: Wow, nasty bug indeed :(

jblievremont:
LGTM!

server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
julienlancelot: Indeed ! :)

PR #133, opened 2015-03-05
SONAR-6214 Document parameter 's' of api/users/search

julienlancelot:
SONAR-6214 Document parameter 's' of api/users/search

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/user/ws/UsersWs.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/user/ws/UsersWs.java
@@ -64,6 +64,10 @@ public class UsersWs implements WebService {
       .setDescription("Comma-separated list of user logins")
       .setExampleValue("admin,sbrandhof");
 
+    action.createParam("s")
+      .setDescription("UTF-8 search query on login or name")
+      .setExampleValue("bran");
+
     RailsHandler.addFormatParam(action);
   }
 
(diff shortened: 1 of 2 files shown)


server/sonar-server/src/main/java/org/sonar/server/user/ws/UsersWs.java
```diff
@@ -64,6 +64,10 @@ private void defineSearchAction(NewController controller) {
       .setDescription("Comma-separated list of user logins")
       .setExampleValue("admin,sbrandhof");
 
+    action.createParam("s")
+      .setDescription("UTF-8 search query")
+      .setExampleValue("sbran");
```
teryk: use an example without starting with a s for clarity sake

server/sonar-server/src/main/java/org/sonar/server/user/ws/UsersWs.java
```diff
@@ -64,6 +64,10 @@ private void defineSearchAction(NewController controller) {
       .setDescription("Comma-separated list of user logins")
       .setExampleValue("admin,sbrandhof");
 
+    action.createParam("s")
+      .setDescription("UTF-8 search query")
```
teryk: specify you can fill a login or name

teryk:
LGTM if remarks applied

teryk:
LGTM

PR #134, opened 2015-03-05
SONAR-6206 Manual Measures of type String are not displayed in the Custom Measure widget

julienlancelot:
SONAR-6206 Manual Measures of type String are not displayed in the Custom Measure widget


jblievremont:
Same condition used twice, could be refactored in a method, else LGTM :)

julienlancelot:
@jblievremont What are you taking about ?

PR #135, opened 2015-03-05
SONAR-6235 Provide predefined issue filters

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/issue/filter/IssueFilterService.java
```diff
@@ -343,6 +353,11 @@ private boolean hasUserSharingPermission(String user) {
     return authorizationDao.selectGlobalPermissions(user).contains(GlobalPermissions.DASHBOARD_SHARING);
   }
 
+  protected boolean isFilterOwnedByLogin(IssueFilterDto filter, final String login) {
```
julienlancelot: I would rename it "isFilterOwnedByUser" or "isFilterOwnedByUserLogin"

server/sonar-server/src/main/java/org/sonar/server/issue/filter/IssueFilterService.java
```diff
@@ -223,19 +227,19 @@ public String getLoggedLogin(UserSession userSession) {
   }
 
   public void verifyCurrentUserCanReadFilter(DefaultIssueFilter issueFilter, String login) {
-    if (!issueFilter.user().equals(login) && !issueFilter.shared()) {
+    if (issueFilter.user() != null && !issueFilter.user().equals(login) && !issueFilter.shared()) {
```
julienlancelot: Why not use the isFilterOwnedByLogin (and allow a null userLogin) ?

server/sonar-server/src/main/java/org/sonar/server/issue/filter/IssueFilterService.java
```diff
@@ -223,19 +227,19 @@ public String getLoggedLogin(UserSession userSession) {
...
       throw new ForbiddenException("User is not authorized to read this filter");
     }
   }
 
   private void verifyCurrentUserCanModifyFilter(DefaultIssueFilter issueFilter, String user) {
-    if (!issueFilter.user().equals(user) && !isAdmin(user)) {
+    if (issueFilter.user() != null && !issueFilter.user().equals(user) && !isAdmin(user)) {
```
julienlancelot: Why not use the isFilterOwnedByLogin (and allow a null userLogin) ?

server/sonar-server/src/main/java/org/sonar/server/issue/filter/IssueFilterService.java
jblievremont: It's not the same class... But I can have another method ``isFilterOwnedByLogin`` which takes a ``DefaultIssueFilter`` instead of an ``IssueFilterDto``

server/sonar-server/src/main/java/org/sonar/server/issue/filter/RegisterIssueFilters.java
```diff
@@ -0,0 +1,96 @@
...
+
+  private void createBuiltinIssueFilters() {
+    Date now = system.newDate();
+
+    DefaultIssueFilter unresolvedIssueFilter = new DefaultIssueFilter().setName(BUILTIN_ISSUE_FILTER_UNRESOLVED)
+      .setShared(true)
+      .setCreatedAt(now)
+      .setUpdatedAt(now)
+      .setData("resolved=false");
+    issueFilterService.save(unresolvedIssueFilter);
```
julienlancelot: Why not use directly the DAO ? This method is only converting the DefaultIssueFilter in IssueFilterDto. 
Moreover, I think that this DefaultIssueFilter makes no more sense, we should replace it in the futur by the DTO.

server/sonar-server/src/main/java/org/sonar/server/issue/filter/IssueFilterService.java
```diff
@@ -98,12 +98,16 @@ public DefaultIssueFilter save(DefaultIssueFilter issueFilter, UserSession userS
     return insertIssueFilter(issueFilter, user);
   }
 
+  DefaultIssueFilter save(DefaultIssueFilter issueFilter) {
```
julienlancelot: This method should be removed, the RegisterIssueFilters should use the DAO

server/sonar-server/src/test/java/org/sonar/server/issue/filter/RegisterIssueFiltersTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.sonar.core.template.LoadedTemplateDto;
+
+import java.util.Date;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class RegisterIssueFiltersTest {
```
julienlancelot: Why not use DbTester ?

server/sonar-server/src/main/java/org/sonar/server/issue/filter/IssueFilterService.java
julienlancelot: Ok, I've haven't noticed that. Hum, I think it would be better to use the DTO everywhere, and to remove DefaultIssueFilter.

server/sonar-server/src/test/java/org/sonar/server/issue/filter/RegisterIssueFiltersTest.java
jblievremont: Did not want to write dbunit XML... But indeed it would avoid mocks.

server/sonar-server/src/main/java/org/sonar/server/issue/filter/IssueFilterService.java
```diff
@@ -222,20 +226,20 @@ public String getLoggedLogin(UserSession userSession) {
     return user;
   }
 
-  public void verifyCurrentUserCanReadFilter(DefaultIssueFilter issueFilter, String login) {
-    if (!issueFilter.user().equals(login) && !issueFilter.shared()) {
+  public void verifyCurrentUserCanReadFilter(IssueFilterDto issueFilter, String login) {
+    if (issueFilter.getUserLogin() != null && !issueFilter.getUserLogin().equals(login) && !issueFilter.isShared()) {
```
julienlancelot: You can use isFilterOwnedByUser() here

server/sonar-server/src/main/java/org/sonar/server/issue/filter/IssueFilterService.java
```diff
@@ -222,20 +226,20 @@ public String getLoggedLogin(UserSession userSession) {
...
       throw new ForbiddenException("User is not authorized to read this filter");
     }
   }
 
-  private void verifyCurrentUserCanModifyFilter(DefaultIssueFilter issueFilter, String user) {
-    if (!issueFilter.user().equals(user) && !isAdmin(user)) {
+  private void verifyCurrentUserCanModifyFilter(IssueFilterDto issueFilter, String user) {
+    if (issueFilter.getUserLogin() != null && !issueFilter.getUserLogin().equals(user) && !isAdmin(user)) {
```
julienlancelot: You can use isFilterOwnedByUser() here

julienlancelot:
LGTM !

PR #136, opened 2015-03-06
SONAR-4624 complete "new FP issues" with resolution "won't fix"

simonbrandhof:
SONAR-4624 complete "new FP issues" with resolution "won't fix"

-> dispatcher becomes "Issues resolved as false positive or won't fix"


teryk:
LGTM

PR #137, opened 2015-03-06
clean up css z-index

stas-vilchik:


server/sonar-web/src/main/less/variables.less
```diff
@@ -154,6 +154,50 @@
 
 /*
  * z-index
+ * =======
```
stas-vilchik: interesting part is below: check the order of elements

jblievremont:
LGTM!

PR #139, opened 2015-03-09
Reentrant db migrations when adding/removing columns or tables

simonbrandhof:
Reentrant db migrations when adding/removing columns or tables


julienlancelot:
LGTM !

PR #140, opened 2015-03-09
SONAR-6262 Move links from batch to compute

julienlancelot:
SONAR-6262 Move links from batch to compute


server/sonar-server/src/main/java/org/sonar/server/db/migrations/v52/FeedProjectLinksComponentUuid.java
```diff
@@ -0,0 +1,56 @@
...
+    super(db);
+  }
+
+  @Override
+  public void execute(Context context) throws SQLException {
+    MassUpdate update = context.prepareMassUpdate().rowPluralName("project links");
+    update.select(
+      "SELECT p.uuid, link.id " +
+        "FROM project_links link " +
+        "INNER JOIN projects p ON p.id=link.project_id " +
```
teryk: are we sure that there are no missing row (a project_links row with an empty project_id or a project_id referencing a deleted projects row) ?

julienlancelot: We are sure that they are indeed some project_links linked on no more existing project ids !
That's why I've updated this ticket : https://jira.codehaus.org/browse/SONAR-5503

server/sonar-server/src/test/resources/org/sonar/server/db/migrations/v52/FeedProjectLinksComponentUuidTest/schema.sql
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE "PROJECT_LINKS" (
```
teryk: I would argue that you could remove all the non significant columns, but it's up to you

server/sonar-web/src/main/webapp/WEB-INF/app/models/measure_filter.rb
```diff
@@ -288,18 +288,22 @@ def load_results(snapshot_ids, base_resource)
...
+          project_uuids << snapshot.project.uuid
+          project_id_by_uuid[snapshot.project.uuid] = snapshot.project_id
           rows_by_project_id[snapshot.project_id] = rows_by_snapshot_id[snapshot.id]
         end
 
         links = []
-        project_ids.each_slice(999) do |safe_for_oracle_ids|
-          links.concat(ProjectLink.all(:conditions => {:project_id => safe_for_oracle_ids}, :order => 'link_type'))
+        project_uuids.each_slice(999) do |safe_for_oracle_ids|
+          links.concat(ProjectLink.all(:conditions => {:component_uuid => safe_for_oracle_ids}, :order => 'link_type'))
```
teryk: rename safe_for_oracle_ids to safe_for_oracle_uuids

server/sonar-server/src/test/resources/org/sonar/server/db/migrations/v52/FeedProjectLinksComponentUuidTest/schema.sql
julienlancelot: In fact I prefer let them all, in the case we need another column, we don't need to update tests.

server/sonar-web/src/main/webapp/WEB-INF/app/models/measure_filter.rb
julienlancelot: Indeed !

sonar-core/src/main/resources/org/sonar/core/persistence/schema-h2.ddl
```diff
@@ -202,7 +202,8 @@ CREATE TABLE "PROPERTIES" (
 
 CREATE TABLE "PROJECT_LINKS" (
   "ID" INTEGER NOT NULL GENERATED BY DEFAULT AS IDENTITY (START WITH 1, INCREMENT BY 1),
-  "PROJECT_ID" INTEGER NOT NULL,
+--   "PROJECT_ID" INTEGER NOT NULL,
```
teryk: remove the line

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v52/FeedProjectLinksComponentUuid.java
teryk: so if you declare it as a foreign key in hibernate etc. won't there be a conflict ?
don't you think you should delete project_links that references a deleted projects id ?

teryk:
LGTM

PR #141, opened 2015-03-09
SONAR-6247 Add createdSince parameter to search issues by lifetime

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/issue/IssueQueryService.java
```diff
@@ -135,6 +137,20 @@ public IssueQuery createFromMap(Map<String, Object> params) {
     }
   }
 
+  private Date buildCreatedAfter(@Nullable Date createdAfter, @Nullable String createdSince) {
```
teryk: I would use Preconditions.checkArgument(createdAfter == null || createdSince == null, "createdAfter and createdSince cannot be set simultaneously") at the beginning of the method

server/sonar-server/src/test/java/org/sonar/server/issue/IssueQueryServiceTest.java
```diff
@@ -434,4 +435,41 @@ public void should_search_on_test_file() throws Exception {
...
+  }
+
+  @Test
+  public void set_created_after_from_created_since() {
+    Date now = DateUtils.parseDateTime("2013-07-25T07:35:00+0100");
+    when(system.now()).thenReturn(now.getTime());
+    Map<String, Object> map = newHashMap();
+
+    map.put("createdSince", "1d");
+    assertThat(issueQueryService.createFromMap(map).createdAfter()).isEqualTo(DateUtils.parseDateTime("2013-07-24T07:35:00+0100"));
```
teryk: Don't you think one test is enough, the rest is testing the ISOStandardPeriod formatting

server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionMediumTest.java
```diff
@@ -88,7 +88,7 @@ public void define_action() throws Exception {
     assertThat(show.isPost()).isFalse();
     assertThat(show.isInternal()).isFalse();
     assertThat(show.responseExampleAsString()).isNotEmpty();
-    assertThat(show.params()).hasSize(39);
+    assertThat(show.params()).hasSize(40);
```
teryk: a monster ws action, 40 possible parameters, woaw … (no action needed)

server/sonar-server/src/test/java/org/sonar/server/issue/IssueQueryServiceTest.java
jblievremont: I see it as a kind of behavior documentation and exploratory test.

teryk: I understand for the exploratory test, but it doesn't need to be kept in production test code. Concerning the behavior documentation, I would add tests in the ISOStandardPeriodTest class, if there aren't already

server/sonar-server/src/main/java/org/sonar/server/issue/IssueQueryService.java
jblievremont: Done.

jblievremont:
Should be better now :)

PR #143, opened 2015-03-10
SONAR-6244 assigned_to_me facet must be sticky with respect to assignees

jblievremont:


server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionMediumTest.java
```diff
@@ -446,7 +446,7 @@ public void filter_by_assigned_to_me() throws Exception {
       .setStatus("OPEN")
       .setKee("82fd47d4-b650-4037-80bc-7b112bd4eac2")
       .setSeverity("MAJOR")
-      .setAssignee("john");
+      .setAssignee("john-bob.polop");
```
julienlancelot: Why ?

server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionMediumTest.java
```diff
@@ -466,15 +466,36 @@ public void filter_by_assigned_to_me() throws Exception {
...
+    IssueDto issue2 = IssueTesting.newDto(rule, file, project)
+      .setStatus("OPEN")
+      .setKee("7b112bd4-b650-4037-80bc-82fd47d4eac2")
+      .setAssignee("alice");
+    IssueDto issue3 = IssueTesting.newDto(rule, file, project)
+      .setStatus("OPEN")
+      .setKee("82fd47d4-4037-b650-80bc-7b112bd4eac2");
+    db.issueDao().insert(session, issue1, issue2, issue3);
+    session.commit();
+    tester.get(IssueIndexer.class).indexAll();
```
julienlancelot: MockUserSession.set() seems missing

server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionMediumTest.java
jblievremont: I had an exception when clicking on my name using dory-latest; turns out it was the ``-`` in the middle of my login which triggered a parse error in ES.

server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionMediumTest.java
jblievremont: I don't need a UserSession here :) I'm testing the unauthenticated query

server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
```diff
@@ -478,8 +476,18 @@ private AggregationBuilder createAssigneesFacet(IssueQuery query, Map<String, Fi
...
+
+    Collection<String> assigneesEscaped = query.assignees();
+    if (assigneesEscaped != null) {
+      assigneesEscaped = Collections2.transform(assigneesEscaped, new Function<String, String>() {
+        @Override
+        public String apply(String input) {
+          return Pattern.quote(input);
+        }
+      });
+    }
```
julienlancelot: I would put this in a separated method

julienlancelot: Where is the test on this ?

server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
```diff
@@ -505,14 +513,17 @@ private void addAssignedToMeFacetIfNeeded(SearchRequestBuilder builder, SearchOp
...
-      .filter(facetName)
+    FilterAggregationBuilder facetTopAggregation = AggregationBuilders
+      .filter(facetName + "__filter")
       .filter(facetFilter)
-      .subAggregation(AggregationBuilders.terms(facetName + "__terms").field(fieldName).include(login)));
+      .subAggregation(AggregationBuilders.terms(facetName + "__terms").field(fieldName).include(login));
+
+    builder.addAggregation(
+      AggregationBuilders.global(facetName)
+        .subAggregation(facetTopAggregation));
```
julienlancelot: Would it be possible to test it in IssueIndexTest or a new IssueIndexFacetsTest ?

julienlancelot: Could you explain me what has changed ?

server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionMediumTest.java
julienlancelot: Ok, so it deserves a specific test case

server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionMediumTest.java
julienlancelot: So you need it !
As UserSession is static, it should be defined in every test case. You're lucky that the test is green, but following in which order test cases are executed, this test case might fail.

server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
jblievremont: Wrapping the ``filter`` aggregation in a ``global`` aggregation, to avoid having the query filter applied to the aggregation.

server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionMediumTest.java
jblievremont: OK

server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionMediumTest.java
jblievremont: OK

server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
jblievremont: OK to extract in another method, see test case with dash in name for explanation over escaping.

server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
julienlancelot: Ok, thanks

jblievremont:
Feedback applied, thanks!

julienlancelot:
LGTM

PR #149, opened 2015-03-13
SONAR-6236 Migrate existing dashboards to use the new issue filter widgets

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/dashboard/db/WidgetDao.java
```diff
@@ -38,7 +38,17 @@ protected WidgetDto doGetNullableByKey(DbSession session, Long widgetId) {
     return mapper(session).selectById(widgetId);
   }
 
+  @Override
+  protected WidgetDto doUpdate(DbSession session, WidgetDto item) {
```
julienlancelot: Please remove usage of BaseDao

server/sonar-server/src/main/java/org/sonar/server/dashboard/db/WidgetPropertyDao.java
```diff
@@ -26,6 +26,7 @@
 import org.sonar.server.db.BaseDao;
 
 import java.util.Collection;
+import java.util.List;
 
 public class WidgetPropertyDao extends BaseDao<WidgetPropertyMapper, WidgetPropertyDto, Long> {
```
julienlancelot: Please remove usage of BaseDao

server/sonar-server/src/main/java/org/sonar/server/startup/RegisterDashboards.java
```diff
@@ -55,7 +51,7 @@
   private final LoadedTemplateDao loadedTemplateDao;
 
   public RegisterDashboards(DashboardTemplate[] dashboardTemplatesArray, DashboardDao dashboardDao,
-    ActiveDashboardDao activeDashboardDao, LoadedTemplateDao loadedTemplateDao) {
+    ActiveDashboardDao activeDashboardDao, LoadedTemplateDao loadedTemplateDao, RegisterIssueFilters startupDependency) {
```
julienlancelot: Please add a comment to explain why RegisterIssueFilters is needed only in the constructor

server/sonar-server/src/main/java/org/sonar/server/startup/RenameIssueWidgets.java
```diff
@@ -0,0 +1,151 @@
...
+  public RenameIssueWidgets(DbClient dbClient) {
+    this.dbClient = dbClient;
+  }
+
+  @Override
+  public void start() {
+    DbSession session = dbClient.openSession(false);
+
+    try {
+      if (dbClient.loadedTemplateDao().countByTypeAndKey(LoadedTemplateDto.ONE_SHOT_TASK_TYPE, getClass().getSimpleName()) != 0) {
```
julienlancelot: I'm not sure using the class name to persist in db in a good idea : it the classe name changed in next version, the task will be executed again.
I think using a constant would be better.

server/sonar-server/src/main/java/org/sonar/server/startup/RenameIssueWidgets.java
```diff
@@ -0,0 +1,151 @@
...
+        // Already done
+        return;
+      }
+
+      IssueFilterDto unresolvedIssues = dbClient.issueFilterDao().selectProvidedFilterByName("Unresolved Issues");
+      IssueFilterDto hiddenDebt = dbClient.issueFilterDao().selectProvidedFilterByName("False Positive and Won't Fix Issues");
+      IssueFilterDto myUnresolvedIssues = dbClient.issueFilterDao().selectProvidedFilterByName("My Unresolved Issues");
+
+      if (unresolvedIssues == null || hiddenDebt == null || myUnresolvedIssues == null) {
+        // One of the filter has been deleted, no need to do anything
```
julienlancelot: Why ?

server/sonar-server/src/main/java/org/sonar/server/startup/RenameIssueWidgets.java
```diff
@@ -0,0 +1,151 @@
...
+    } finally {
+      MyBatis.closeQuietly(session);
+    }
+  }
+
+  private void updateWidget(DbSession session, List<Long> widgetIdsWithPropertiesToDelete, WidgetDto widget) {
+    dbClient.widgetDao().update(session,
+      widget.setWidgetKey(PROJECT_ISSUE_FILTER_WIDGET_KEY)
+        .setConfigured(true));
+    widgetIdsWithPropertiesToDelete.add(widget.getId());
```
julienlancelot: Hard to understand this "widgetIdsWithPropertiesToDelete" name. Why not "widgetIdsToDelete" ?
Moreover, add this id in this method is misleading.

server/sonar-server/src/main/java/org/sonar/server/startup/RenameIssueWidgets.java
```diff
@@ -0,0 +1,151 @@
...
+            widgetPropertiesToCreate.add(
+              new WidgetPropertyDto()
+                .setWidgetId(widget.getId())
+                .setPropertyKey(FILTER_PROPERTY)
+                .setTextValue(hiddenDebt.getId().toString()));
+            widgetPropertiesToCreate.add(
+              new WidgetPropertyDto()
+                .setWidgetId(widget.getId())
+                .setPropertyKey(DISTRIBUTION_AXIS_PROPERTY)
+                .setTextValue("resolutions"));
```
julienlancelot: Creating all this WidgetPropertyDto is too verbose, I would create a method to simplify it, something like createPropertyDto(widgetId, propertyKey, textValue).

server/sonar-server/src/test/java/org/sonar/server/startup/RenameIssueWidgetsTest.java
```diff
@@ -0,0 +1,84 @@
...
+    dbTester.prepareDbUnit(this.getClass(), "empty.xml");
+
+    doStart();
+  }
+
+  @Test
+  public void should_skip_when_already_executed() throws Exception {
+    dbTester.prepareDbUnit(this.getClass(), "after.xml");
+
+    doStart();
```
julienlancelot: You should add a assertDbUnit in order to be suer that no new widget are created, otherwise you're tested nothing !

server/sonar-server/src/test/java/org/sonar/server/startup/RenameIssueWidgetsTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    dbTester.assertDbUnit(this.getClass(), "after.xml", "widgets", "widget_properties", "loaded_templates");
+  }
+
+  @Test
+  public void should_skip_when_filter_removed() throws Exception {
+    dbTester.prepareDbUnit(this.getClass(), "empty.xml");
+
+    doStart();
```
julienlancelot: You should add a assertDbUnit or you should check that the widgets and widget_propertes table are empty

server/sonar-server/src/main/java/org/sonar/server/dashboard/db/WidgetPropertyDao.java
```diff
@@ -42,4 +49,8 @@ protected WidgetPropertyDto doGetNullableByKey(DbSession session, Long propertyI
     return mapper(session).selectByDashboard(dashboardKey);
   }
 
+  public void deleteByWidgetIds(DbSession session, List<Long> widgetIdsWithPropertiesToDelete) {
+    mapper(session).deleteByWidgetIds(widgetIdsWithPropertiesToDelete);
```
julienlancelot: If the number of ids is more than 1000, it will fail on Oracle. 
Please use DaoUtils.executeLargeInputs() to prevent such issues.

server/sonar-server/src/main/java/org/sonar/server/startup/RenameIssueWidgets.java
jblievremont: Actually this should not happen:
- on a new install or a migrated database, those issue filters have just been created
- on a restart, this task has already been executed
So I can just remove this case

julienlancelot: Perfect !

server/sonar-server/src/main/java/org/sonar/server/startup/RenameIssueWidgets.java
```diff
@@ -0,0 +1,161 @@
...
+      }
+
+      Map<String, IssueFilterDto> filterByWidgetKey = loadRequiredIssueFilters();
+
+      Map<String, String> distributionAxisByWidgetKey = ImmutableMap.of(
+        WIDGET_FALSE_POSITIVES, "resolutions",
+        WIDGET_MY_UNRESOLVED, "severities",
+        WIDGET_UNRESOLVED_BY_DEVELOPER, "assignees",
+        WIDGET_UNRESOLVED_BY_STATUS, "statuses"
+      );
```
julienlancelot: Far better ! And in fact you could replace WIDGETS_TO_REPLACE set by this map, and use WIDGETS_TO_REPLACE.keys() when you only need the widget keys.

server/sonar-server/src/main/java/org/sonar/server/startup/RenameIssueWidgets.java
```diff
@@ -0,0 +1,161 @@
...
+      List<WidgetPropertyDto> newWidgetProperties = Lists.newArrayList();
+
+      for (WidgetDto widget : dbClient.widgetDao().findAll(session)) {
+        String widgetKey = widget.getWidgetKey();
+        if (WIDGETS_TO_REPLACE.contains(widgetKey)) {
+          newWidgetProperties.add(createFilterProperty(filterByWidgetKey.get(widgetKey), widget));
+          newWidgetProperties.add(createDistributionAxisProperty(distributionAxisByWidgetKey.get(widgetKey), widget));
+          updateWidget(session, widget);
+          updatedWidgetIds.add(widget.getId());
+        }
```
julienlancelot: Great !

server/sonar-server/src/test/java/org/sonar/server/startup/RenameIssueWidgetsTest.java
```diff
@@ -0,0 +1,84 @@
...
+    dbTester.assertDbUnit(this.getClass(), "after.xml", "widgets", "widget_properties", "loaded_templates");
+  }
+
+  @Test
+  public void should_skip_when_already_executed() throws Exception {
+    dbTester.prepareDbUnit(this.getClass(), "after.xml");
+
+    doStart();
+
+    dbTester.assertDbUnit(this.getClass(), "after.xml", "widgets", "widget_properties", "loaded_templates");
```
julienlancelot: :+1:

julienlancelot:
Apply my last feedback and LGTM !

PR #151, opened 2015-03-14
Update maven-gpg-plugin to 1.6

mkordas:


ehartmann:
Thanks, @mkordas.
According to the release notes of gpg plugin 1.6, mainly this upgrade will allow to encrypt the passphrase with Maven encryption mechanism.
Is it why you opened this PR, @mkordas  ?

mkordas:
@ehartmann, yes, I think it makes sense to use latest version. It was released 2 months ago, so it should be stable enough.

PR #152, opened 2015-03-14
Update maven-jxr-plugin to 2.5

mkordas:


simonbrandhof:
Do we still need JXR ?

mkordas:
That's a good question, probably not. Should I prepare PR with JXR removed?

simonbrandhof:
No, it's ok, let's keep it.

PR #154, opened 2015-03-16
Fix an issue with TimeZone ID of UTC

jblievremont:
In the current code, when the server's default time zone is UTC, the formatted string is ``Z`` which is not a valid ``TimeZone`` ID.

### DIFF
--- a/sonar-plugin-api/src/main/java/org/sonar/api/utils/System2.java
+++ b/sonar-plugin-api/src/main/java/org/sonar/api/utils/System2.java
@@ -28,6 +28,7 @@ import javax.annotation.CheckForNull;
 import java.util.Date;
 import java.util.Map;
 import java.util.Properties;
+import java.util.TimeZone;
 
 /**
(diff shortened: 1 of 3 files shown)


sonar-plugin-api/src/main/java/org/sonar/api/utils/System2.java
```diff
@@ -126,6 +127,10 @@ public Date newDate() {
     return new Date();
   }
```
simonbrandhof: "@since 5.2" is missing here

simonbrandhof:
Missing javadoc then LGTM

sonar-plugin-api/src/main/java/org/sonar/api/utils/System2.java
jblievremont: 5.1 you mean?

simonbrandhof: yes

PR #155, opened 2015-03-16
Merge multiple issues column requests alter into one request

julienlancelot:
Merge multiple issues column requests alter into one request


server/sonar-server/src/main/java/org/sonar/server/db/migrations/AddColumnsBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+    return this;
+  }
+
+  public String build() {
+    StringBuilder sql = new StringBuilder().append("ALTER TABLE ").append(tableName).append(" ");
+    addSql(sql);
+    return sql.toString();
+  }
+
+  private void addSql(StringBuilder sql){
```
simonbrandhof: Minor. Personal feeling, I would remove this private method as it breaks readability and does not reduce complexity of build().

server/sonar-server/src/main/java/org/sonar/server/db/migrations/AddColumnsBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+        addColumns(sql, null);
+        break;
+      default:
+        sql.append("ADD (");
+        addColumns(sql, null);
+        sql.append(")");
+    }
+  }
+
+  private void addColumns(StringBuilder sql, @Nullable String columnPrefix) {
```
simonbrandhof: Nullable parameters should be avoided as much as possible. I would replace null by empty "". It also removes the condition if (columnPrefix != null)

server/sonar-server/src/main/java/org/sonar/server/db/migrations/DropColumnsBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+        sql.append("DROP (");
+        dropColumns(sql, null);
+        sql.append(")");
+        break;
+      default:
+        throw new IllegalStateException(String.format("Unsupported database '%s'", dialect.getId()));
+    }
+  }
+
+  private void dropColumns(StringBuilder sql, @Nullable String columnPrefix) {
```
simonbrandhof: Same feedback on nullable here.

sonar-core/src/test/java/org/sonar/core/persistence/DbTester.java
```diff
@@ -311,6 +310,42 @@ public void assertDbUnit(Class testClass, String filename, String... tables) {
     }
   }
 
+  public void assertColumnDefinition(String table, String column, int expectedType, @Nullable Integer expectedSize) {
+    try (Connection connection = openConnection();
```
simonbrandhof: coverage decreased because of try-with-resource :-)

sonar-core/src/test/java/org/sonar/core/persistence/DbTester.java
```diff
@@ -311,6 +310,42 @@ public void assertDbUnit(Class testClass, String filename, String... tables) {
...
+      PreparedStatement stmt = connection.prepareStatement("select * from " + table);
+      ResultSet res = stmt.executeQuery()) {
```
simonbrandhof: Statement and ResultSet are not closed.

sonar-core/src/test/java/org/sonar/core/persistence/DbTester.java
```diff
@@ -311,6 +310,42 @@ public void assertDbUnit(Class testClass, String filename, String... tables) {
...
+      Integer columnIndex = getColumnIndex(res, column);
+      if (columnIndex == null) {
+        fail("The column '" + column + "' does not exists");
```
simonbrandhof: "does not exist" without s!

server/sonar-server/src/main/java/org/sonar/server/db/migrations/AddColumnsBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+      this.isNullable = isNullable;
+      return this;
+    }
+
+    public ColumnDef setLimit(@Nullable Integer limit) {
+      this.limit = limit;
+      return this;
+    }
+
+    public ColumnDef setName(String name) {
```
simonbrandhof: Can you check if name must always be lower-case ? If yes, then check should be done here.

simonbrandhof:
Good! LGTM

server/sonar-server/src/main/java/org/sonar/server/db/migrations/AddColumnsBuilder.java
julienlancelot: Indeed, I've introduced it because I was thinking of managing both ADD and DROP operation in the same classe, but I changed my mind.

server/sonar-server/src/main/java/org/sonar/server/db/migrations/AddColumnsBuilder.java
julienlancelot: Good idea !

sonar-core/src/test/java/org/sonar/core/persistence/DbTester.java
julienlancelot: Are you sure ? There are in the try-with-resources, so there are automatically closed

sonar-core/src/test/java/org/sonar/core/persistence/DbTester.java
julienlancelot: Thanks

server/sonar-server/src/main/java/org/sonar/server/db/migrations/AddColumnsBuilder.java
julienlancelot: Good point !

PR #159, opened 2015-03-19
SONAR-6236 Use correct replacement for widgets on global dashboard without project

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/startup/RenameIssueWidgets.java
```diff
@@ -137,11 +139,18 @@ private WidgetPropertyDto createWidgetProperty(String key, String value, WidgetD
...
     dbClient.widgetDao().update(session,
-      widget.setWidgetKey(WIDGET_PROJECT_ISSUE_FILTER)
+      widget.setWidgetKey(getReplacementWidgetKey(session, widget))
         .setUpdatedAt(system.newDate())
         .setConfigured(true));
   }
 
+  private String getReplacementWidgetKey(DbSession session, WidgetDto widget) {
+    DashboardDto dashboard = dbClient.dashboardDao().getNullableByKey(session, widget.getDashboardId());
+    boolean isOnGlobalDashboard = dashboard != null && dashboard.getGlobal();
```
julienlancelot: What happen if dashboard is null ?

julienlancelot: You could for instance throw an exception, or ignore it with a warning, or remove the widget.

julienlancelot:
Just one feedback, and LGTM !

PR #163, opened 2015-03-20
SONAR-6012 Issues on files on removed modules should contain the module key

julienlancelot:
SONAR-6012 Issues on files on removed modules should contain the module key

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/component/db/ComponentDao.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/component/db/ComponentDao.java
@@ -95,12 +95,16 @@ public class ComponentDao extends BaseDao<ComponentMapper, ComponentDto, String>
     return mapper(session).findSubProjectsByComponentUuids(keys);
   }
 
-  public List<ComponentDto> selectModulesTree(DbSession session, String rootComponentUuid) {
-    return mapper(session).selectModulesTree(rootComponentUuid, Scopes.PROJECT);
+  public List<ComponentDto> selectDescendantModules(DbSession session, String rootComponentUuid) {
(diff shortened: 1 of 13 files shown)


server/sonar-server/src/main/java/org/sonar/server/component/db/ComponentDao.java
```diff
@@ -95,12 +95,19 @@ public boolean existsById(Long id, DbSession session) {
...
-  public List<ComponentDto> selectModulesTree(DbSession session, String rootComponentUuid) {
-    return mapper(session).selectModulesTree(rootComponentUuid, Scopes.PROJECT);
+  public List<ComponentDto> selectEnabledModulesTree(DbSession session, String rootComponentUuid) {
+    return mapper(session).selectModulesTree(rootComponentUuid, Scopes.PROJECT, true);
   }
 
-  public List<FilePathWithHashDto> selectModuleFilesTree(DbSession session, String rootComponentUuid) {
-    return mapper(session).selectModuleFilesTree(rootComponentUuid, Scopes.FILE);
+  /**
+   * Return all child and grand child modules (including disabled ones) from a given project uuid.
```
teryk: selectEnabledModulesTree do not return children and grand children ? (BTW children instead of child if you mean plural)

teryk:
lgtm-pulp-fiction

server/sonar-server/src/main/java/org/sonar/server/component/db/ComponentDao.java
julienlancelot: In fact I just wanted to add a comment to precise that this method also returns disabled modules. But maybe this comment is not needed... WDYT ?
And I will indeed replace child by children

PR #165, opened 2015-03-20
SONAR-6275 Feed measures in compute report

henryju:


server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistMeasuresStep.java
```diff
@@ -95,7 +94,12 @@ public MeasureDto apply(BatchReport.Measure batchMeasure) {
 
   @VisibleForTesting
   MeasureDto toMeasureDto(BatchReport.Measure in, BatchReport.Component component) {
-    checkMeasure(in);
+    if (!in.hasValueType()) {
```
teryk: that's all we can check for sure ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistMeasuresStep.java
```diff
@@ -111,13 +115,33 @@ MeasureDto toMeasureDto(BatchReport.Measure in, BatchReport.Component component)
...
     setData(in, out);
     return out;
   }
 
+  /**
+   * return the numerical value as a double. It's the type used in db.
+   * Returns null if no numerical value found
+   */
+  @CheckForNull
+  private static Double valueAsDouble(BatchReport.Measure measure) {
```
teryk: it would be nice to keep the tests, change the visibility as package-private add @VisibleForTesting

teryk:
LGTM

PR #167, opened 2015-03-23
SONAR-6257 Persist duplication measures in compute

julienlancelot:
SONAR-6257 Persist duplication measures in compute

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistMeasuresStep.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistMeasuresStep.java
@@ -117,6 +117,7 @@ public class PersistMeasuresStep implements ComputationStep {
     out.setMetricId(metricCache.get(in.getMetricKey()).getId());
     out.setRuleId(in.hasRuleKey() ? ruleCache.get(RuleKey.parse(in.getRuleKey())).getId() : null);
     out.setCharacteristicId(in.hasCharactericId() ? in.getCharactericId() : null);
+    out.setPersonId(in.hasPersonId() ? in.getPersonId() : null);
     out.setValue(valueAsDouble(in));
     setData(in, out);
     return out;
--- a/server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistMeasuresStepTest.java
+++ b/server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistMeasuresStepTest.java
@@ -178,6 +178,7 @@ public class PersistMeasuresStepTest extends BaseStepTest {
       .setMetricKey("metric-key")
       .setRuleKey("repo:rule-key")
       .setCharactericId(123456)
+      .setPersonId(5432)
       .build();
 
     MeasureDto measure = sut.toMeasureDto(batchMeasure, component);
@@ -321,6 +322,7 @@ public class PersistMeasuresStepTest extends BaseStepTest {
       .setComponentId(2L)
       .setSnapshotId(3L)
       .setCharacteristicId(123456)
+      .setPersonId(5432)
       .setValue(123.123d)
       .setTendency(2)
       .setVariation(1, 1.1d)
--- a/server/sonar-server/src/test/java/org/sonar/server/measure/persistence/MeasureDaoTest.java
+++ b/server/sonar-server/src/test/java/org/sonar/server/measure/persistence/MeasureDaoTest.java
@@ -141,6 +141,7 @@ public class MeasureDaoTest {
       .setSnapshotId(2L)
       .setMetricId(3)
       .setCharacteristicId(4)
+      .setPersonId(23)
       .setRuleId(5)
       .setComponentId(6L)
       .setValue(2.0d)
--- a/server/sonar-server/src/test/resources/org/sonar/server/measure/persistence/MeasureDaoTest/insert-result.xml
+++ b/server/sonar-server/src/test/resources/org/sonar/server/measure/persistence/MeasureDaoTest/insert-result.xml
@@ -6,7 +6,7 @@
       characteristic_id="4"
       rule_id="5"
       project_id="6"
-      person_id="[null]"
+      person_id="23"
       value="2.0"
       text_value="measure-value"
       tendency="42"
--- a/sonar-batch-protocol/src/main/gen-java/org/sonar/batch/protocol/Constants.java
+++ b/sonar-batch-protocol/src/main/gen-java/org/sonar/batch/protocol/Constants.java
@@ -560,13 +560,13 @@ public final class Constants {
       "\006\n\002CI\020\004B\034\n\030org.sonar.batch.protocolH\001"
     };
     com.google.protobuf.Descriptors.FileDescriptor.InternalDescriptorAssigner assigner =
-        new com.google.protobuf.Descriptors.FileDescriptor.    InternalDescriptorAssigner() {
-          public com.google.protobuf.ExtensionRegistry assignDescriptors(
-              com.google.protobuf.Descriptors.FileDescriptor root) {
-            descriptor = root;
-            return null;
-          }
-        };
+      new com.google.protobuf.Descriptors.FileDescriptor.InternalDescriptorAssigner() {
+        public com.google.protobuf.ExtensionRegistry assignDescriptors(
+            com.google.protobuf.Descriptors.FileDescriptor root) {
+          descriptor = root;
+          return null;
+        }
+      };
     com.google.protobuf.Descriptors.FileDescriptor
       .internalBuildGeneratedFileFrom(descriptorData,
         new com.google.protobuf.Descriptors.FileDescriptor[] {
--- a/sonar-batch-protocol/src/main/gen-java/org/sonar/batch/protocol/input/BatchInput.java
+++ b/sonar-batch-protocol/src/main/gen-java/org/sonar/batch/protocol/input/BatchInput.java
@@ -8,10 +8,10 @@ public final class BatchInput {
   public static void registerAllExtensions(
       com.google.protobuf.ExtensionRegistry registry) {
   }
-  public interface ServerIssueOrBuilder extends
-      // @@protoc_insertion_point(interface_extends:ServerIssue)
-      com.google.protobuf.MessageOrBuilder {
+  public interface ServerIssueOrBuilder
+      extends com.google.protobuf.MessageOrBuilder {
 
+    // optional string key = 1;
     /**
      * <code>optional string key = 1;</code>
      */
@@ -26,6 +26,7 @@ public final class BatchInput {
     com.google.protobuf.ByteString
         getKeyBytes();
 
+    // optional string module_key = 2;
     /**
      * <code>optional string module_key = 2;</code>
      */
@@ -40,6 +41,7 @@ public final class BatchInput {
     com.google.protobuf.ByteString
         getModuleKeyBytes();
 
+    // optional string path = 3;
     /**
      * <code>optional string path = 3;</code>
      */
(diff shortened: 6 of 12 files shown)


server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
```diff
@@ -0,0 +1,142 @@
...
+    return new String[] {Qualifiers.PROJECT};
+  }
+
+  @Override
+  public void execute(ComputationContext context) {
+    DbSession session = dbClient.openSession(true);
+    try {
+      MetricDto duplicationMetric = dbClient.metricDao().selectByKey(session, CoreMetrics.DUPLICATIONS_DATA_KEY);
+      int rootComponentRef = context.getReportMetadata().getRootComponentRef();
+      recursivelyProcessComponent(session, context, null, rootComponentRef, duplicationMetric);
```
teryk: several steps have now a very similar logic. we should probably factor code at some point

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
```diff
@@ -0,0 +1,142 @@
...
+      session.commit();
+    } finally {
+      MyBatis.closeQuietly(session);
+    }
+  }
+
+  private void recursivelyProcessComponent(DbSession session, ComputationContext context, @Nullable Integer parentComponentRef, int componentRef, MetricDto duplicationMetric) {
```
teryk: several methods with 6/7 parameters in this class. Is there something that can be done to improve that ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
```diff
@@ -0,0 +1,142 @@
...
+
+  private static void appendDuplication(StringBuilder xml, String componentKey, BatchReport.DuplicationBlock duplicationBlock) {
+    int length = duplicationBlock.getEndLine() - duplicationBlock.getStartLine();
+    xml.append("<b s=\"").append(duplicationBlock.getStartLine())
+      .append("\" l=\"").append(length)
+      .append("\" r=\"").append(StringEscapeUtils.escapeXml(componentKey))
+      .append("\"/>");
+  }
+
+  private static String createKey(String moduleKey, String path, @Nullable String branch) {
```
teryk: why not use the method directly with a static import ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
```diff
@@ -0,0 +1,142 @@
...
+    String duplicationXml = createXmlDuplications(reportReader, parentComponent, component.getPath(), duplications, branch);
+    MeasureDto measureDto = new MeasureDto()
+      .setMetricId(duplicationMetric.getId())
+      .setData(duplicationXml)
+      .setComponentId(component.getId())
+      .setSnapshotId(component.getSnapshotId());
+    dbClient.measureDao().insert(session, measureDto);
+  }
+
+  private String createXmlDuplications(BatchReportReader reportReader, BatchReport.Component parentComponent, String componentPath,
```
teryk: Would it make sense to have the xml generation in a separate class ?

server/sonar-server/src/main/java/org/sonar/server/measure/persistence/MetricDao.java
```diff
@@ -20,34 +20,28 @@
...
-import java.util.List;
+import javax.annotation.CheckForNull;
 
-public class MetricDao extends BaseDao<MetricMapper, MetricDto, String> implements ServerComponent, DaoComponent {
+import java.util.List;
 
-  public MetricDao() {
-    this(System2.INSTANCE);
-  }
+public class MetricDao implements ServerComponent, DaoComponent {
```
teryk: Nice !

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStepTest.java
```diff
@@ -0,0 +1,393 @@
...
+  public TemporaryFolder temp = new TemporaryFolder();
+
+  @ClassRule
+  public static DbTester dbTester = new DbTester();
+
+  DbSession session;
+
+  DbClient dbClient;
+
+  ComputationStep step;
```
teryk: Could you change with PersistDuplicationMeasuresStep, to show that's the class which is tested ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
julienlancelot: indeed

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
julienlancelot: Why not indeed !

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
julienlancelot: Indeed, I can create a kind of DuplicationContext where I set the session, the context, and the metric

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
julienlancelot: Why ? It will only be used here, and moreover, duplication measures should be removed soner or later

server/sonar-server/src/main/java/org/sonar/server/measure/persistence/MetricDao.java
julienlancelot: :+1:

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStepTest.java
julienlancelot: yep

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStepTest.java
```diff
@@ -0,0 +1,393 @@
...
+  @After
+  public void tearDown() throws Exception {
+    session.close();
+  }
+
+  @Test
+  public void nothing_to_do_when_no_duplication() throws Exception {
+    saveDuplicationMetric();
+
+    File reportDir = temp.newFolder();
```
teryk: metadata and components and the same in this test and the next one. Is it possible to factor metadata, components and duplication writing ? in this class, but it would probably makes sense to have a BatchReportTesting to generate data easily

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStepTest.java
```diff
@@ -0,0 +1,393 @@
...
+      .setRef(2)
+      .setType(Constants.ComponentType.FILE)
+      .setUuid("UUID_B")
+      .setSnapshotId(11L)
+      .setPath("file")
+      .build());
+
+    step.execute(new ComputationContext(new BatchReportReader(reportDir), ComponentTesting.newProjectDto("PROJECT")));
+
+    assertThat(dbTester.countRowsOfTable("project_measures")).isEqualTo(0);
```
teryk: nice, the countRowsOfTable

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStepTest.java
```diff
@@ -0,0 +1,393 @@
...
+        .setEndLine(10)
+        .build())
+      .build();
+    writer.writeComponentDuplications(2, newArrayList(duplication));
+
+    step.execute(new ComputationContext(new BatchReportReader(reportDir), ComponentTesting.newProjectDto("PROJECT")));
+
+    assertThat(dbTester.countRowsOfTable("project_measures")).isEqualTo(1);
+
+    Map<String, Object> dto =  dbTester.selectFirst("select snapshot_id as \"snapshotId\", metric_id as \"metricId\", text_value as \"textValue\" from project_measures");
```
teryk: are quotes needed in snapshot_id as "snapshotId" ? Can't you replace without quotes ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
teryk: seems good

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
teryk: ok, it's up to you then. Maybe it doesn't apply in this case, but sometimes you can create package-private class. It's more a matter of responsabilities.

teryk:
@henryju do you confirm that you send only duplications that must be persisted ? (I'm thinking of the shouldPersistMeasure() method which was called in the DuplicationPersister)

henryju:
@teryk yes I confirm that optimization (best value, persistance mode = memory) is applied on batch side. But I think this is only a temporary state.

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStepTest.java
julienlancelot: You're right creation of the report folder with a project should be factorized in this test.

It should also be a good idea to have a kind of Testing utils, but I think we should wait for this spring to be finished to have a bigger picture.

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStepTest.java
julienlancelot: Removing them make the test fail, because without quote, and for an unknown reason, column name will be returned in upper case.
But we should improve our dbTester framework to simplify such query. No idea for the moment on how to do that.

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStepTest.java
teryk: yep, agreed on waiting for the big picture

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStepTest.java
teryk: let it like that for the time being then

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistDuplicationMeasuresStep.java
julienlancelot: I understand, but I think it's the responsability of this class to persist measures of duplications, and these measures are stored as XML.

teryk: ok, it's fine for me. It is responsible for persisting AND building the duplications. As long as the class is not too long, KISS applies.

PR #168, opened 2015-03-24
Do not show the debt (sub)characteristic of a rule when it has been overloaded as removed

jblievremont:


julienlancelot:
LGTM !

PR #169, opened 2015-03-24
SONAR-6320 Symbol in compute

julienlancelot:
SONAR-6320 Symbol in compute

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistSymbolsStep.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistSymbolsStep.java
@@ -1,0 +1,108 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * SonarQube is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistSymbolsStep.java
```diff
@@ -0,0 +1,108 @@
...
+
+    for (Integer childRef : component.getChildRefList()) {
+      recursivelyProcessComponent(context, childRef);
+    }
+  }
+
+  private void processSymbols(BatchReport.Component component, List<BatchReport.Symbol> symbols) {
+    symbolsByLineForLastProcessedFile = newHashMap();
+    if (!symbols.isEmpty()) {
+      int symbolId = 1;
```
teryk: symbolId is always equal at one ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistSymbolsStep.java
```diff
@@ -0,0 +1,108 @@
...
+        symbolId++;
+      }
+    }
+  }
+
+  private void processSymbolRange(BatchReport.Range range, int symboleId){
+    int startLine = range.getStartLine();
+    if (startLine != range.getEndLine()) {
+      // TODO support symbols on multiple lines when source will be in compute, in order to be able to know the end line in this case
+      throw new IllegalStateException("To be implemented : Symbols on multiple lines are not supported for the moment");
```
teryk: what is the behavior today ? shouldn't we have a symbol that goes until the end of the start line instead of throwing an exception ?

teryk:
lgtm-dog

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistSymbolsStep.java
julienlancelot: It's not possible to support it for the moment as we don't persist sources in compute engine, but it will be fixed very soon when sources will be persisted by the compute engine.
This is only a temporary code to not forget this use case.

PR #172, opened 2015-03-26
SONAR-6318 Highlighting in compute

julienlancelot:


teryk:
I find it a little bit hard to review code that is temporary but in the meantime:

lgtm-pulp-fiction

PR #174, opened 2015-03-27
SONAR-6338 Coverage in compute

julienlancelot:
SONAR-6338 Coverage in compute

### DIFF
--- a/sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/BatchReportReader.java
+++ b/sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/BatchReportReader.java
@@ -37,7 +37,7 @@ public class BatchReportReader {
 
   public BatchReport.Metadata readMetadata() {
     File file = fileStructure.metadataFile();
-    if (isNotAnExistingFile(file)) {
+    if (!doesFileExists(file)) {
       throw new IllegalStateException("Metadata file is missing in analysis report: " + file);
     }
     return ProtobufUtil.readFile(file, BatchReport.Metadata.PARSER);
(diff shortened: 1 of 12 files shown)


sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/BatchReportReader.java
```diff
@@ -37,7 +38,7 @@ public BatchReportReader(File dir) {
 
   public BatchReport.Metadata readMetadata() {
     File file = fileStructure.metadataFile();
-    if (isNotAnExistingFile(file)) {
+    if (!isFileExists(file)) {
```
teryk: I would go for doesFileExist(file) or isFileFound(file). WDYT ?

sonar-batch-protocol/src/main/protobuf/batch_report.proto
```diff
@@ -210,6 +210,24 @@ message Symbols {
   }
 }
 
+message Coverage {
```
teryk: why do I have the feeling it looks like the graph tests I'm trying to move on compute stack ?

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/BatchReportReader.java
julienlancelot: Why not doesFileExist.

teryk:
lgtm-bush

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/BatchReportReader.java
teryk: because it's correct english ?

PR #176, opened 2015-03-29
Fix example response of /api/authentication/validate

hupfis:
The boolean field is not "validate" but "valid" (tested with v 5.0.1)


simonbrandhof:
Thanks, https://jira.codehaus.org/browse/SONAR-6372 created

PR #178, opened 2015-03-30
SONAR-6350 /api/sources/scm should read information from index

julienlancelot:
SONAR-6350 /api/sources/scm should read information from index


server/sonar-server/src/test/resources/org/sonar/server/source/ws/ScmActionTest/accept_negative_value_in_from_parameter.json
```diff
@@ -0,0 +1,15 @@
+{
+  "scm": [
+    [1, "julien", "2015-03-30T12:34:56+0000", "123-456-789"],
+    [2, "julien", "2015-03-29T12:34:56+0000", "123-456-710"],
+    [3, "julien", "2015-03-27T12:34:56+0000", "456-789-101"]
+  ]
+}
+
```
jblievremont: It might be a display artifact, but why are there so many empty lines in your JSON fixtures?

julienlancelot: It's a mistake, I'll remove them

jblievremont:
I would have split the refactoring with the marker interface in a separate commit, but still LGTM!

julienlancelot:
Indeed you're right, I should have create 2 commits. Thanks for the review

PR #183, opened 2015-03-31
SONAR-6300 Add WS to select the default quality profile for a language

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileSetDefaultAction.java
```diff
@@ -0,0 +1,106 @@
...
+  private String getProfileKeyFromLanguageAndName(String language, String profileName) {
+    QProfile profile = profileLookup.profile(profileName, language);
+    if (profile == null) {
+      throw new NotFoundException(String.format("Unable to find a profile for language '%s' with name '%s'", language, profileName));
+    }
+    return profile.key();
+  }
+
+  @Override
+  public void define(WebService.NewController controller) {
```
julienlancelot: It's a personal point of view, but I would put this in the top of the class, it allow to first see the description and the parameter before looking the implementation.

julienlancelot: That what you've done in ProjectAssociationActions for instance.

julienlancelot:
LGTM

PR #184, opened 2015-03-31
SONAR-6338 Stream reading / writing coverage message

julienlancelot:
SONAR-6338 Stream reading / writing coverage message

### DIFF
--- a/sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ProtobufUtil.java
+++ b/sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ProtobufUtil.java
@@ -19,17 +19,13 @@
  */
 package org.sonar.batch.protocol;
 
+import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.Message;
 import com.google.protobuf.Parser;
 
-import java.io.BufferedInputStream;
-import java.io.BufferedOutputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
+import java.io.*;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
 
 public class ProtobufUtil {
   private ProtobufUtil() {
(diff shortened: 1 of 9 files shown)


sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ProtobufUtil.java
```diff
@@ -44,6 +39,43 @@ private ProtobufUtil() {
     }
   }
 
+  public static <MESSAGE extends Message> Iterable<MESSAGE> readFileMessages(File file, final Parser<MESSAGE> parser) {
```
henryju: Personally I don't mind but some people prefer that generic parameter is named with a single letter. So maybe consider renaming MESSAGE -> M

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ProtobufUtil.java
```diff
@@ -44,6 +39,43 @@ private ProtobufUtil() {
...
+    try {
+      final InputStream inputStream = new BufferedInputStream(new FileInputStream(file));
```
henryju: InputStream is never closed. You can close it when last "next()" or last "hasNext()" is called. This should do the trick since we are always supposed to read the full content.

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ProtobufUtil.java
```diff
@@ -44,6 +39,43 @@ private ProtobufUtil() {
...
+      return new Iterable<MESSAGE>() {
+        @Override
+        public Iterator<MESSAGE> iterator() {
```
henryju: It is better to construct the inputStream here so that Iterable is "reentrant".
Iterable iterable = xxx;
iterable.iterator().next() => return first element
iterable.iterator().next() => return first element

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ProtobufUtil.java
```diff
@@ -44,6 +39,43 @@ private ProtobufUtil() {
...
+                currentMessage = parser.parseDelimitedFrom(inputStream);
+              } catch (InvalidProtocolBufferException e) {
+                throw new IllegalStateException("Failed to read input stream", e);
+              }
+              return currentMessage != null;
+            }
+
+            @Override
+            public MESSAGE next() {
+              return currentMessage;
```
henryju: With this design it is mandatory to call hasNext() before each call to next(). And it is not possible to call hasNext() several times. Later someone can be surprised by this behavior that is not the contract of Iterator.
iterator.hasNext();
iterator.next() -> return first element
iterator.hasNext();
iterator.hasNext();
iterator.next() -> return third element

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/BatchReportWriter.java
```diff
@@ -113,9 +113,18 @@ public void writeComponentSyntaxHighlighting(int componentRef, Iterable<BatchRep
...
 
-  public void writeFileCoverage(BatchReport.Coverage coverage) {
-    File file = fileStructure.fileFor(FileStructure.Domain.COVERAGE, coverage.getFileRef());
-    ProtobufUtil.writeToFile(coverage, file);
+  public void writeFileCoverage(int componentRef, Iterable<BatchReport.Coverage> coverageList) {
+    File file = fileStructure.fileFor(FileStructure.Domain.COVERAGE, componentRef);
+    ProtobufUtil.writeMessagesToFile(coverageList, file);
+  }
+
+  public OutputStream openCoverageOutputStream(int componentRef) {
```
henryju: Is this method still used?

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ProtobufUtil.java
julienlancelot: ok

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ProtobufUtil.java
julienlancelot: Good catch !

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ProtobufUtil.java
julienlancelot: ok

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ProtobufUtil.java
julienlancelot: ouch, you're right. Have you an idea to improve this ?

julienlancelot: In fact this design allowed me to implement the hasNext, otherwise I do not know how to do that...

henryju:
LGTM after removal of openCoverageOutputStream

PR #185, opened 2015-03-31
feature/improve_report_stream

julienlancelot:


sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/ReportStream.java
```diff
@@ -0,0 +1,78 @@
...
+      @Override
+      public void remove() {
+        throw new UnsupportedOperationException();
+      }
+    };
+  }
+
+  @Override
+  public void close() throws IOException {
+    IOUtils.closeQuietly(inputStream);
```
henryju: Do you really need to use IOUtils.closeQuietly here? You'll hide exception.

julienlancelot: Oh yes, you're right !

PR #187, opened 2015-04-01
add FluentIterable fork in new module sonar-utils

sns-seb:
FluentIterable is a very useful class to write functionnal code in Java 7 but was introduced with Guava 12. Good thing is, only 2 methods rely on code not available in Guava 10 so the fork is quite easy.
Unit test was also forked from Guava's

Please note that I added guava's guava-testlib library as a new dependency (was used by `FluentIterable` test) which contains pretty convenient classes such as `NullPointerTester`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,7 @@
     <maven.min.version>3.0.5</maven.min.version>
     <maven.api.version>3.0.5</maven.api.version>
     <jdk.min.version>1.7</jdk.min.version>
+    <guava.version>10.0.1</guava.version>
     <timestamp>${maven.build.timestamp}</timestamp>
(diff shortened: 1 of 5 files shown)


simonbrandhof:
We should discuss about the location of the new module sonar-utils. I don't think it should be in server stack.

sns-seb:
I agree this might be the definitive location for this module.

What about moving it around when we have a little more visibility about its content and usages ?

sns-seb:
as discussed during meeting, @simonbrandhof will look into solving the class loader dependency between SQ and the core-api by introducing an intermediary classloader between the server's CL and the plugins classloaders.
If it works, we will be able to use Guava 18 in SQ so this PR will become useless

simonbrandhof:
As discussed we should upgrade to guava 19 instead of forking some parts.

PR #188, opened 2015-04-01
SONAR-6277 Add File source in report

julienlancelot:
SONAR-6277 Add File source in report

### DIFF
--- a/sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/FileStream.java
+++ b/sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/FileStream.java
@@ -1,0 +1,71 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * SonarQube is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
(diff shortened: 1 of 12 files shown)


sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/FileStream.java
```diff
@@ -0,0 +1,63 @@
...
+
+  private final File file;
+  private InputStream inputStream;
+
+  public FileStream(File file) {
+    this.file = file;
+  }
+
+  @Override
+  public Iterator<String> iterator() {
```
henryju: No protection against multiple calling of iterator() for the same FileStream?

julienlancelot: In fact I've been able to fix this. As the input stream is created here, we can call iterator() multiple times, it will work as expected.
(The same in ReportStream, See for instance ReportStreamTest.next_should_be_reentrant())

henryju: But inputStream is erased so no more able to properly close it.
fileStream.iterator()
fileStream.iterator()
fileStream.close() -> close the second inputStream, the first is lost

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/BatchReportReader.java
```diff
@@ -130,6 +131,15 @@ public Issues readDeletedComponentIssues(int deletedComponentRef) {
     return null;
   }
 
+  @CheckForNull
+  public FileStream readSourceLines(int fileRef) {
+    File file = fileStructure.fileFor(FileStructure.Domain.SOURCE, fileRef);
+    if (doesFileExists(file)) {
+      return new FileStream(file);
+    }
+    return null;
```
julienlancelot: Should we return null if no source find ? Will all component of type FILE have source ? If yes, I think we should fail here, WDYT ?

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/FileStream.java
julienlancelot: ouch...

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/BatchReportReader.java
henryju: Currently I don't see a case where file would be missing. So ok to raise an exception.

julienlancelot: ok

henryju:
LGTM

PR #189, opened 2015-04-02
SONAR-5921 WS to list project/profile associations

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileProjectsAction.java
```diff
@@ -0,0 +1,164 @@
...
+      writeProjects(response.newJsonWriter(), pagedAuthorizedProjects, paging);
+    } finally {
+      session.close();
+    }
+  }
+
+  private void checkProfileExists(String profileKey, DbSession session) {
+    try {
+      dbClient.qualityProfileDao().getNonNullByKey(session, profileKey);
+    } catch (IllegalArgumentException doesNotExist) {
```
julienlancelot: I do not like catching an exception is this case. I think it would be better (and more efficient) to use dbClient.qualityProfileDao().getByKey(), and return NotFoundException(String.format("Could not find a quality profile with key '%s'", profileKey)) if the result is null.

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileProjectsAction.java
```diff
@@ -0,0 +1,164 @@
...
+    json.name("results").beginArray();
+    for (ProjectQprofileAssociationDto project : projects) {
+      json.beginObject()
+        .prop("key", project.getProjectUuid())
+        .prop("name", project.getProjectName())
+        .prop("selected", project.isAssociated())
+        .endObject();
+    }
+    json.endArray();
+    json.prop("more", paging.hasNextPage());
```
julienlancelot: I thought this "more" parameter for paging was deprecated (used in rails) and that we should use these parameters instead : total, ps, p.

sonar-core/src/main/java/org/sonar/core/qualityprofile/db/ProjectQprofileAssociationDto.java
```diff
@@ -0,0 +1,58 @@
...
+    return profileKey;
+  }
+
+  public boolean isAssociated() {
+    return profileKey != null;
+  }
+
+  @Override
+  public int compareTo(ProjectQprofileAssociationDto o) {
+    return new CompareToBuilder().append(projectName, o.projectName).toComparison();
```
julienlancelot: Would'nt it be better to use project uuid instead of project name ?

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileProjectsAction.java
jblievremont: :+1:

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileProjectsAction.java
jblievremont: No, it is used by the "select-list.js" component (which will in turn use the page/pageSize parameters).

sonar-core/src/main/java/org/sonar/core/qualityprofile/db/ProjectQprofileAssociationDto.java
jblievremont: The goal is to sort projects by name :)

server/sonar-server/src/test/java/org/sonar/server/qualityprofile/ws/QProfileProjectsActionTest.java
```diff
@@ -0,0 +1,224 @@
...
+    xooP2 = QProfileTesting.newXooP2();
+    dbClient.qualityProfileDao().insert(session, xooP1, xooP2);
+  }
+
+  private TestRequest newRequest() {
+    return wsTester.newGetRequest("api/qualityprofiles", "projects");
+  }
+
+  private ComponentDto newProject(String uuid, String name) {
+    return ComponentTesting.newProjectDto().setUuid(uuid).setName(name);
```
julienlancelot: It's better to use ComponentTesting.newProjectDto(uuid), otherwise the uuid of the project will be different than the projectUuid, and you have strange results if you create 2 projects with newProjectDto().setUuid(uuid)

server/sonar-server/src/test/java/org/sonar/server/qualityprofile/ws/QProfileProjectsActionTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+    associateProjectsWithProfile(session, xooP1, project1, project2);
+
+    session.commit();
+
+    newRequest().setParam("key", xooP1.getKey()).setParam("selected", "all").setParam("query", "project t").execute().assertJson(this.getClass(), "all_filtered.json");
+  }
+
+  @Test(expected = NotFoundException.class)
+  public void should_fail_on_unexisting_profile() throws Exception {
```
julienlancelot: type : unexisting -> nonexistent

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileProjectsAction.java
julienlancelot: But shouldn't we have a standard on paging ? We should not depend on js implementation

sonar-core/src/main/java/org/sonar/core/qualityprofile/db/ProjectQprofileAssociationDto.java
julienlancelot: Oh yes, I better understand. In fact I think we should not "poluate" a DTO with comparison implementation. It's up to each caller of DTOs to implement their own comparator.

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileProjectsAction.java
julienlancelot: @stas-vilchik WDYT ?

server/sonar-server/src/test/java/org/sonar/server/qualityprofile/ws/QProfileProjectsActionTest.java
jblievremont: :+1:

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileProjectsAction.java
stas-vilchik: well, I can update select list component to not depend on "more" parameter

julienlancelot: @stas-vilchik it would be a good idea, as I thing we should have a common way to define paging in WS.
But it's up to you web guys to validate this :)

jblievremont: It's a separate ticket: there are several places that use the select-list component and they all use the "more"+"page/pageSize" pagination convention (users/groups, groups/users, permissions/users, permissions/groups, project permissions/users, qualitygates/projects, views/projects).

julienlancelot: After IRL discussion, let's do this in a separate ticket.
I think we should already manage old and new way to manage paging, but it's up to you

julienlancelot:
Apply last small feedback and LGTM

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileProjectsAction.java
jblievremont: For the record, here is the ticket for modification of pagination in select-list: http://jira.codehaus.org/browse/SONAR-6391

PR #190, opened 2015-04-02
Pull define(...) method into a dedicated interface

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/BaseQProfileWsAction.java
```diff
@@ -19,13 +19,14 @@
...
 
+import org.sonar.api.server.ws.Definable;
 import org.sonar.api.server.ws.RequestHandler;
 import org.sonar.api.server.ws.WebService;
 
 /**
  * Marker interface for quality profile related web service end points
  */
-public interface BaseQProfileWsAction extends RequestHandler {
+public interface BaseQProfileWsAction extends RequestHandler, Definable<WebService.NewController> {
```
teryk: why not create a WsAction interface which extends RequestHandler and Definable<WebService.NewController> ? then, when creating a new marker interface action, we would extend WsAction only. WDYT ?

jblievremont: :+1: on my side, but keeping WsAction as an internal interface of sonar-server (with RequestHandler and Definable<T> in sonar-plugin-api)

teryk: :+1:

julienlancelot: Seems good to me

julienlancelot: @jblievremont Will you add the WsAction ?

jblievremont: Yep.

PR #193, opened 2015-04-03
fix start.sh on ubuntu with zsh

sns-seb:
this PR fixes start.sh on Ubuntu with zsh

can anyone check `start.sh` still works on mac, plz ?

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 if [[ "$OSTYPE" == "darwin"* ]]; then
   OS='macosx-universal-64'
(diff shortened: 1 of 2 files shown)


julienlancelot:
@sns-seb Unfortunately, on mac it's no more working, it's rebuilding sources each time start.sh is called.

jblievremont:
Note that the current start.sh does not work on my Ubuntu bash either (but I don't use it so I don't really mind).

The issue here is that the hashbang line ``#!/bin/sh`` enforces raw Bourne shell compatibility mode on GNU bash - which does not understand ``[[ ]]`` conditionals - whereas this is not the case on OSX, which still interprets "advanced" shell features normally only available in evolved Bourne shells.

On my machine it works when replacing ``#!/bin/sh`` with ``#!/bin/bash``, not sure how Zsh handles it though.

sns-seb:
The failing `[[ ]]` has not been introduced by me. So, this script is basically not working on Linux.

What about moving to bash ? It would work on MacOs and Ubuntu. I don't think that would be an issue for sonarsourcers. Do you know if this script used by the community ?

@julienlancelot so it seams the ``` around `ls` command I introduced is not supported on MacOs. Works with raw Bourne Shell on Linux though, weird.

simonbrandhof:
@sns-seb go for bash. No link with community.

sns-seb:
pushed a new version of `start.sh` (and `stop.sh` for consistency) that uses bash

`start.sh` uses another to test for existing file and directory so that hopefully it works on both Linux and MacOs

sns-seb:
merged into master

PR #194, opened 2015-04-03
SONAR-6348 Property sonar.path.temp is ignored by Ruby on Rails application

simonbrandhof:
SONAR-6348 Property sonar.path.temp is ignored by Ruby on Rails application


jblievremont:
LGTM :+1:

PR #195, opened 2015-04-03
Hasnext() should not call doNext() when iterator already closed

julienlancelot:
Hasnext() should not call doNext() when iterator already closed


simonbrandhof:
LGTM

GithubIntegration:
SonarQube analysis reported no new issue.

PR #197, opened 2015-04-07
add db migration from java

sns-seb:
This PR adds a component that wraps around the ActiveRecord based migration process and handles concurrency as well as asynchronous execution of it

this will be called from the new WS to trigger DB migration (another PR)

Sorry this PR is a little bit dense, but individual commit do not really make sense


sns-seb:
rebased on master

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * The time and day the last migration was started.
+   * <p>
+   * If no migration was ever started, the returned date is {@code null}. This value is reset when {@link #startIt()} is
+   * called.
+   * </p>
+   *
+   * @return a {@link Date} or {@code null}
+   */
+  @Nullable
```
simonbrandhof: For now we use CheckForNull annotation for method returns and Nullable for method parameters. That was the rule for a correct support by Findbugs. I have to check now the status with latest Java plugin.

simonbrandhof: I confirm that @CheckForNull must be used (see http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS2259)

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
```diff
@@ -0,0 +1,71 @@
...
+   * Starts the migration status and returns immediately.
+   * <p>
+   * Migration can not be started twice but calling this method wont raise an error.
+   * On the other hand, calling this method when no migration is needed will start the process anyway.
+   * </p>
+   * <p>
+   * <strong>Do not rename this method to {@code start} otherwise it will be called by the pico container</strong>
+   * </p>
+   */
+  void startIt();
```
julienlancelot: Shouldn't it better to rename it 'startMigration()' ?

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
```diff
@@ -0,0 +1,71 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.db.migration;
+
+import javax.annotation.Nullable;
+import java.util.Date;
+
+public interface DatabaseMigration {
```
julienlancelot: Why have you created an interface ?

server/sonar-server/src/main/java/org/sonar/server/platform/AbstractStopableExecutorService.java
```diff
@@ -0,0 +1,121 @@
...
+    @Override
+    public void stop() {
+        delegate.shutdown(); // Disable new tasks from being submitted
+        try {
+            // Wait a while for existing tasks to terminate
+            if (!delegate.awaitTermination(5, TimeUnit.SECONDS)) {
+                delegate.shutdownNow(); // Cancel currently executing tasks
+                // Wait a while for tasks to respond to being cancelled
+                if (!delegate.awaitTermination(5, TimeUnit.SECONDS)) {
+                    System.err.println("DBMigration pool did not terminate");
```
julienlancelot: Why are you using System.err.println ?

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
sns-seb: I should have written in the javadoc "This method should be named {@code start} but it can not be because it will be called by the pico container and will cause unwanted behavior"

`startMigration` has the "migration" word duplicated, it terms of object design, that's not so good. Currently, calling this method can be written `databaseMigration.startIt()`. Since I can not use `start`, I really don't know which is worse of `startIt` and `startMigration` :(

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
sns-seb: I should have written in the javadoc "This method should be named {@code start} but it can not be because it will be called by the pico container and will cause unwanted behavior"

`startMigration` has the "migration" word duplicated, it terms of object design, that's not so good. Currently, calling this method can be written `databaseMigration.startIt()`. Since I can not use `start`, I really don't know which is worse of `startIt` and `startMigration` :(

server/sonar-server/src/main/java/org/sonar/server/platform/ruby/PlatformRackBridge.java
```diff
@@ -0,0 +1,64 @@
...
+import org.jruby.rack.RackApplicationFactory;
+import org.picocontainer.Startable;
+
+import javax.servlet.ServletContext;
+import javax.servlet.ServletContextEvent;
+
+/**
+ * Implementation of {@link RackBridge} which get access to the Rack object through the {@link ServletContext}.
+ * <p>
+ * This component implements {@link Startable} so that
```
julienlancelot: Missing Javadoc

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
sns-seb: "loose coupling in other components"/"mocking in tests"/"best practice"...

why does it surprise you ?

server/sonar-server/src/main/java/org/sonar/server/platform/AbstractStopableExecutorService.java
sns-seb: good catch

server/sonar-server/src/main/java/org/sonar/server/platform/ruby/PlatformRackBridge.java
sns-seb: Actually, some javadoc cleaning was missing. Good catch

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
julienlancelot: Indeed both are not good, let's keep it like it is

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
julienlancelot: Because I think that when an interface is not used (only one implementation), it's better to have only the implementation, it reduces the number of classes :)

sns-seb:
rebased on master and took feeback into account

sns-seb:
rebased on master
+ fix build instability caused by `AbstractClosableExecutorService`
+ fix incorrect indent in source files

GithubIntegration:
SonarQube analysis reported 7 new issues:
* 1 blocking
* 4 major
* 1 minor
* 1 info

simonbrandhof:
The new package org.sonar.server.db.migration conflicts with org.sonar.server.db.migrations. Moreover both contain a class named DatabaseMigration.

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
```diff
@@ -0,0 +1,72 @@
...
+   * <p>
+   * Migration can not be started twice but calling this method wont raise an error.
+   * On the other hand, calling this method when no migration is needed will start the process anyway.
+   * </p>
+   * <p>
+   * <strong>This method should be named {@code start} but it can not be because it will be called by the pico container
+   * and this will cause unwanted behavior</strong>
+   * </p>
+   */
+  void startIt();
```
simonbrandhof: To avoid ambiguity, why not naming the method "migrate" ?

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
```diff
@@ -0,0 +1,72 @@
...
+   *
+   * @return a {@link Date} or {@code null}
+   */
+  @CheckForNull
+  Date startedAt();
+
+  /**
+   * Current status of the migration.
+   *
+   * @return a {@link Status}
```
simonbrandhof: Seriously, it returns a Status ? :-) This javadoc can be dropped!

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
```diff
@@ -0,0 +1,72 @@
...
+   */
+  Status status();
+
+  /**
+   * The error of the last migration if it failed, otherwise {@code null}.
+   * <p>
+   * This value is reset when {@link #startIt()} is called.
+   * </p>
+   * @return a {@link Throwable} or {@code null}
```
simonbrandhof: this javadoc duplicates the signature too.

server/sonar-server/src/main/java/org/sonar/server/db/migration/PlatformDatabaseMigration.java
```diff
@@ -0,0 +1,126 @@
...
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
+import java.util.Date;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * Handles concurrency to make sure only one DB migration can run at a time.
+ */
+public class PlatformDatabaseMigration implements DatabaseMigration {
```
simonbrandhof: What's the need for the interface DatabaseMigration if there's only a single implementation ?

simonbrandhof:
The package org.sonar.server.platform was always badly named IMO. According to our past discussion it should be "system", so the package "ruby" deserves to be directly in org.sonar.server.
Moreover the file package_info.java is missing in this directory.

server/sonar-server/src/main/java/org/sonar/server/db/migration/PlatformDatabaseMigration.java
```diff
@@ -0,0 +1,126 @@
...
+    running.getAndSet(true);
+    executorService.execute(new Runnable() {
+      @Override
+      public void run() {
+        status = Status.RUNNING;
+        startDate = new Date();
+        failureError = null;
+        try {
+          Loggers.get(PlatformDatabaseMigration.class).info("Starting DB Migration at {}", startDate);
+          rubyBridge.databaseMigration().trigger();
```
simonbrandhof: why RubyBridge does not directly execute migration (for instance "rubyBridge.migrate()") ?

server/sonar-server/src/main/java/org/sonar/server/db/migration/DatabaseMigration.java
sns-seb: such name introduces a "duplication" which is not very Object oriented `DatabaseMigration.migrate`

subject has been discussed with @julienlancelot here : https://github.com/SonarSource/sonarqube/pull/197#discussion-diff-28219803

server/sonar-server/src/main/java/org/sonar/server/db/migration/PlatformDatabaseMigration.java
sns-seb: point also raised by @julienlancelot (https://github.com/SonarSource/sonarqube/pull/197#discussion-diff-28219923)

basically, out of habit and because it really bugs me to mock a concrete class and it is also more costly => slower tests

simonbrandhof: ok, makes sense.

sns-seb:
> The new package org.sonar.server.db.migration conflicts with org.sonar.server.db.migrations. Moreover both contain a class named DatabaseMigration

Wanted to put new `DatabaseMigration` into package `org.sonar.server.db.migration` but a class with the same name already existed in the package so I created a new package.

Existing class `DatabaseMigration` is poorly named imo, should be named `MigrationStep` or something alike.

If I could rename the existing class, I would gladly drop package `org.sonar.server.db.migrations`

simonbrandhof:
Ok, go for renaming to `MigrationStep` and to drop `org.sonar.server.db.migrations`

server/sonar-server/src/main/java/org/sonar/server/db/migration/PlatformDatabaseMigration.java
sns-seb: I intended `RubyBridge` to create wrappers around Ruby code which could then be put into pico and this way avoid creating the underlying wrapping code each time.

Did not get that far because it meant to instance classes and wrappers in `ServerComponents` which would have messed the code.

It can be removed.

simonbrandhof: no pb here. You can forget.

sns-seb:
@simonbrandhof  I guess you mean to keep `org.sonar.server.db.migrations` (I did the mistake myself) since it is the legacy package (and is full of stuff) ?

simonbrandhof:
exact. The goal is to move your code in package `org.sonar.server.db.migrations`

server/sonar-server/src/main/java/org/sonar/server/db/migration/PlatformDatabaseMigration.java
sns-seb: In the end, I think I'll leave it as it is as it keeps open the possibility to create the `RubyDatabaseMigration` object returned by `rubyBridge.databaseMigration()` in the constructor of `PlatformDatabaseMigration` and save on all the Ruby script parsing and casting which occurs when `RubyBrigde` methods are called

sns-seb:
updated PR :

* moved `ruby` package to `org.sonar.server`
* renamed `DatabaseMigration` to `DatabaseStep` (and `DatabaseMigrations` to `DatabaseSteps`), ran upgrade-sonar-it to check it was ok
* removed package `org.sonar.server.db.migration`
* moved `StoppableExecutorService` to `org.sonar.server.util` (and fixed typo in name)

simonbrandhof:
LGTM

PR #200, opened 2015-04-07
'Days since last commit' metric in platform out of the box - SONAR-6353

teryk:
end to end test done: check the measure is correctly computed for a given project in the « Measures » page.

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputationSteps.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputationSteps.java
@@ -41,6 +41,7 @@ public class ComputationSteps {
       ParseReportStep.class,
 
       // Persist data
+      PersistNumberOfDaysSinceLastCommitStep.class,
       PersistMeasuresStep.class,
       PersistIssuesStep.class,
       PersistComponentLinksStep.class,
--- a/server/sonar-server/src/main/java/org/sonar/server/source/index/SourceLineIndex.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/source/index/SourceLineIndex.java
@@ -28,6 +28,7 @@ import org.sonar.server.es.BaseIndex;
 import org.sonar.server.es.EsClient;
 import org.sonar.server.exceptions.NotFoundException;
 
+import java.util.Date;
 import java.util.List;
 
 import static com.google.common.base.Preconditions.checkArgument;
@@ -111,5 +112,22 @@ public class SourceLineIndex extends BaseIndex {
       return new SourceLineDoc(result[0].sourceAsMap());
     }
     throw new NotFoundException(String.format("No source found on line %s for file '%s'", line, fileUuid));
+  }
+
+  public Date lastCommitedDateOnProject(String projectUuid) {
+    SearchRequestBuilder request = getClient().prepareSearch(SourceLineIndexDefinition.INDEX)
+      .setTypes(SourceLineIndexDefinition.TYPE)
+      .setSize(1)
+      .setQuery(QueryBuilders.boolQuery()
+        .must(QueryBuilders.termQuery(SourceLineIndexDefinition.FIELD_PROJECT_UUID, projectUuid)))
+      .setFetchSource(new String[]{SourceLineIndexDefinition.FIELD_SCM_DATE}, null)
+      .addSort(SourceLineIndexDefinition.FIELD_SCM_DATE, SortOrder.DESC);
+
+    SearchHit[] result = request.get().getHits().getHits();
+    if (result.length > 0) {
+      return new SourceLineDoc(result[0].sourceAsMap()).scmDate();
+    }
+
+    throw new NotFoundException(String.format("No source found on project '%s'", projectUuid));
   }
 }
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistNumberOfDaysSinceLastCommitStep.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistNumberOfDaysSinceLastCommitStep.java
@@ -1,0 +1,122 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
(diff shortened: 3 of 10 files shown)


server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputationSteps.java
```diff
@@ -41,6 +41,7 @@
       ParseReportStep.class,
 
       // Persist data
+      ComputeAndPersistDaysSinceLastCommitStep.class,
```
julienlancelot: I would rename the class "PersistDaysSinceLastCommitStep", no need to specify the "compute" word from my point of view

GithubIntegration:
SonarQube analysis reported no new issue.

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputeAndPersistDaysSinceLastCommitStep.java
```diff
@@ -0,0 +1,122 @@
...
+public class ComputeAndPersistDaysSinceLastCommitStep implements ComputationStep {
+
+  private static final long MILLISECONDS_PER_DAY = 1000 * 60 * 60 * 24L;
+
+  private final DbClient dbClient;
+  private final SourceLineIndex sourceLineIndex;
+  private final MetricCache metricCache;
+  private final System2 system;
+
+  private long lastCommitTimestamp = 0L;
```
julienlancelot: A step should be stateless, but having a shared variable is very dangerous if we forget to reset it. I would recommand to use it as a local variable and to use it as a method parameter.

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputationSteps.java
teryk: Compute is the "Decorator" part and persistence is the persistence part … for the time being we could consider 2 separate operations. WDYT ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputeAndPersistDaysSinceLastCommitStep.java
teryk: Why the step should be stateless ? Each component are created for one analysis and one analysis only, executed sequentially ? Why do this step should be specifically stateless ?
In this specific case, I'm ok as being stateless is not too painful.

server/sonar-server/src/main/java/org/sonar/server/source/index/SourceLineIndex.java
```diff
@@ -112,4 +113,20 @@ public SourceLineDoc getLine(String fileUuid, int line) {
     }
     throw new NotFoundException(String.format("No source found on line %s for file '%s'", line, fileUuid));
   }
+
+  public Date lastCommitedDateOnProject(String projectUuid) {
+    SearchRequestBuilder request = getClient().prepareSearch(SourceLineIndexDefinition.INDEX)
+      .setTypes(SourceLineIndexDefinition.TYPE)
+      .setSize(1)
+      .setQuery(QueryBuilders.boolQuery()
```
julienlancelot: Specify the field to return with setFetchSource(new String[]{SourceLineIndexDefinition.FIELD_SCM_REVISION}) to decrease memory used by the query

server/sonar-server/src/test/java/org/sonar/server/computation/step/ComputeAndPersistDaysSinceLastCommitStepTest.java
```diff
@@ -0,0 +1,137 @@
...
+    reportWriter.writeComponentScm(
+      BatchReport.Scm.newBuilder()
+        .setComponentRef(2)
+        .addChangeset(
+          BatchReport.Scm.Changeset.newBuilder()
+            .setDate(threeDaysAgo)
+        )
+        .build()
+      );
+    ComputationContext context = new ComputationContext(new BatchReportReader(dir), new ComponentDto().setId(456_789L).setUuid("project-uuid"));
```
julienlancelot: Setting the id is useless. I would also use ComponentTesting.newProject("project-uuid")

julienlancelot:
Apply my feedback, then LGTM

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputationSteps.java
julienlancelot: For the time being I'm not sure it makes sense to have these both verbs, but as you wish, we'll change that when working in the "Decorator" sprint.

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputeAndPersistDaysSinceLastCommitStep.java
julienlancelot: Indeed, I forget that the step is executed once per project. So ok to keep as it is.

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputationSteps.java
teryk: done

server/sonar-server/src/main/java/org/sonar/server/source/index/SourceLineIndex.java
teryk: done, I suppose you were talking about SourceLineIndexDefinition.FIELD_SCM_DATE

server/sonar-server/src/test/java/org/sonar/server/computation/step/ComputeAndPersistDaysSinceLastCommitStepTest.java
teryk: done

server/sonar-server/src/main/java/org/sonar/server/source/index/SourceLineIndex.java
julienlancelot: Indeed !

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputeAndPersistDaysSinceLastCommitStep.java
julienlancelot: Just one thing, if tomorrow we're using more than one worker, don't we gonna have an issue with this ?

teryk: Not if we load the components per analysis

GithubIntegration:
SonarQube analysis reported no new issue.

PR #203, opened 2015-04-07
SONAR-6258 Persist source, coverage and scm into file sources

julienlancelot:
SONAR-6258 Persist source, coverage and scm into file sources


GithubIntegration:
SonarQube analysis reported 24 new issues:
* 2 major
* 22 info

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLine.java
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.server.computation.source;
+
+import org.sonar.server.source.db.FileSourceDb;
+
+public interface StreamLine {
```
teryk: is it a stream or a line ? if it's a stream as I understand it, the name should be LineStream instead of StreamLine

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
```diff
@@ -0,0 +1,104 @@
...
+package org.sonar.server.computation.source;
+
+import org.sonar.batch.protocol.output.BatchReport;
+import org.sonar.server.source.db.FileSourceDb;
+
+import javax.annotation.CheckForNull;
+
+import java.util.Iterator;
+
+public class StreamLineCoverage implements StreamLine {
```
teryk: do you think the StreamLine name is important in the StreamLineCoverage ? What is it ? Is it important to specify in the class name ?

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Override
+  public void readLine(int line, FileSourceDb.Line.Builder lineBuilder) {
+    BatchReport.Coverage reportCoverage = getNextReport(line);
+    if (reportCoverage != null) {
+      processUnitTest(lineBuilder, reportCoverage);
+      processIntegrationTest(lineBuilder, reportCoverage);
+      processOverallTest(lineBuilder, reportCoverage);
+
+      // Reset coverage
```
teryk: uneeded comment IMO. the statement is clear as it is. Otherwise extract method resetCoverage() which could be called in the constructor

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
```diff
@@ -0,0 +1,104 @@
...
+    if (hasUtLineHits(reportCoverage)) {
+      lineBuilder.setUtLineHits(1);
+    }
+    if (reportCoverage.hasConditions() && reportCoverage.hasUtCoveredConditions()) {
+      lineBuilder.setUtConditions(reportCoverage.getConditions());
+      lineBuilder.setUtCoveredConditions(reportCoverage.getUtCoveredConditions());
+    }
+  }
+
+  private static boolean hasUtLineHits(BatchReport.Coverage reportCoverage){
```
teryk: what do you think of the name hasUnitTests(…) ?

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
```diff
@@ -0,0 +1,104 @@
...
+    return reportCoverage.hasUtHits() && reportCoverage.getUtHits();
```
teryk: I'm surprised by the getUtHits() method which returns a boolean. I would have expected to return the unit tests. WDYT ?

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private static void processIntegrationTest(FileSourceDb.Line.Builder lineBuilder, BatchReport.Coverage reportCoverage){
+    if (hasItLineHits(reportCoverage)) {
```
teryk: same here for hasIntegrationTests(...)

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLine.java
julienlancelot: LineStream is fine for me

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
julienlancelot: After internal discussion, let's rename it LineCoverageReader

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
julienlancelot: Ok to remove the comment

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
julienlancelot: :+1:

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
julienlancelot: We do not have the information of the number of unit tests that touch a line, that's why the batch is only sending a boolean.

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
julienlancelot: :+1:

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
```diff
@@ -0,0 +1,104 @@
...
+      lineBuilder.setOverallLineHits(1);
+    }
+    if (reportCoverage.hasConditions() && reportCoverage.hasOverallCoveredConditions()) {
+      lineBuilder.setOverallConditions(reportCoverage.getConditions());
+      lineBuilder.setOverallCoveredConditions(reportCoverage.getOverallCoveredConditions());
+    }
+  }
+
+  @CheckForNull
+  private BatchReport.Coverage getNextReport(int currentLine) {
```
teryk: I don't understand this method. It gets the next line if it's the current line ? Why return coverage only if coverage.getLine() == currentLine ?

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineScm.java
```diff
@@ -0,0 +1,56 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.server.computation.source;
+
+import org.sonar.batch.protocol.output.BatchReport;
+import org.sonar.server.source.db.FileSourceDb;
+
+public class StreamLineScm implements StreamLine {
```
teryk: rename to ScmLineReader according to our chat

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
julienlancelot: Here's what it's doing : 
- If current element is null, it try to return the next one (if exists)
- If current element exists, it's returning it only if it match the current line, otherwise it return null. This is because it's not mandatory for each line to have coverage info, and the only way to know on which line is the coverage is to read it.
Is it more clear ?

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineScm.java
julienlancelot: :+1:

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
teryk: yep, I understand that. I think the name is misleading. I would have preferred hasUtHits even if I realize that hasHasUtHits is a pain to read

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
teryk: yes and no. I understand the algorithm but still had the feeling it shouldn't work. 
You rely on the fact that you won't go to the next line until it has been asked. I would think that a LineReader would return nextLine or nextLineWithConditions maybe, but I wouldn't mind the current line

teryk: I would rename it getNextLineIfMatchLine(int lineNumber). WDYT ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistFileSourcesStep.java
```diff
@@ -0,0 +1,226 @@
...
+        linesIterator.close();
+      }
+    }
+
+    for (Integer childRef : component.getChildRefList()) {
+      recursivelyProcessComponent(fileSourcesContext, childRef);
+    }
+  }
+
+  private List<StreamLine> createStreamLines(BatchReportReader reportReader, int componentRef) {
```
teryk: rename in lineReaders to be coherent with linesIterator(...) method ?

teryk: maybe better names to know the difference between the report line and the file lines ? WDYT ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistFileSourcesStep.java
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  @Override
+  public void execute(ComputationContext context) {
+    int rootComponentRef = context.getReportMetadata().getRootComponentRef();
+    // Don't use batch insert for file_sources since keeping all data in memory can produce OOM for big files
+    DbSession session = dbClient.openSession(false);
+    try {
+      final Map<String, FileSourceDto> previousFileSourcesByUuid = new HashMap<>();
+      session.select("org.sonar.core.source.db.FileSourceMapper.selectHashesForProject", context.getProject().uuid(), new ResultHandler() {
```
teryk: probably a stupid question, but couldn't it be done with a query on a index ?

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
julienlancelot: Then -> getNextLineCoverageIfMatchLine() ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistFileSourcesStep.java
julienlancelot: No, we do not have all information in the index (missing src hash), and it has nothing to do in it. Moreover, as the goal is to update source file, it seems more safe to use the db as a referential.

server/sonar-server/src/test/java/org/sonar/server/computation/source/ReportIteratorTest.java
```diff
@@ -71,6 +71,15 @@ public void read_report() throws Exception {
     assertThat(sut.next().getLine()).isEqualTo(1);
   }
 
+  @Test
+  public void call_has_next_when_already_closed() throws Exception {
+    sut = new ReportIterator<>(file, BatchReport.Coverage.PARSER);
+    assertThat(sut.next().getLine()).isEqualTo(1);
+    assertThat(sut.hasNext()).isFalse();
+    // not fail
```
teryk: remove comment and add info in the name method … like do_not_fail_when_calling_has_next_with_iterator_already_closed() or something similar

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistFileSourcesStep.java
julienlancelot: I do not understand your second question. For the first one, ok for lineReaders.

server/sonar-server/src/main/java/org/sonar/server/computation/source/StreamLineCoverage.java
julienlancelot: For your question about the algorithm, let's have a call

server/sonar-server/src/test/java/org/sonar/server/computation/source/ReportIteratorTest.java
julienlancelot: ok

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistFileSourcesStep.java
```diff
@@ -0,0 +1,226 @@
...
+
+  private List<StreamLine> createStreamLines(BatchReportReader reportReader, int componentRef) {
+    List<StreamLine> streamLines = newArrayList();
+    File coverageFile = reportReader.readFileCoverage(componentRef);
+    if (coverageFile != null) {
+      streamLines.add(new StreamLineCoverage(new ReportIterator<>(coverageFile, BatchReport.Coverage.PARSER)));
+    }
+    BatchReport.Scm scmReport = reportReader.readComponentScm(componentRef);
+    if (scmReport != null) {
+      streamLines.add(new StreamLineScm(scmReport));
```
teryk: not covered by tests

julienlancelot: Good catch !

teryk:
where do the +24 issues come from ?
for the time being, the continuous integration does not work

julienlancelot:
@teryk Feedback applied, can you check that it's ok ?

teryk:
lgtm-big-lebowski

GithubIntegration:
* 2 major
* 2 info

PR #206, opened 2015-04-09
SONAR-6308 WS to show a profile's changelog

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileChangelogAction.java
```diff
@@ -0,0 +1,142 @@
...
+
+import java.util.Date;
+import java.util.Map.Entry;
+
+public class QProfileChangelogAction implements BaseQProfileWsAction {
+
+  private static final String PARAM_SINCE = "since";
+  private static final String PARAM_TO = "to";
+
+  private QProfileService service;
```
julienlancelot: In order to minimize the dependency of this action, I would remove the usage of QProfileService. To do so, I would create a new Service QProfileActivityService that would only do searchActivities(). And this service would be used in this action, and in the deprecated RubyQProfileActivityService.
Or you could even not create a new service and put the code of the searchActivities(). (It would create a duplication, but as soon as RubyQProfileActivityService is removed, then the code from QProfileService could be removed also).

It would prevent the usage of a ServerTester !

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileChangelogAction.java
```diff
@@ -0,0 +1,142 @@
...
+    json.prop(Param.PAGE_SIZE, paging.pageSize());
+    json.name("events").beginArray();
+    for (QProfileActivity event : result.getHits()) {
+      json.beginObject()
+        .prop("date", DateUtils.formatDateTime(event.getCreatedAt()))
+        .prop("authorLogin", event.getLogin())
+        .prop("authorName", event.authorName())
+        .prop("action", event.getAction())
+        .prop("ruleKey", event.ruleKey().toString())
+        .prop("ruleName", event.ruleName());
```
julienlancelot: Open question : should we repeat each time the authorName and ruleName, or should we set them in separate repositories (like /api/issues/search) ?

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileChangelogAction.java
```diff
@@ -0,0 +1,142 @@
...
+      .setHandler(this)
+      .setResponseExample(getClass().getResource("example-changelog.json"));
+
+    QProfileIdentificationParamUtils.defineProfileParams(changelog, languages);
+
+    changelog.addPagingParams(50);
+
+    changelog.createParam(PARAM_SINCE)
+      .setDescription("Start date for the changelog.")
+      .setExampleValue("2011-04-25");
```
julienlancelot: Are DateTime allowed ?

server/sonar-server/src/test/java/org/sonar/server/qualityprofile/ws/QProfileChangelogActionMediumTest.java
```diff
@@ -0,0 +1,141 @@
...
+import org.sonar.server.rule.RuleTesting;
+import org.sonar.server.tester.ServerTester;
+import org.sonar.server.user.MockUserSession;
+import org.sonar.server.ws.WsTester;
+
+import java.util.Date;
+
+import static org.sonar.server.qualityprofile.QProfileTesting.XOO_P1_KEY;
+
+public class QProfileChangelogActionMediumTest {
```
julienlancelot: I would add a test of updating a rule without param, and another one of updating a rule with no user (System user, as it's done at server startup)

server/sonar-server/src/test/java/org/sonar/server/qualityprofile/ws/QProfileChangelogActionMediumTest.java
```diff
@@ -0,0 +1,141 @@
...
+      .execute().assertJson(getClass(), "changelog_nominal.json");
+    wsTester.newGetRequest(QProfilesWs.API_ENDPOINT, "changelog").setParam("profileKey", XOO_P1_KEY).setParam("since", DateUtils.formatDate(tomorrow))
+      .execute().assertJson(getClass(), "changelog_empty.json");
+
+    // Tests with "to"
+    wsTester.newGetRequest(QProfilesWs.API_ENDPOINT, "changelog").setParam("profileKey", XOO_P1_KEY).setParam("to", DateUtils.formatDate(yesterday))
+      .execute().assertJson(getClass(), "changelog_empty.json");
+    wsTester.newGetRequest(QProfilesWs.API_ENDPOINT, "changelog").setParam("profileKey", XOO_P1_KEY).setParam("to", DateUtils.formatDate(tomorrow))
+      .execute().assertJson(getClass(), "changelog_nominal.json");
+  }
```
julienlancelot: I would add a test with both "from" and "to" params.

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileChangelogAction.java
jblievremont: Reading the Rails code, it seemed that it was only date, but looking at [this thread on the user ML](http://sonarqube.15.x6.nabble.com/Link-to-quality-profiles-changes-doesn-t-work-in-the-Events-widget-tt5033551.html), it seems that indeed datetimes are allowed...

julienlancelot: I think it should be a good idea to allow DateTime

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileChangelogAction.java
jblievremont: :+1:

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileChangelogAction.java
jblievremont: It's easier to parse on client side, it's only author login + name, and pagination can help to reduce the size if needed. So let's keep it like this.

jblievremont:
Applied some feedback.

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileChangelogAction.java
julienlancelot: ok

server/sonar-server/src/test/java/org/sonar/server/qualityprofile/ws/QProfileChangelogActionTest.java
```diff
@@ -0,0 +1,205 @@
...
+import static org.mockito.Mockito.mock;
+import static org.sonar.server.qualityprofile.QProfileTesting.XOO_P1_KEY;
+
+public class QProfileChangelogActionTest {
+
+  @ClassRule
+  public static DbTester dbTester = new DbTester();
+
+  @ClassRule
+  public static EsTester esTester = new EsTester().addDefinitions(new ActivityIndexDefinition(new Settings()));
```
julienlancelot: Oh yes !!! :100:

julienlancelot:
Good job, LGTM !

PR #209, opened 2015-04-10
SONAR-6258 Persist symbols into file sources

julienlancelot:
SONAR-6258 Persist symbols into file sources


GithubIntegration:
SonarQube analysis reported 19 new issues:
* 1 blocking
* 12 major
* 1 minor
* 5 info

GithubIntegration:
* 3 major
* 1 info

server/sonar-server/src/main/java/org/sonar/server/computation/source/SymbolsLineReader.java
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+    return lineSymbols;
+  }
+
+  private static boolean matchLine(BatchReport.Range range, int line) {
+    return range.getStartLine() <= line && range.getEndLine() >= line;
+  }
+
+  private void sortSymbols() {
```
teryk: where can I find the old way or sorting ?

sonar-batch/src/test/java/org/sonar/batch/index/SourceDataFactoryTest.java
```diff
@@ -364,6 +364,46 @@ public void applySymbolReferences() throws Exception {
   }
 
   @Test
+  public void applySymbolReferences2() throws Exception {
```
teryk: would it be possible to have more meaningfull name for this method and the next ?

teryk:
any idea about the new sq issues ? 
I think it would important to have some performance test before persisting file_sources on server side (just a memo).

teryk:
apply minor feedback and LGTM !

server/sonar-server/src/main/java/org/sonar/server/computation/source/SymbolsLineReader.java
julienlancelot: What do you mean ?

sonar-batch/src/test/java/org/sonar/batch/index/SourceDataFactoryTest.java
julienlancelot: Good point !

julienlancelot:
For SQ issues, as I've rebased from the master, I'll have a look.
For perf test, it will be indeed needed to do it.

server/sonar-server/src/main/java/org/sonar/server/computation/source/SymbolsLineReader.java
teryk: I wanted to compare with batch way of sorting to check if it's the same ordering

julienlancelot:
@teryk Here is where the batch sort symbols : https://github.com/SonarSource/sonarqube/blob/master/sonar-batch/src/main/java/org/sonar/batch/index/SourceDataFactory.java#L250

teryk:
thanks, so I confirm my LGTM

PR #210, opened 2015-04-10
Update environment.rb for Oracle optimizations

sberthiaume:
Instead of using a SELECT to fetch sequence #, assign value directly from the sequence; it would lower cursor usage for people with tight DB restrictions and could improve performance by up to 50% based on this: http://oracle-base.com/articles/11g/plsql-new-features-and-enhancements-11gr1.php#sequences_in_plsql_expressions.

However it would only be compatible with Oracle 11g and up, but Oracle 10g (even R2) was dropped from even extended support for almost 2 years. If support for Oracle 10g must be kept, perhaps a support for specific dialects like "Oracle9g", "Oracle10g", "Oracle11g" and "Oracle12c" would be necessary (a bit like Hibernate does although they stopped at 10g).

### DIFF
--- a/server/sonar-web/src/main/webapp/WEB-INF/config/environment.rb
+++ b/server/sonar-web/src/main/webapp/WEB-INF/config/environment.rb
@@ -208,7 +208,7 @@ class ActiveRecord::Migration
           FOR EACH ROW
         BEGIN
            IF :new.id IS null THEN
-             SELECT #{table}_seq.nextval INTO :new.id FROM dual;
+             :new.id := #{table}_seq.nextval;
            END IF;
         END;})
   end


simonbrandhof:
In which web page this optimization makes sense ? Almost all INSERT requests are executed through Java app but not from Rails. 
Note that Java code does not call sequences. A trigger is defined on db-side to generate ids.

simonbrandhof:
Moreover you mentioned Hibernate. Note that it is being dropped from version 5.2.

simonbrandhof:
And Ruby on Rails will also being dropped at medium term (5.x LTS or 6.x) . So please reopen this pull request if you have a valid use-case at short term. Thanks.

sberthiaume:
To answer your questions (in order):
 1- This Ruby code is executed when the Oracle DB tables are created (and perhaps upgraded); it creates triggers that is later called by any SQL statement (from any framework);
 2- I mentioned Hibernate only as an example of multiple supported Oracle dialect, wasn't even sure which persistence layer you were using;
 3- My fix is actually a PL/SQL trigger fix, it just happens that the Ruby code is the one creating the triggers; if you were to create the triggers in pure Java, the fix would conceptually still apply.

One of the reason I'm proposing this fix is that our DBA impose tight control over the number of cursors a session can have opened and it just happens that the way the triggers are coded could double that trigger usage. In Java, whenever you create a Statement or PreparedStatement, it creates a Cursor in Oracle and, if I'm not mistaken, you are using DBCP to do some parallel SQL processing which can open MANY such statements; for each inserted record, because of the trigger, Oracle will open a 2nd cursor for the SELECT to fetch the sequence number thus possibly doubling the number of opened cursor in parallel processing.

My suggested fix fetches the sequence value directly, without doing a SELECT which is both 50% faster and doesn't require a 2nd cursor; it's something only supported since Oracle 11g (which is already many years old).

simonbrandhof:
Ok, I didn't see that the trigger code was touched, not the client part of Rails code.
We will discuss internally about the support of Oracle 10g then I'll get back to you.
Thanks

sberthiaume:
No problem, if you already have a branch with the new non-ruby code creating the triggers, I'll be more than happy to take a look at it as well if you don't mind dropping Oracle 10g support of find a way to support both.

simonbrandhof:
JIRA ticket created : https://jira.codehaus.org/browse/SONAR-6421.
For my information is it possible that the trigger supports both versions at the same time (something like "if 10g else") ?

sberthiaume:
Doing the check IN the trigger would require a SELECT which is counter-productive for what we are trying to do. However, it would be possible to have the database creation Ruby code detect the version and create the triggers with the right format depending on that version. The only possible downside to that idea would be if one migrates his database from 10g to 11g+ he would still have the old format, but my guess is that this is a fringe case.

I'm by no mean a Ruby programmer, never even edited a Ruby file before this one, but I could try to update my fork to include that logic and send another pull request.

Interested?

simonbrandhof:
Not interested for a good reason: Ruby on Rails is being dropped. That's the reason why I linked the ticket to https://jira.codehaus.org/browse/SONAR-5471.
A simple solution to support oracle upgrades (from 10g to 11g for instance) is to automatically update all triggers during a SonarQube server startup.

sberthiaume:
No problem, conceptually speaking, in whatever language this ends-up being, you can get the Oracle version with the following SQL: "SELECT version FROM product_component_version WHERE product LIKE 'Oracle Database %';"; it should be a 5-number versioning scheme like "11.2.0.1.0" (in this case indicating 11gR2), From there the triggers could be created based on version optimizations. This query worked at least since Oracle 9i (support stopped almost 5 years ago).

The only benefit I see of doing it in Ruby would be to validate the concept before migrating it to new language/framework.

simonbrandhof:
This pull request won't be applied as this. Trigger will be improved for sure when moving to Java-based migrations. I completed https://jira.codehaus.org/browse/SONAR-6421 and now close this request. Thanks for the contribution.

simonbrandhof:
FYI we decided to stop the support of Oracle 10g in SonarQube 5.2. : http://jira.codehaus.org/browse/SONAR-6434

PR #213, opened 2015-04-14
add -D option to start.sh to remote debug web server

sns-seb:
this PR adds support for a `-D` option to start the webserver with remote debugging enabled on port 5005

this PR needs to be reviewed by someone with a MAC to make sure the script works correctly on this platform

thanks

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,9 +1,14 @@
 #!/bin/bash
+###############################
+# usage: use -D option to enable remote debugging of the web server on port 5005
+###############################
 
 if [[ "$OSTYPE" == "darwin"* ]]; then
   OS='macosx-universal-64'
+  SED_DISABLE_BACKUP=" ''"
 else
   OS='linux-x86-64'
+  SED_DISABLE_BACKUP=""
 fi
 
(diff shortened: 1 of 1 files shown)


GithubIntegration:
SonarQube analysis reported 2 new issues:
* 1 major
* 1 info

simonbrandhof:
Using -D is quite misleading for a Java guy. Don't you think that --debugPort 5005 would be more intuitive ? Note that I don't know the cost of implementation !

sns-seb:
I meant to have a very short option to go into debug mode, hence the one letter option, and did not mean to support custom port in this PR.

I could go with `-d` if it is less confusing (implementation cost is 0).

With time in hand, I would use standard command options:

* `-d` and `--debug` would do the same (enable debug port on web server)
* if either is followed by a value, that would be the port and would overwrite the default (5005)

Do we have time for this ?

simonbrandhof:
Not sure about the ROI of the second proposal. So let's go with -d for now.
LGTM, it works on Mac.

sns-seb:
changed script to use `-d` instead of `-D`

rebased on master and merged

PR #215, opened 2015-04-15
Move tests persistence from batch to compute - SONAR-6255

teryk:
3 parts:
- define index and check the 3 use cases
- define batch report format
- create file_sources.test_data

You might want to review the 1st commit which does not include db unit files updates.


server/sonar-server/src/main/java/org/sonar/server/test/index/TestDoc.java
```diff
@@ -0,0 +1,104 @@
...
+  public String type() {
+    return getField(FIELD_TYPE);
+  }
+
+  public TestDoc setType(String type) {
+    setField(FIELD_TYPE, type);
+    return this;
+  }
+
+  public List<Map<String, Object>> coverageBlocks() {
```
julienlancelot: Why Object ?

server/sonar-server/src/test/java/org/sonar/server/test/index/TestIndexTest.java
```diff
@@ -0,0 +1,111 @@
...
+  TestIndex sut = new TestIndex(es.client());
+
+  @Before
+  public void before() throws Exception {
+    es.truncateIndices();
+  }
+
+  @Test
+  public void lines_covered_a_test_method() throws Exception {
+    List<Map<String, Object>> coverageBlocks = new ArrayList<>();
```
julienlancelot: This list is useless (same for next tests)

server/sonar-server/src/test/java/org/sonar/server/test/index/TestIndexTest.java
```diff
@@ -0,0 +1,111 @@
...
+    coverageBlocks.add(newCoverageBlock("2"));
+    coverageBlocks.add(newCoverageBlock("3"));
+
+    es.putDocuments(TestIndexDefinition.INDEX, TestIndexDefinition.TYPE,
+      newTestDoc("1", newCoverageBlock("3"), newCoverageBlock("4"), newCoverageBlock("5")),
+      newTestDoc("2", newCoverageBlock("5"), newCoverageBlock("6"), newCoverageBlock("7")));
+
+    List<Map<String, Object>> result = sut.coveredLines("uuid-1", "name-1");
+
+    assertThat(result).hasSize(3);
```
julienlancelot: You should check the content, not only the length

server/sonar-server/src/main/java/org/sonar/server/test/index/TestDoc.java
julienlancelot: From what I understand, this Map<String, Object> is the inner object. You should create a new Doc for that and return a List<Doc> instead of a List<Map>

server/sonar-server/src/main/java/org/sonar/server/es/NewIndex.java
```diff
@@ -97,7 +100,7 @@ public NewIndexType createLongField(String fieldName) {
     }
 
     public NewIndexType createDynamicNestedField(String fieldName) {
-      return setProperty(fieldName, ImmutableMap.of("type", "nested", "dynamic", "true"));
+      return setProperty(fieldName, ImmutableMap.of("type", "nested", "dynamic", "true", "include_in_parent", "true"));
```
julienlancelot: For my information, what is this new parameter ?

server/sonar-server/src/main/java/org/sonar/server/test/index/TestDoc.java
julienlancelot: Ok, this will be done in a next PR, nothing to do here for the moment.

server/sonar-server/src/test/java/org/sonar/server/test/index/TestIndexTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+    List<TestDoc> result = sut.testsCovering("main-uuid-5", 82);
+
+    assertThat(result).hasSize(2);
+  }
+
+  private Map<String, Object> newCoverageBlock(String id) {
+    Map<String, Object> coverageBlock = new HashMap<>();
+    coverageBlock.put("key", "project:file-" + id);
+    coverageBlock.put("uuid", "main-uuid-"+ Integer.valueOf(id));
```
julienlancelot: No need to use Integer.valueOf as id is already a String

server/sonar-server/src/test/java/org/sonar/server/test/index/TestIndexTest.java
```diff
@@ -0,0 +1,111 @@
...
+    coverageBlock.put("coveredLines", 42);
```
julienlancelot: If my understanding is correct, coveredLines is the number of line covered by the test. As you've added only 3 lines in the 'lines'  parameter, you should set 3 in 'coveredLines'.
Do not forget that tests are also used to understand what is doing the class !

server/sonar-server/src/test/java/org/sonar/server/test/index/TestIndexTest.java
```diff
@@ -0,0 +1,111 @@
...
+  public void before() throws Exception {
+    es.truncateIndices();
+  }
+
+  @Test
+  public void lines_covered_a_test_method() throws Exception {
+    List<Map<String, Object>> coverageBlocks = new ArrayList<>();
+    coverageBlocks.add(newCoverageBlock("1"));
+    coverageBlocks.add(newCoverageBlock("2"));
+    coverageBlocks.add(newCoverageBlock("3"));
```
julienlancelot: I In fact the these 4 lines are useless, please remove them (same for next tests)

server/sonar-server/src/main/java/org/sonar/server/test/index/TestIndexDefinition.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @Override
+  public void define(IndexDefinitionContext context) {
+    NewIndex index = context.create(INDEX);
+
+    EsUtils.refreshHandledByIndexer(index);
+    EsUtils.setShards(index, settings);
+
+    NewIndex.NewIndexType nestedMapping = index.createType(TYPE);
+    nestedMapping.stringFieldBuilder(FIELD_COVERAGE_BLOCK_KEY).build();
```
julienlancelot: What is this key ? If this is the key of the component, it should not be inserted in E/S as keys can be changed.

sonar-batch-protocol/src/main/protobuf/file_source_test_db.proto
```diff
@@ -0,0 +1,59 @@
...
+// structure of db column FILE_SOURCES.TEST_DATA
+
+// Temporarily in sonar-batch-protocol
+
+package org.sonar.server.source.db;
+option optimize_for = SPEED;
+
+message Test {
+  optional string uuid = 1;
+  optional string key = 2;
```
julienlancelot: Key should not be used here as it can be changed

sonar-batch-protocol/src/main/protobuf/file_source_test_db.proto
```diff
@@ -0,0 +1,59 @@
...
+  optional string method_name = 3;
+  optional string status = 4;
+  optional string test_message = 5;
+  optional string type = 6;
+  repeated CoverageBlock coverage_block = 7;
+
+  message CoverageBlock {
+    optional string uuid = 1;
+    optional string key = 2;
```
julienlancelot: Key should not be used here as it can be changed

sonar-batch-protocol/src/main/protobuf/file_source_test_db.proto
```diff
@@ -0,0 +1,59 @@
...
+    optional string long_name = 3;
+    repeated int32 lines = 4;
+    optional int32 nb_covered_lines = 5;
+  }
+}
+
+message Tests {
```
julienlancelot: Shouldn't we remove this to allow doing some streaming in the futur ?

julienlancelot:
Please apply my feedback

server/sonar-server/src/main/java/org/sonar/server/test/index/TestDoc.java
teryk: I add a todo for the time being

server/sonar-server/src/main/java/org/sonar/server/es/NewIndex.java
teryk: Something I used during my tests, I remove it. It changes how ES index my nested type. It includes nested documents in the parent document. For example, I could do a query referencing coverageBlocks.uuid directly, without using a nested query.
More information on [elastic search website](http://www.elastic.co/guide/en/elasticsearch/reference/1.x/mapping-nested-type.html)

server/sonar-server/src/test/java/org/sonar/server/test/index/TestIndexTest.java
teryk: yes, covered lines is removed as it's the number of lines.

server/sonar-server/src/main/java/org/sonar/server/test/index/TestIndexDefinition.java
teryk: I thought we'd go with a denomalized schema. It means I'll have to request component to get the key then ?

sonar-batch-protocol/src/main/protobuf/file_source_test_db.proto
teryk: it will be indeed be removed. It's already done on my current development, working on feeding the database

sonar-batch-protocol/src/main/protobuf/file_source_test_db.proto
teryk: ok. the model has already changed in fact. I suggest you review this part thouroughly when database feeding is done

teryk:
please review 5946ddce3fc62e53a0de04dd2ca22fec59770b4b

server/sonar-server/src/main/java/org/sonar/server/es/NewIndex.java
julienlancelot: And in PROD, what will be the value ?

server/sonar-server/src/main/java/org/sonar/server/test/index/TestIndexDefinition.java
julienlancelot: Indeed, but as the component key can change we don't want to reindex all indexes using component keys each time a key is updated

julienlancelot:
LGTM (but next PR should remove usage of component keys)

server/sonar-server/src/main/java/org/sonar/server/es/NewIndex.java
SonarTech: :new: MINORDefine a constant instead of duplicating this literal "nested" 3 times.[](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1192)

SonarTech:
SonarQube analysis reported 2 new issues:
* 2 minor

server/sonar-server/src/main/java/org/sonar/server/es/EsUtils.java
```diff
@@ -79,4 +82,17 @@ public static String formatDateTime(@Nullable Date date) {
     }
     return null;
   }
+
+  public static void setShards(NewIndex index, Settings settings) {
```
simonbrandhof: I don't think this should be extracted in "yet another util method". Why not moving it to NewIndex ?

teryk: Good idea. It's a better alternative than EsUtils. What I don't want is to repeat code in several places.

simonbrandhof: Exact, this code is too much duplicated.

PR #216, opened 2015-04-15
SONAR-6376 add Java Ws /api/plugins/installed

sns-seb:
provides the same data as RoR WS /api/updatecenter/installed_plugins but only in JSON format

another PR will make the new WS return more info about each plugin

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
@@ -1,0 +1,125 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * SonarQube is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.plugins.ws;
+
+import com.google.common.base.Function;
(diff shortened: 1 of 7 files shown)


server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
```diff
@@ -0,0 +1,110 @@
...
+      return !input.isCore();
+    }
+  }
+
+  private enum PluginMetadataNameComparator implements Comparator<PluginMetadata> {
+    INSTANCE;
+
+    @Override
+    public int compare(PluginMetadata o1, PluginMetadata o2) {
+      return o1.getName().compareTo(o2.getName());
```
jblievremont: Note: this is already the natural order for DefaultPluginMetadata (which is the only implementation of PluginMetadata).

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public void handle(Request request, Response response) throws Exception {
+    List<PluginMetadata> pluginMetadatas = retrieveAndSortPluginMetadata();
+    writeMetadataList(response, pluginMetadatas);
+  }
+
+  private List<PluginMetadata> retrieveAndSortPluginMetadata() {
+    List<PluginMetadata> pluginMetadatas = newArrayList(filter(pluginRepository.getMetadata(), NotCorePluginsPredicate.INSTANCE));
+    sort(pluginMetadatas, PluginMetadataNameComparator.INSTANCE);
```
jblievremont: You could use a SortedSet here, to avoid explicit sort.

jblievremont:
Some very minor feedback, else LGTM :+1:!

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
sns-seb: the sort is now part of the WS contract, it should therefor be enforced by the WS implementation (and this order is not part of the method `getMetaData`'s contract so it is weak to rely on it)

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
sns-seb: using a SortedSet instead of a list will have an impact: Plugins with the same name but, for example, a different key will be considered equal and only one will appear in the JSON response.

@jblievremont, are plugin names unique ?

sns-seb:
applied feedback on using a Set + added subsorting on key to avoid losing plugin if ever two have the same name and added unit tests for that

merged

PR #217, opened 2015-04-15
Add WS for the quality profiles "permalinks" section

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/qualityprofile/QProfileBackuper.java
```diff
@@ -47,6 +47,8 @@
 
 public class QProfileBackuper implements ServerComponent {
 
+  public static final String MEDIA_TYPE_XML = "text/xml";
```
simonbrandhof: Can be replaced by org.sonar.server.plugins.MimeTypes#XML. Note that this class could be moved in an appropriate package and that value is different ("application/xml"). It seems to have the same behavior than "text/xml" since HTTP 1.1.

jblievremont: Actually, I was expecting to be able to use Guava's MediaType class, but it's only available since Guava 12 :stuck_out_tongue_winking_eye:

simonbrandhof:
WS api/qualityprofiles/exporters is not available

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileExportersAction.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Override
+  public void handle(Request request, Response response) throws Exception {
+    JsonWriter json = response.newJsonWriter().beginObject().name("exporters").beginArray();
+    for (ProfileExporter exporter : exporters) {
+      json.beginObject()
+        .prop("key", exporter.getKey())
+        .prop("name", exporter.getName());
+      json.name("languages").beginArray();
+      for (String language : exporter.getSupportedLanguages()) {
```
simonbrandhof: just for information this loop can be inlined in something like `json.values(Arrays.asList(exporter.getSupportedLanguages()));`

server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ws/QProfileExportersAction.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public QProfileExportersAction() {
+    this(new ProfileExporter[0]);
+  }
+
+  @Override
+  public void define(NewController context) {
+    context.createAction("exporters")
+      .setDescription("Lists available profile export formats.")
```
simonbrandhof: I'd mention the WS api/qualityprofiles/export that uses these exporters

server/sonar-server/src/test/java/org/sonar/server/qualityprofile/ws/QProfileExportActionTest.java
```diff
@@ -0,0 +1,198 @@
...
+  DbClient dbClient;
+
+  DbSession session;
+
+  QProfileBackuper backuper;
+
+  QProfileExporters exporters;
+
+  @Before
+  public void before() throws Exception {
```
simonbrandhof: this is definitely a pain to test a WS !

simonbrandhof:
LGTM

PR #218, opened 2015-04-15
SONAR-6376 better Java WS to list installed plugins

sns-seb:
This PR : 
* changes the WS response to use a root object so that the response can be easily extended in the future (best practice)
* adds more properties to the WS response for each plugin

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
@@ -34,18 +34,24 @@ import org.sonar.api.utils.text.JsonWriter;
 import java.util.Collection;
 import java.util.SortedSet;
 
-import static com.google.common.base.Objects.firstNonNull;
 import static com.google.common.collect.Iterables.filter;
 
 /**
  * Implementation of the {@code installed} action for the Plugins WebService.
  */
 public class InstalledPluginsWsAction implements PluginsWsAction {
-  private static final String DEFAULT_VERSION = "-";
   private static final String PROPERTY_KEY = "key";
   private static final String PROPERTY_NAME = "name";
+  private static final String PROPERTY_DESCRIPTION = "description";
+  private static final String PROPERTY_LICENSE = "license";
   private static final String PROPERTY_VERSION = "version";
   private static final String ARRAY_PLUGINS = "plugins";
+  private static final String PROPERTY_ORGANIZATION_NAME = "organizationName";
+  private static final String PROPERTY_ORGANIZATION_URL = "organizationUrl";
+  private static final String OBJECT_URLS = "urls";
+  private static final String PROPERTY_HOMEPAGE = "homepage";
+  private static final String PROPERTY_ISSUE_TRACKER = "issueTracker";
+  private static final String OBJECT_ARTIFACT = "artifact";
 
   private final PluginRepository pluginRepository;
 
@@ -71,6 +77,7 @@ public class InstalledPluginsWsAction implements PluginsWsAction {
     Collection<PluginMetadata> pluginMetadatas = retrieveAndSortPluginMetadata();
 
     JsonWriter jsonWriter = response.newJsonWriter();
+    jsonWriter.setSerializeEmptys(false);
     jsonWriter.beginObject();
 
     writeMetadataList(jsonWriter, pluginMetadatas);
@@ -90,16 +97,49 @@ public class InstalledPluginsWsAction implements PluginsWsAction {
       jsonWriter.name(ARRAY_PLUGINS);
       jsonWriter.beginArray();
       for (PluginMetadata pluginMetadata : pluginMetadatas) {
-        writeMetadata(jsonWriter, pluginMetadata);
+        writePluginMetadata(jsonWriter, pluginMetadata);
       }
       jsonWriter.endArray();
   }
 
-  private void writeMetadata(JsonWriter jsonWriter, PluginMetadata pluginMetadata) {
+  private void writePluginMetadata(JsonWriter jsonWriter, PluginMetadata pluginMetadata) {
     jsonWriter.beginObject();
+
+    writeMetadata(jsonWriter, pluginMetadata);
+
+    writeUrls(jsonWriter, pluginMetadata);
+
+    writeArtifact(jsonWriter, pluginMetadata);
+
+    jsonWriter.endObject();
+  }
+
+  private void writeMetadata(JsonWriter jsonWriter, PluginMetadata pluginMetadata) {
     jsonWriter.prop(PROPERTY_KEY, pluginMetadata.getKey());
     jsonWriter.prop(PROPERTY_NAME, pluginMetadata.getName());
-    jsonWriter.prop(PROPERTY_VERSION, firstNonNull(pluginMetadata.getVersion(), DEFAULT_VERSION));
+    jsonWriter.prop(PROPERTY_DESCRIPTION, pluginMetadata.getDescription());
+    jsonWriter.prop(PROPERTY_VERSION, pluginMetadata.getVersion());
+    jsonWriter.prop(PROPERTY_LICENSE, pluginMetadata.getLicense());
+    jsonWriter.prop(PROPERTY_ORGANIZATION_NAME, pluginMetadata.getOrganization());
+    jsonWriter.prop(PROPERTY_ORGANIZATION_URL, pluginMetadata.getOrganizationUrl());
+  }
+
+  private void writeUrls(JsonWriter jsonWriter, PluginMetadata pluginMetadata) {
+    jsonWriter.name(OBJECT_URLS);
+    jsonWriter.beginObject();
+    jsonWriter.prop(PROPERTY_HOMEPAGE, pluginMetadata.getHomepage());
+    jsonWriter.prop(PROPERTY_ISSUE_TRACKER, pluginMetadata.getIssueTrackerUrl());
+    jsonWriter.endObject();
+  }
+
+  private void writeArtifact(JsonWriter jsonWriter, PluginMetadata pluginMetadata) {
+    if (pluginMetadata.getFile() == null) {
+      return;
+    }
+
+    jsonWriter.name(OBJECT_ARTIFACT);
+    jsonWriter.beginObject();
+    jsonWriter.prop(PROPERTY_NAME, pluginMetadata.getFile().getName());
     jsonWriter.endObject();
   }
 
--- a/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
@@ -45,6 +45,7 @@ public class InstalledPluginsWsAction implements PluginsWsAction {
   private static final String PROPERTY_KEY = "key";
   private static final String PROPERTY_NAME = "name";
   private static final String PROPERTY_VERSION = "version";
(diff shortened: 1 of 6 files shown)


server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
```diff
@@ -78,21 +93,44 @@ public void handle(Request request, Response response) throws Exception {
...
+
+    writeUrls(jsonWriter, pluginMetadata);
+
+    jsonWriter.endObject();
+  }
+
+  private void writeOrganisation(JsonWriter jsonWriter, PluginMetadata pluginMetadata) {
+    jsonWriter.name(OBJECT_ORGANIZATION);
+    jsonWriter.beginObject();
+    jsonWriter.prop(PROPERTY_NAME, pluginMetadata.getOrganization());
```
simonbrandhof: I don't know why, but many fields of pluginMetadata (organization, org url, urls, license, ...) do not return null but empty strings when missing. For this reason the JSON is quite strange: 
```
{
  "key": "uiplugin",
  "name": "SonarQube Integration Tests :: Plugins :: UI",
  "description": "Main plugin for UI tests",
  "version": "1.0-SNAPSHOT",
  "license": "",
  "organization": {
    "name": "",
    "url": ""
},
"urls": {
  "homepage": "",
  "issueTracker": ""
}
}
```

server/sonar-server/src/test/java/org/sonar/server/plugins/ws/InstalledPluginsWsActionTest.java
```diff
@@ -87,84 +90,117 @@ public void core_plugin_are_not_returned() throws Exception {
...
+        "  [" +
+        "    {\"key\": \"A\"}" +
+        "  ]" +
+        "}"
+      );
     assertThat(response.outputAsString()).containsOnlyOnce("name2");
   }
 
   @Test
   public void dash_is_returned_when_version_is_null() throws Exception {
```
simonbrandhof: why ? This is probably a trick to be done for UI in webapp. I would not expect the WS to replace missing values by dash.

simonbrandhof:
The ruby WS api/updatecenter/installed_plugins is still implemented and available. It should be dropped.

server/sonar-server/src/test/java/org/sonar/server/plugins/ws/InstalledPluginsWsActionTest.java
sns-seb: according to Ruby code, `-` was output in JSON for the version.

This can easily be droped. I agree is does not make sense unless we want to implement a kink of mandatory field.

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
sns-seb: If this is really a general guideline, this should be implemented at the `JSONWriter` level and enabled by default, WDYT ?

But I don't see any support for this in `com.google.gson.stream.JsonWriter` as there is for `nulls` :(

sns-seb:
droping a ROR webservice requires to answer the question of its support as deprecated service and it has been decided with Fabrice that a specific JIRA will be created for each old WS

server/sonar-server/src/test/java/org/sonar/server/plugins/ws/InstalledPluginsWsActionTest.java
simonbrandhof: I'm not for keeping such tricks. Let's keep clean WS !

simonbrandhof:
Decided with @fmallet the opposite: please remove the Ruby on Rails WS. No intermediary deprecated status.

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
simonbrandhof: The endless question about null versus empty strings. JsonWriter should still support the writing of empty strings, so I think about 2 options:
1. add a property JsonWriter#setIgnoreEmptyStrings(boolean). Default is false. The pb is that it will be applied to all field values.
2. convert empty to null in InstalledPluginsWsAction

I tend to prefer second option as long as there are no other use-cases.

sns-seb:
ok

I will remove the old WS in a separate JIRA and PR for SOC though

sns-seb:
@simonbrandhof updated PR:

* remove the "-" in version
* add support to disable serialization of empty String to JSON in JsonWriter (review welcome on this one)
* no more fields with empty string are displayed in JSON

simonbrandhof:
No need for a separate JIRA ticket for removal of old ruby WS.

sonar-plugin-api/src/main/java/org/sonar/api/utils/text/JsonWriter.java
```diff
@@ -48,17 +49,24 @@
  *     .endObject()
  *     .close();
  * </pre>
+ * </p>
+ * <p>
+ * By default, emptry strings are serialized. To disable empty string serialization,
```
simonbrandhof: typo "emptry"

sonar-plugin-api/src/main/java/org/sonar/api/utils/text/JsonWriter.java
```diff
@@ -48,17 +49,24 @@
  *     .endObject()
  *     .close();
  * </pre>
+ * </p>
```
simonbrandhof: my bad, javadoc on nulls is missing. Could you please add "<p>By default, null values are not serialized. Use {@link #setSerializeNulls(true)} to enable writing of null values.</p>" ? 
Thanks

sonar-plugin-api/src/main/java/org/sonar/api/utils/text/JsonWriter.java
```diff
@@ -22,15 +22,16 @@
 import org.sonar.api.utils.DateUtils;
 
 import javax.annotation.Nullable;
-
 import java.io.Writer;
 import java.util.Date;
 import java.util.Map;
 
+import static com.google.common.base.Strings.emptyToNull;
```
simonbrandhof: please do not use guava in sonar-plugin-api. I'm working on remove it ! Thanks

sonar-plugin-api/src/test/java/org/sonar/api/utils/text/JsonWriterTest.java
```diff
@@ -39,6 +39,8 @@
 
 public class JsonWriterTest {
 
+  private static final String EMPTY_STRING = "";
```
simonbrandhof: Minor - you can use StringUtils.EMPTY from commons-lang. Having such dependencies is accepted in tests.

sonar-plugin-api/src/test/java/org/sonar/api/utils/text/JsonWriterTest.java
```diff
@@ -103,13 +105,13 @@ public void type_of_values() throws Exception {
...
   public void ignore_null_values_by_default() throws Exception {
     writer.beginObject()
-      .prop("nullNumber", (Number) null)
-      .prop("nullString", (String) null)
-      .name("nullNumber").value((Number) null)
-      .name("nullString").value((String) null)
-      .name("nullDate").valueDate(null)
-      .name("nullDateTime").valueDate(null)
-      .endObject().close();
+        .prop("nullNumber", (Number) null)
```
simonbrandhof: it seems that you don't use https://github.com/SonarSource/sonar-developer-toolset/tree/master/eclipse

sonar-plugin-api/src/main/java/org/sonar/api/utils/text/JsonWriter.java
sns-seb: done

sonar-plugin-api/src/main/java/org/sonar/api/utils/text/JsonWriter.java
sns-seb: done

sonar-plugin-api/src/test/java/org/sonar/api/utils/text/JsonWriterTest.java
sns-seb: I do, but IDEA default formatting and eclipse do not work well together, so I need to do explicit formatting and I guess I forgot once

sns-seb:
PR updated according to comments

sns-seb:
just added a commit to the PR to add the name of the jar to the JSON response for consistency with other plugin WebServices and also, as this information can be useful to match listed plugins to the content of `extensions/plugins` directory

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
```diff
@@ -78,21 +94,61 @@ public void handle(Request request, Response response) throws Exception {
...
+  private void writeArchive(JsonWriter jsonWriter, PluginMetadata pluginMetadata) {
+    if (pluginMetadata.getFile() == null) {
+      return;
     }
-    jsonWriter.endArray();
-    jsonWriter.close();
+
+    jsonWriter.name(OBJECT_ARCHIVE);
+    jsonWriter.beginObject();
+    jsonWriter.prop(PROPERTY_NAME, pluginMetadata.getFile().getName());
```
simonbrandhof: having a hierarchy only for one field is useless. Please use flat structure. If "archive" is there for future changes, then it's a bad reason !

sns-seb: "archive" is there because the new WS is supposed to return more information that it currently can with the underlying components, see my email about that

server/sonar-server/src/main/resources/org/sonar/server/plugins/ws/example-installed_plugins.json
```diff
@@ -1,17 +1,64 @@
...
+      "version": "3.2",
+      "license": "GNU LGPL 3",
+      "organization": {
+        "name": "SonarSource",
+        "url": "http://www.sonarsource.com"
+      },
+      "urls": {
+        "homepage": "http://redirect.sonarsource.com/plugins/findbugs.html",
+        "issueTracker": "http://jira.codehaus.org/browse/SONARFBUGS"
+      }
```
simonbrandhof: archive information are missing in the example.

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
simonbrandhof: ok, so to be consistent with update center project, please rename it "artifact"

simonbrandhof:
LGTM if "archive" is renamed "artifact"

server/sonar-server/src/main/resources/org/sonar/server/plugins/ws/example-installed_plugins.json
sns-seb: good catch

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstalledPluginsWsAction.java
sns-seb: > "archive" is there because the new WS is supposed to return more information that it currently can with the underlying components, see my email about that

this comment applied to the PR for WS listing pending installation and removal of plugins

nevertheless, the archive object will be used across all plugin Webservice but either under plugin or release

PR #219, opened 2015-04-15
SONAR-6364 Line number should be unset on closed issues

simonbrandhof:
SONAR-6364 Line number should be unset on closed issues


henryju:
LGTM

PR #222, opened 2015-04-16
SONAR-6406 Add active severity facet on rules WS

jblievremont:


server/sonar-server/src/test/java/org/sonar/server/rule/index/RuleIndexMediumTest.java
```diff
@@ -827,6 +827,45 @@ public void search_by_profile_and_inheritance() throws InterruptedException {
   }
 
   @Test
+  public void search_by_profile_and_active_severity() throws InterruptedException {
+    QualityProfileDto qualityProfileDto1 = QProfileTesting.newXooP1();
+    QualityProfileDto qualityProfileDto2 = QProfileTesting.newXooP2();
+    db.qualityProfileDao().insert(dbSession, qualityProfileDto1, qualityProfileDto2);
+
+    RuleDto rule1 = RuleTesting.newXooX1();
```
julienlancelot: You should set the severity, it would be easier to understand which rules have their severity overridden.

julienlancelot:
Apply minor feedback, then LGTM !

PR #224, opened 2015-04-16
Remove coupling of sonar-plugin-api on Guava 10

simonbrandhof:
Remove coupling of sonar-plugin-api on Guava 10

in order to prepare upgrade to Guava 18 in core codebase


julienlancelot:
LGTM

PR #226, opened 2015-04-16
SONAR-6433 Stop supporting the concept of "group/ecosystem" of plugins

simonbrandhof:
Requires release of updatecenter 1.3 for http://jira.sonarsource.com/browse/UPC-49


sonar-core/src/main/java/org/sonar/core/plugins/DefaultPluginMetadata.java
```diff
@@ -284,6 +273,11 @@ public DefaultPluginMetadata setImplementationBuild(String implementationBuild)
   }
 
   @Override
+  public String getParent() {
```
jblievremont: also, @Deprecated ?

sonar-plugin-api/src/main/java/org/sonar/api/platform/PluginMetadata.java
```diff
@@ -57,6 +58,12 @@
 
   String getBasePlugin();
 
+  /**
+   * Always return <code>null</code> since version 5.2
+   * @deprecated in 5.2. Concept of parent relationship is removed. See https://jira.codehaus.org/browse/SONAR-6433
+   */
+  @Deprecated
+  @CheckForNull
```
jblievremont: Too bad there is no ``@AlwaysNull`` annotation :)

jblievremont:
LGTM :+1:

PR #228, opened 2015-04-17
SONAR-6377 add WS listing available plugins

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/plugins/ws/AvailablePluginsWsAction.java
```diff
@@ -0,0 +1,194 @@
...
+      jsonWriter.prop(PROPERTY_DESCRIPTION, child.getDescription());
+      jsonWriter.endObject();
+    }
+    jsonWriter.endArray();
+
+    jsonWriter.endObject();
+  }
+
+  private static String toJSon(PluginUpdate.Status status) {
+    switch (status) {
```
jblievremont: Why the conversion, instead of a straightforward .toString()?

server/sonar-server/src/test/java/org/sonar/server/plugins/ws/AvailablePluginsWsActionTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+    WebService.Action action = controller.actions().iterator().next();
+    assertThat(action.isPost()).isFalse();
+    assertThat(action.description()).isNotEmpty();
+    assertThat(action.responseExample()).isNotNull();
+  }
+
+  @Test
+  public void empty_array_is_returned_when_there_is_no_plugin_available() throws Exception {
+    underTest.handle(request, response);
```
jblievremont: Instead of an explicit call to handle, you can use wsTester.newGetRequest("api/plugins", "available").execute().assertJson(...);

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/AvailablePluginsWsAction.java
sns-seb: to avoid exposing internal data in the WS (and having a regression the day some decides the values of the enum are too long)
also, it is less verbose => JSON message is shorter => better performance on the network and the client side

server/sonar-server/src/test/java/org/sonar/server/plugins/ws/AvailablePluginsWsActionTest.java
sns-seb: I prefer calling `handle` directly to avoid invoking all the `WebServiceEngine` and `WebTester` code. Less code executing means better performance and more robust unit tests (thinks sensitivity to refactoring). These are unit tests after all.

also, calling `handle` method is typesafe when using the `"api/plugins", "available"` parameters is not

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/AvailablePluginsWsAction.java
jblievremont: I get your point. However, in its current state, the ``default`` block is not reachable. But it's a kind of defensive programming.

server/sonar-server/src/test/java/org/sonar/server/plugins/ws/AvailablePluginsWsActionTest.java
jblievremont: Well, it's probably OK with this WS since it does not have any request parameter. But for WS that have mandatory parameters, you will want to check that they fail when the parameters are not set.

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/AvailablePluginsWsAction.java
sns-seb: yes, I gave a 2 minutes though about that and gave up having coverage on that line

jblievremont:
LGTM!

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/AvailablePluginsWsAction.java
```diff
@@ -0,0 +1,194 @@
...
+    jsonWriter.beginObject();
+    jsonWriter.prop(PROPERTY_NAME, plugin.getOrganization());
+    jsonWriter.prop(PROPERTY_URL, plugin.getOrganizationUrl());
+    jsonWriter.endObject();
+  }
+
+  private void writeTermsAndConditions(JsonWriter jsonWriter, Plugin plugin) {
+    jsonWriter.name(OBJECT_TERMS_AND_CONDITIONS);
+    jsonWriter.beginObject();
+    jsonWriter.prop(PROPERTY_URL, plugin.getTermsConditionsUrl());
```
simonbrandhof: Please use flat structure as much as possible. In the current case, the field must be "termsAndConditionsUrl".

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/AvailablePluginsWsAction.java
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  private static String toJSon(PluginUpdate.Status status) {
+    switch (status) {
+      case COMPATIBLE:
+        return "COMPATIBLE";
+      case INCOMPATIBLE:
+        return "INCOMPATIBLE";
+      case REQUIRE_SONAR_UPGRADE:
+        return "SONAR_UPGRADE";
```
simonbrandhof: Prefix "sonar" is useless. Here I would use "REQUIRES_UPGRADE"

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/AvailablePluginsWsAction.java
```diff
@@ -0,0 +1,194 @@
...
+      case DEPENDENCIES_REQUIRE_SONAR_UPGRADE:
+        return "DEPS_SONAR_UPGRADE";
```
simonbrandhof: and here "DEPS_REQUIRE_UPGRADE"

simonbrandhof:
LGTM after my last feedbacks

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/AvailablePluginsWsAction.java
sns-seb: done

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/AvailablePluginsWsAction.java
sns-seb: done

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/AvailablePluginsWsAction.java
sns-seb: agreed, done

PR #229, opened 2015-04-17
SONAR-6383 add java WS listing pending plugin installs and removal

sns-seb:
once backing components have been updated to read plugin's jarfile manifests, more information will be added to the JSON response

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PendingPluginsWsAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PendingPluginsWsAction.java
@@ -1,0 +1,100 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * SonarQube is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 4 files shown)


server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PendingPluginsWsAction.java
```diff
@@ -0,0 +1,100 @@
...
+import static java.lang.String.CASE_INSENSITIVE_ORDER;
+
+/**
+ * Implementation of the {@code pending} action for the Plugins WebService.
+ */
+public class PendingPluginsWsAction implements PluginsWsAction {
+
+  private static final String ARRAY_INSTALLING = "installing";
+  private static final String ARRAY_REMOVING = "removing";
+  private static final String OBJECT_ARCHIVE = "archive";
```
simonbrandhof: to be consistent, please rename to "artifact"

server/sonar-server/src/test/java/org/sonar/server/plugins/ws/PendingPluginsWsActionTest.java
```diff
@@ -0,0 +1,190 @@
...
+  @InjectMocks
+  private PendingPluginsWsAction underTest;
+
+  @Mock
+  private Request request;
+  private WsTester.TestResponse response = new WsTester.TestResponse();
+
+  @Before
+  public void createAndWireMocks() throws Exception {
+    initMocks(this);
```
simonbrandhof: Just FYI (again), I personally much prefer the pattern `PluginDownloader pluginDownloader = mock(PluginDownloader.class);`
More concise and no need for `@Before`method

simonbrandhof:
LGTM if "archive" is renamed.

server/sonar-server/src/test/java/org/sonar/server/plugins/ws/PendingPluginsWsActionTest.java
sns-seb: I agree it is more concise.

I wrote it that way initially but then I changed my mind.

When I have a bunch of properties at the beginning of the test class, I find annotations much more readable that pure Java code, and yes, at the cost of the a little reflection.

I couldn't read my own code, so, out of pure kindness to my fellow developers, I went that in that direction instead of adding useless empty lines of moving constants to the end of the file :)

PR #230, opened 2015-04-17
separate test and coverage details in batch protocol - SONAR-6255

teryk:


sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/FileStructure.java
```diff
@@ -36,7 +36,8 @@
     SCM("scm-", Domain.PB),
     SYMBOLS("symbol-", Domain.PB),
     COVERAGE("coverage-", Domain.PB),
-    TEST_RESULT("test-", Domain.PB),
+    TESTS("tests-", Domain.PB),
```
henryju: Other are singular

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/FileStructure.java
```diff
@@ -36,7 +36,8 @@
...
+    COVERAGE_DETAILS("coverage-details-", Domain.PB),
```
henryju: Idem

henryju:
Except two minor previous comments, LGTM

PR #231, opened 2015-04-17
SONAR-6298 Add project count in qprofile search WS response

jblievremont:


simonbrandhof:
The example of WS response is not up-to-date with the new field "projectCount".

sonar-core/src/main/resources/org/sonar/core/qualityprofile/db/QualityProfileMapper.xml
```diff
@@ -151,6 +151,14 @@
     </where>
   </select>
 
+  <select id="countProjectsByProfile" resultType="org.sonar.core.qualityprofile.db.QualityProfileProjectCount">
```
simonbrandhof: @jblievremont @julienlancelot do you think that this SQL request should filter on PROJECTS.ENABLED=true ?

julienlancelot: Yes ! And maybe it should also use INNER JOIN instead of only JOIN (I do not know what is the default behaviour of JOIN)

jblievremont: Default behavior is INNER, but explicit > implicit. I'll also add the where, thanks!

simonbrandhof: Indeed explicit INNER is preferred as I don't believe the default behaviors of all our supported databases.

julienlancelot: :+1:

simonbrandhof:
@jblievremont example response is still not up-to-date

jblievremont:
Good catch. I also made sure that the count is only displayed for non-default profiles.

simonbrandhof:
LGTM

PR #202, opened 2015-04-07
SONAR-6366 new ws for db migration

sns-seb:
this PR adds the new WS `server/migrate` 

IT tests coming up as soon as I know how they are supposed to be written

this PR is based on branch `feature/add_db_migration_from_java` reviewed with PR #197

PR #233, opened 2015-04-20
SONAR-6366 new ws for db migration

sns-seb:
this PR adds the new WS server/migrate

this PR is the same as #202, but recreated based on master

IT tests are currently failing because restarting the Java WS after db upgrade does not work anymore

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigration.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigration.java
@@ -19,15 +19,18 @@
  */
 package org.sonar.server.db.migrations;
 
-import org.sonar.api.utils.log.Logger;
-import org.sonar.api.utils.log.Loggers;
-import org.sonar.server.ruby.RubyBridge;
-
-import javax.annotation.CheckForNull;
-import javax.annotation.Nullable;
 import java.util.Date;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantLock;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
+
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
+import org.sonar.api.utils.log.Profiler;
+import org.sonar.server.platform.Platform;
+import org.sonar.server.ruby.RubyBridge;
 
 /**
  * Handles concurrency to make sure only one DB migration can run at a time.
@@ -41,6 +44,7 @@ public class PlatformDatabaseMigration implements DatabaseMigration {
    * ExecutorService implements threads management.
    */
   private final PlatformDatabaseMigrationExecutorService executorService;
+  private final Platform platform;
   /**
    * This lock implements thread safety from concurrent calls of method {@link #startIt()}
    */
@@ -60,9 +64,10 @@ public class PlatformDatabaseMigration implements DatabaseMigration {
   @Nullable
   private Throwable failureError;
 
-  public PlatformDatabaseMigration(RubyBridge rubyBridge, PlatformDatabaseMigrationExecutorService executorService) {
+  public PlatformDatabaseMigration(RubyBridge rubyBridge, PlatformDatabaseMigrationExecutorService executorService, Platform platform) {
     this.rubyBridge = rubyBridge;
     this.executorService = executorService;
+    this.platform = platform;
   }
 
   @Override
@@ -96,18 +101,45 @@ public class PlatformDatabaseMigration implements DatabaseMigration {
         status = Status.RUNNING;
         startDate = new Date();
         failureError = null;
+        Profiler profiler = Profiler.create(LOGGER);
         try {
-          LOGGER.info("Starting DB Migration at {}", startDate);
-          rubyBridge.databaseMigration().trigger();
-          LOGGER.info("DB Migration ended successfully at {}", new Date());
+          profiler.startInfo("Starting DB Migration");
+          upgradeDb();
+          restartContainer();
+          recreateWebRoutes();
           status = Status.SUCCEEDED;
+          profiler.stopInfo("DB Migration ended successfully");
         } catch (Throwable t) {
-          LOGGER.error("DB Migration failed and ended at " + startDate + " with an exception", t);
+          profiler.stopInfo("DB migration failed");
+          LOGGER.error(
+            "DB Migration or container restart failed. Process ended with an exception", t
+            );
           status = Status.FAILED;
           failureError = t;
         } finally {
           running.getAndSet(false);
         }
+      }
+
+      private void upgradeDb() {
+        Profiler profiler = Profiler.createIfTrace(LOGGER);
+        profiler.startTrace("Starting DB Migration");
+        rubyBridge.databaseMigration().trigger();
+        profiler.stopTrace("DB Migration ended");
+      }
+
+      private void restartContainer() {
+        Profiler profiler = Profiler.createIfTrace(LOGGER);
+        profiler.startTrace("Restarting container");
+        platform.doStart();
+        profiler.stopTrace("Container restarted successfully");
+      }
+
+      private void recreateWebRoutes() {
+        Profiler profiler = Profiler.createIfTrace(LOGGER);
+        profiler.startTrace("Recreating web routes");
+        rubyBridge.railsRoutes().recreate();
+        profiler.startTrace("Routes recreated successfully");
       }
     });
   }
--- a/server/sonar-server/src/main/java/org/sonar/server/platform/Platform.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/platform/Platform.java
@@ -19,25 +19,27 @@
  */
 package org.sonar.server.platform;
 
-import org.sonar.api.platform.ComponentContainer;
-import org.sonar.api.platform.Server;
-import org.sonar.api.utils.log.Loggers;
-import org.sonar.core.persistence.DatabaseVersion;
-
-import javax.annotation.CheckForNull;
-import javax.servlet.ServletContext;
 import java.util.Collection;
 import java.util.Properties;
+import javax.annotation.CheckForNull;
+import javax.servlet.ServletContext;
+import org.sonar.api.platform.ComponentContainer;
+import org.sonar.api.platform.Server;
+import org.sonar.api.utils.log.Logger;
+import org.sonar.api.utils.log.Loggers;
+import org.sonar.core.persistence.DatabaseVersion;
 
 /**
  * @since 2.2
  */
 public class Platform {
 
+  private static final Logger LOGGER = Loggers.get(Platform.class);
+
   private static final Platform INSTANCE = new Platform();
 
   private ServerComponents serverComponents;
-  private ComponentContainer level1Container, level2Container, level3Container, level4Container;
+  private ComponentContainer level1Container, level2Container, safeModeContainer, level3Container, level4Container;
   private ComponentContainer currentContainer;
   private boolean dbConnected = false;
   private boolean started = false;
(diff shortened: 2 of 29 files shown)


sns-seb:
PR updated : 
* rebased on master
* fixed upgrade IT test broken by introduction of safemode in `Platform` class
* split upgrade_and_start single ruby call into 2 ruby calls (upgrade and recreate web route) and a pure java call in between to start the container in normal mode
* fix unit test `PlatformDatabaseMigrationAsynchronousTest` which class had been created but contained no actual test
* RoR code checking if SQ instance is accessible is now based on the Java DB migration code
* updated upgrade IT test using the new WS is visible on branch https://github.com/SonarSource/sonar-tests-core/tree/feature/add_java_ws_db_upgrade and will be merged into master once this PR has made it to master and gone gold

sns-seb:
PR updated : 
* rebased on master
* WS is now `/api/system/migrate_db` instead of `/api/server/migrate`
* had to remove authentication code in RoR on `/api/system` to code it in Java so I took the opportunity to remove authentication from `/api/system/upgrades` and `/api/system/restart` which wasn't actually required

server/sonar-server/src/main/resources/org/sonar/server/platform/ws/example-migrate.json
```diff
@@ -0,0 +1,6 @@
+{
+  "operational": "false",
```
simonbrandhof: Value is a boolean, not a string.

server/sonar-server/src/main/resources/org/sonar/server/platform/ws/example-migrate.json
```diff
@@ -0,0 +1,6 @@
...
+  "state": "MIGRATION_RUNNING",
+  "message": "Database migration is running",
+  "createdAt": "2015-02-23T18:54:23+0100"
```
simonbrandhof: Example is not up-to-date. Field is named "startedAt" but not "createdAt".

simonbrandhof:
I don't know if it's a regression or an existing bug, but POST api/system/restart does not return a valid error message when server is not in the required DEV mode:

```
{"errors":[{"msg":""}]}
```

server/sonar-server/src/main/java/org/sonar/server/platform/Platform.java
```diff
@@ -78,14 +82,29 @@ public void init(Properties properties, ServletContext servletContext) {
...
   public void doStart() {
-    if (!started && getDatabaseStatus() == DatabaseVersion.Status.UP_TO_DATE) {
+    if (started && !isInSafeMode()) {
+      return;
+    }
+
+    if (getDatabaseStatus() == DatabaseVersion.Status.UP_TO_DATE) {
       startLevel34Containers();
-      started = true;
+      if (isInSafeMode()) {
```
simonbrandhof: Not sure that "safe mode" is clear enough. Don't you think that name should mention the fact that db is connected but schema is not up-to-date ?

server/sonar-server/src/main/java/org/sonar/server/platform/ws/MigrateDbSystemWsAction.java
```diff
@@ -0,0 +1,168 @@
...
+    this.database = database;
+    this.databaseMigration = databaseMigration;
+  }
+
+  @Override
+  public void define(WebService.NewController controller) {
+    controller.createAction("migrate_db")
+      .setDescription("Migrate the database to match the current version of SonarQube." +
+        "<br/>" +
+        "Sending a POST request to this URL will start the DB migration. " +
```
simonbrandhof: Minor feedback about writing style: no need for future tense here (see section "Writing Style" of http://confluence.internal.sonarsource.com/display/PLAT/Design+Guidelines)

server/sonar-server/src/main/java/org/sonar/server/platform/ws/MigrateDbSystemWsAction.java
```diff
@@ -0,0 +1,168 @@
...
+        "It is strongly advised to <strong>make a database backup</strong> before invoking this WS." +
+        "<br/>" +
+        "State values are:" +
+        "<ul>" +
+        "<li>NO_MIGRATION: DB is up to date with current version of SonarQube</li>" +
+        "<li>MIGRATION_NEEDED: DB needs to be upgraded in order for the current version of SonarQube to run</li>" +
```
simonbrandhof: This state is misleading from a functional PoV. How is it possible as a migration was requested ? (I know the technical reason!)

server/sonar-server/src/main/java/org/sonar/server/platform/ws/MigrateDbSystemWsAction.java
```diff
@@ -0,0 +1,168 @@
...
+        case FAILED:
+        case SUCCEEDED:
+          writeResponse(response, databaseMigration);
+          break;
+        case NONE:
+          databaseMigration.startIt();
+          writeResponse(response, databaseMigration);
+          break;
+        default:
+          throw new RuntimeException(UNSUPPORTED_DATABASE_MIGRATION_STATUS);
```
simonbrandhof: We usually use IllegalStateException instead of raw RuntimeException. In any case it will raise an issue on Dory because of https://dory.internal.sonarsource.com/coding_rules#rule_key=squid%3AS00112

server/sonar-server/src/main/java/org/sonar/server/platform/ws/MigrateDbSystemWsAction.java
```diff
@@ -0,0 +1,168 @@
...
+      }
+    }
+  }
+
+  private void writeNotSupportedResponse(Response response) {
+    JsonWriter jsonWriter = response.newJsonWriter();
+    jsonWriter.beginObject()
+      .prop("operational", false)
+      .prop("state", statusToJson(NONE))
+      .prop("message", "Upgrade is not supported. Please use a <a href=\"http://redirect.sonarsource.com/doc/requirements.html\">production-ready database</a>.")
```
simonbrandhof: If message is not raw text, then I would rename the field "htmlMessage" to prevent any bad usage.

server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigration.java
```diff
@@ -97,18 +99,38 @@ public void run() {
...
+      private void restartContainer() {
+        LOGGER.info("Restarting container at {}", new Date());
+        platform.doStart();
+        LOGGER.info("Container restarted successfully at {}", new Date());
+      }
+
+      private void recreateWebRoutes() {
+        LOGGER.info("Recreating web routes at {}", new Date());
+        rubyBridge.railsRoutes().recreate();
+        LOGGER.info("Routes recreated successfully at {}", new Date());
```
simonbrandhof: Not sure such logs are needed, particularly the explicit end date as it's the first column of any log.

server/sonar-server/src/main/java/org/sonar/server/ui/JRubyFacade.java
```diff
@@ -373,4 +377,34 @@ public String getPeriodLabel(String mode, String param, String date) {
...
+      return true;
+    }
+
+    Database database = container.getComponentByType(Database.class);
+    if (!database.getDialect().supportsMigration()) {
+      return false;
+    }
+
+    DatabaseMigration databaseMigration = container.getComponentByType(DatabaseMigration.class);
+    Loggers.get(JRubyFacade.class).info("databaseMigration.status()=" + databaseMigration.status());
```
simonbrandhof: Noisy log:
```
2015.04.30 10:59:47 INFO  app[o.s.p.m.Monitor] Process[web] is up
2015.04.30 10:59:47 INFO  web[o.s.s.ui.JRubyFacade] databaseMigration.status()=NONE
2015.04.30 10:59:52 INFO  web[o.s.s.ui.JRubyFacade] databaseMigration.status()=NONE
2015.04.30 10:59:58 INFO  web[o.s.s.ui.JRubyFacade] databaseMigration.status()=NONE
```

simonbrandhof:
When db is being migrated or when pico is being started (up to level "5"), then webapp should still display "maintenance" page. Currently webapp displays pages with probably errors.

server/sonar-server/src/main/resources/org/sonar/server/platform/ws/example-migrate.json
sns-seb: well spotted

this is an example of a practice that I think should be enforced when writing WS: a unit test should use the example as expected JSON.

this way, examples will always be consistent and accurate

sns-seb:
> I don't know if it's a regression or an existing bug, but POST api/system/restart does not return a valid error message when server is not in the required DEV mode:

reproduced with an instance 5.1

server/sonar-server/src/main/resources/org/sonar/server/platform/ws/example-migrate.json
simonbrandhof: good proposal indeed

server/sonar-server/src/main/java/org/sonar/server/platform/ws/MigrateDbSystemWsAction.java
sns-seb: ok. done

server/sonar-server/src/main/java/org/sonar/server/platform/ws/MigrateDbSystemWsAction.java
sns-seb: there is no "state" concerned here, feels weird to use `IllegalStateException`.

simonbrandhof: IllegalArgumentException ?

server/sonar-server/src/main/java/org/sonar/server/platform/ws/MigrateDbSystemWsAction.java
sns-seb: description was incorrect. This state does not exist! It comes from the RoR based WS. So, we actually improved that :)

but the technical case I raised can still happen. We would be get a "NO_MIGRATION" status associated to a "Database is up-to-date, no migration needed." message, which also be wrong. 

So I fixed it.

server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigration.java
sns-seb: I reduced it to two INFO logs displayed by the `Profiler` class (so include timing in ms) and Profiler logs at TRACE level for each step

server/sonar-server/src/main/java/org/sonar/server/ui/JRubyFacade.java
sns-seb: well spoted

server/sonar-server/src/main/java/org/sonar/server/platform/ws/MigrateDbSystemWsAction.java
sns-seb: why not. is it common practice?

server/sonar-server/src/main/java/org/sonar/server/platform/ws/MigrateDbSystemWsAction.java
sns-seb: for a lack of better alternative, I'll go with `IllegalArgumentException`

server/sonar-server/src/main/java/org/sonar/server/platform/Platform.java
sns-seb: I find safemode appropriate because here, it's more a matter of describing the state in which the SQ instance is than describing the reason it is in this state.

The reason it is in this state might evolve over time.
eg. SQ instance could start even if when a plugin problem occurs at startup, this way avoiding having an instance down. This is especially something we may want if we implement the plugin update evolution we talked about.

sns-seb:
> When db is being migrated or when pico is being started (up to level "5"), then webapp should still display "maintenance" page. Currently webapp displays pages with probably errors.

fixed

server/sonar-server/src/main/java/org/sonar/server/platform/ws/MigrateDbSystemWsAction.java
sns-seb: I worked around that by removing this message with HTML completly. I think providing a link should be the responsability of the UI not the WS.

In addition, I introduced a `NOT_SUPPORTED` status to correctly represent that case instead of using status `NONE` with a specific message.

sns-seb:
PR updated:
* rebased on master
* feeback taken into account

simonbrandhof:
LGTM, I let PO do final validation.

PR #234, opened 2015-04-20
better java ws listing installed and pending plugins

sns-seb:
this PR:

* adds more info in the response of the pending WS (in fact, the same info as the WS for installed plugins)
* refactors code in plugins WS actions to share duplicated code
* removes the "url" object from the installed plugins WS


simonbrandhof:
LGTM

PR #235, opened 2015-04-20
SONAR-6256 Replace component id columns by component uuid in dependencies table

julienlancelot:


server/sonar-server/src/main/java/org/sonar/server/db/migrations/v52/AddDependenciesComponentUuidColumns.java
```diff
@@ -0,0 +1,67 @@
...
+  public void execute(Context context) throws SQLException {
+    context.execute(generateSql());
+  }
+
+  private String generateSql() {
+    return new AddColumnsBuilder(db.getDialect(), "dependencies")
+      .addColumn(
+        new AddColumnsBuilder.ColumnDef()
+          .setName("from_component_uuid")
+          .setType(AddColumnsBuilder.ColumnDef.Type.STRING)
```
teryk: use import to have a shorter reference

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v52/AddDependenciesComponentUuidColumns.java
```diff
@@ -0,0 +1,67 @@
...
+          .setLimit(50)
+          .setNullable(true)
+      )
+      .addColumn(
+        new AddColumnsBuilder.ColumnDef()
+          .setName("to_component_uuid")
+          .setType(AddColumnsBuilder.ColumnDef.Type.STRING)
```
teryk: use import to have a shorter reference

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v52/FeedDependenciesComponentUuids.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public FeedDependenciesComponentUuids(Database db) {
+    super(db);
+  }
+
+  @Override
+  public void execute(Context context) throws SQLException {
+    MassUpdate update = context.prepareMassUpdate().rowPluralName("dependencies");
+    update.select(
+      "SELECT from_component.uuid, to_component.uuid, dependency.id " +
```
teryk: how will you handle case where component is not found in projects ? (partiel db cleaner)

server/sonar-server/src/test/java/org/sonar/server/db/migrations/v52/DropDependenciesComponentColumnsTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class DropDependenciesComponentColumnsTest {
+
+  DropDependenciesComponentColumns migration;
+
+  Database database;
+
+  @Before
+  public void setUp() throws Exception {
+    database = mock(Database.class);
```
teryk: what is missing in DbTester to use it to test column dropping ?

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v52/AddDependenciesComponentUuidColumns.java
julienlancelot: Why ?

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v52/FeedDependenciesComponentUuids.java
julienlancelot: Indeed, but it should be taking into account by the task that will clean the db (if db could be "corrupted", it should be regularly clean up) -> I've added a comment in https://jira.codehaus.org/browse/SONAR-5503

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v52/AddDependenciesComponentUuidColumns.java
teryk: better readability ?

server/sonar-server/src/test/java/org/sonar/server/db/migrations/v52/DropDependenciesComponentColumnsTest.java
julienlancelot: Nothing, the issue is not on DbTester side but on H2 -> It does not support drop of multiple columns

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v52/AddDependenciesComponentUuidColumns.java
julienlancelot: That can be discussed, but why not

teryk:
LGTM

PR #236, opened 2015-04-20
SONAR-6427 Internal WS to get global navigation configuration

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/ui/ws/NavigationWs.java
```diff
@@ -0,0 +1,45 @@
...
+  private final NavigationAction[] actions;
+
+  public NavigationWs(NavigationAction... actions) {
+    this.actions = actions;
+  }
+
+  @Override
+  public void define(Context context) {
+    NewController navigation = context.createController("api/navigation")
+      .setDescription("Get information about navigation options")
```
julienlancelot: I'm not sure "options" is the right word...

server/sonar-server/src/main/java/org/sonar/server/ui/ws/GlobalNavigationAction.java
```diff
@@ -0,0 +1,99 @@
...
+    writePages(json);
+    writeLogoProperties(json);
+    json.endObject().close();
+  }
+
+  private void writeDashboards(JsonWriter json, List<DashboardDto> dashboards) {
+    json.name("globalDashboards").beginArray();
+    for (DashboardDto dashboard : dashboards) {
+      json.beginObject()
+        .prop("id", dashboard.getId())
```
julienlancelot: "id" should not be returned in WS. Even if we still return the db id, we should rename it "key", and when tomorrow we'll return an functional key or a uuid, there will be no need to change the WS.

sns-seb: Is this WS internal ? Otherwise I find it disturbing to have the content of a field modified, at least in principle.

Lets say a consumer uses this information to math with a field from another WS, if the content of the field changes, the consumer's implementation will break.

disclaimer: I'm a rookie, my comment might not apply here :)

server/sonar-server/src/main/java/org/sonar/server/ui/ws/GlobalNavigationAction.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+    json.endArray();
+  }
+
+  private void writePages(JsonWriter json) {
+    json.name("globalPages").beginArray();
+    for (ViewProxy<Page> page : views.getPages(NavigationSection.HOME)) {
+      json.beginObject()
+        .prop("name", page.getTitle())
+        .prop("url", page.isController() ? page.getId() : String.format("/plugins/home/%s", page.getId()))
```
julienlancelot: When the page is not a controller, should we prefix it by the web context ?

server/sonar-server/src/test/java/org/sonar/server/ui/ws/GlobalNavigationActionTest.java
```diff
@@ -0,0 +1,169 @@
...
+    wsTester = new WsTester(new NavigationWs(new GlobalNavigationAction(activeDashboardDao, new Views(), new Settings())));
+
+    MockUserSession.set();
+
+    wsTester.newGetRequest("api/navigation", "global").execute().assertJson(getClass(), "empty.json");
+  }
+
+  @Test
+  public void nominal_call_for_anonymous() throws Exception {
+    nominalSetup();
```
julienlancelot: It's a personal choice, but I would prefer to not call here a common class, in order to know exactly what this test is testing -> so I would do here : 

DashboardDto defaultDashboardForAnonymous = createDashboard(2L, "Default Dashboard for Anonymous", true);
dashboardDao.insert(defaultDashboardForAnonymous);
activeDashboardDao.insert(createActiveDashboard(user.getId(), defaultDashboardForUser.getId(), 1));
session.commit();

server/sonar-server/src/test/java/org/sonar/server/ui/ws/GlobalNavigationActionTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+    MockUserSession.set();
+
+    wsTester.newGetRequest("api/navigation", "global").execute().assertJson(getClass(), "anonymous.json");
+  }
+
+  @Test
+  public void nominal_call_for_user() throws Exception {
+    nominalSetup();
```
julienlancelot: Same as before

server/sonar-server/src/test/java/org/sonar/server/ui/ws/GlobalNavigationActionTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+    MockUserSession.set().setLogin("obiwan");
+
+    wsTester.newGetRequest("api/navigation", "global").execute().assertJson(getClass(), "user.json");
+  }
+
```
julienlancelot: I would also create a specific test for the display of the logo setting

sonar-core/src/test/java/org/sonar/core/dashboard/ActiveDashboardDaoTest.java
```diff
@@ -23,8 +23,7 @@
 import org.junit.Test;
 import org.sonar.core.persistence.AbstractDaoTestCase;
 
-import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.assertThat;
+import static org.assertj.core.api.Assertions.assertThat;
 
 public class ActiveDashboardDaoTest extends AbstractDaoTestCase {
```
julienlancelot: AbstractDaoTestCase should not be used anymore (I will deprecate it in order to not forget to remove it each time we see it!), it should be replaced by a call to DbTester.

julienlancelot:
Please apply my feedback

server/sonar-server/src/main/java/org/sonar/server/ui/ws/GlobalNavigationAction.java
julienlancelot: Yes it's internal -> see https://github.com/SonarSource/sonarqube/pull/236/files#diff-c8b18efcb7cf25f3080b2c4001c8d09fR54, so it will only be used by us !

server/sonar-server/src/main/java/org/sonar/server/ui/ws/GlobalNavigationAction.java
jblievremont: It is not prefixed in the Rails version, the UI already prepends the baseUrl.

server/sonar-server/src/test/java/org/sonar/server/ui/ws/GlobalNavigationActionTest.java
jblievremont: This is on purpose: the goal is to test that the anonymous user only sees what is allowed/configured for them, whereas a connected user gets their own configuration.

server/sonar-server/src/main/java/org/sonar/server/ui/ws/GlobalNavigationAction.java
julienlancelot: ok

server/sonar-server/src/test/java/org/sonar/server/ui/ws/GlobalNavigationActionTest.java
julienlancelot: Good point ! It this case, to make it clear to the reader, I would use 2 methods : createDashboardForAnonymous() and createDashboardForUser() (Or something like that)

julienlancelot:
Last minor feedback to apply then LGTM !

julienlancelot:
Last info : the package-info.java file is missing in 'src/main/java/org/sonar/server/ui/ws' !

PR #237, opened 2015-04-21
SONAR-6379 add Java WS to list plugin updates

sns-seb:
This PR:

* adds a new WS to list installed plugins for which a newer version is available
* adds explicit sorting of plugins in response of available WS (done in this PR because code is common with updatable WS)

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/platform/ServerComponents.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/platform/ServerComponents.java
@@ -222,6 +222,7 @@ import org.sonar.server.plugins.ws.PendingPluginsWsAction;
 import org.sonar.server.plugins.ws.PluginWSCommons;
 import org.sonar.server.plugins.ws.PluginsWs;
 import org.sonar.server.plugins.ws.UpdatePluginsWsAction;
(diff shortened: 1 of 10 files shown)


simonbrandhof:
@sns-seb as written in the description, multiple versions of the same plugin can be returned. How this case is handled by the JSON response ? Multiple "plugins" with same key ?

sns-seb:
I double-checked the `UpdateCenter` code, only one version is returned per plugin (the latest compatible with the current version of SQ). The description is inaccurate and I shall try and find a new and non ambiguous way to write it.

simonbrandhof:
That's strange. That means that it's not possible to see the changelog of all the versions greater than the installed one.

sns-seb:
Well, triple-checking wasn't a waste. I read the install part of the update center code.

currently, each plugin version will be displayed as a separate plugin object in the list. I don't think this appropriate but that's how UpdateCenter provides the information.

simonbrandhof:
I know we should think about future changes, but in this specific case shouldn't we prepare the JSON response with an appropriate format ?

sns-seb:
PR updated : 

* rebased on master
* URL renamed to /api/plugins/updates

simonbrandhof:
LGTM

PR #239, opened 2015-04-21
SONAR-6381 add java WS to install a plugin

sns-seb:
just the WS glue to the underlying components.
note : only the latest version of the plugin can be installed

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstallPluginsWsAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstallPluginsWsAction.java
@@ -1,0 +1,105 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * SonarQube is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.plugins.ws;
+
+import com.google.common.base.Predicate;
+import com.google.common.collect.Iterables;
+import org.sonar.api.server.ws.Request;
+import org.sonar.api.server.ws.Response;
+import org.sonar.api.server.ws.WebService;
+import org.sonar.core.permission.GlobalPermissions;
+import org.sonar.server.plugins.PluginDownloader;
+import org.sonar.server.plugins.UpdateCenterMatrixFactory;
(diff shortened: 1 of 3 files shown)


server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstallPluginsWsAction.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Override
+  public void define(WebService.NewController controller) {
+    WebService.NewAction action = controller.createAction("install")
+      .setPost(true)
+      .setDescription("Installs the latest version of a plugin specified by its key.")
+      .setHandler(this);
+
+    action.createParam(PARAM_KEY).setRequired(true)
+      .setDescription("The key identifying the plugin to install");
```
jblievremont: Minor feedback: can you provide an example value?

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstallPluginsWsAction.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Override
+  public void handle(Request request, Response response) throws Exception {
+    String key = request.mandatoryParam(PARAM_KEY);
+    PluginUpdate pluginUpdate = findAvailablePluginByKey(key);
+    pluginDownloader.download(key, pluginUpdate.getRelease().getVersion());
+    response.noContent();
+  }
+
+  @Nonnull
```
simonbrandhof: No need for @Nonnull annotation on method returns as this is the default behavior. @CheckForNull should be used when result can be null.

simonbrandhof:
LGTM

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstallPluginsWsAction.java
simonbrandhof: This WS does not return any JSON response. We could think about documenting such behavior, but how ?

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstallPluginsWsAction.java
```diff
@@ -0,0 +1,102 @@
...
+    PluginDownloader pluginDownloader) {
+    this.updateCenterFactory = updateCenterFactory;
+    this.pluginDownloader = pluginDownloader;
+  }
+
+  @Override
+  public void define(WebService.NewController controller) {
+    WebService.NewAction action = controller.createAction("install")
+      .setPost(true)
+      .setDescription("Installs the latest version of a plugin specified by its key.")
```
simonbrandhof: You should maybe mention that the plugin is downloaded from update center.

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstallPluginsWsAction.java
```diff
@@ -0,0 +1,102 @@
...
+      .setHandler(this);
+
+    action.createParam(PARAM_KEY).setRequired(true)
+      .setDescription("The key identifying the plugin to install");
+  }
+
+  @Override
+  public void handle(Request request, Response response) throws Exception {
```
simonbrandhof: Security - check that user is a system administrator.

simonbrandhof:
Removed my previous LGTM comment - see blocker comment about permission

sns-seb:
PR update with blocking comment taken into account and rebase on master

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/InstallPluginsWsAction.java
sns-seb: +1

simonbrandhof:
Added comment on description + LGTM

PR #240, opened 2015-04-21
SONAR-6445 Decrease the max memory used by server bootstrapper process

simonbrandhof:
See https://jira.codehaus.org/browse/SONAR-6445. Decreasing to 8Mb is now possible with Java 7.


julienlancelot:
I can make the server start with 8m of Xmx...

julienlancelot:
LGTM

PR #243, opened 2015-04-22
SONAR-6380 add Java WS to uninstall a plugin

sns-seb:
add a Java WS to uninstall an installed plugin

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/UninstallPluginsWsAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/UninstallPluginsWsAction.java
@@ -1,0 +1,93 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * SonarQube is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.plugins.ws;
+
+import com.google.common.base.Predicate;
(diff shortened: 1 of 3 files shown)


server/sonar-server/src/main/java/org/sonar/server/plugins/ws/UninstallPluginsWsAction.java
```diff
@@ -0,0 +1,93 @@
...
+      .setDescription("The key identifying the plugin to uninstall")
+      .setRequired(true);
+  }
+
+  @Override
+  public void handle(Request request, Response response) throws Exception {
+    String key = request.mandatoryParam(PARAM_KEY);
+    ensurePluginIsInstalled(key);
+    pluginJarsInstaller.uninstall(key);
+  }
```
jblievremont: There is no response to this WS. In this case, you should call response.noContent() to have a correct 204 status code.

jblievremont:
Just some tiny feedback, then LGTM!

sns-seb:
applied feedback and rebased on master

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/UninstallPluginsWsAction.java
sns-seb: you're right, I'll also add an assertion in the unit test making sure the response is empty (unfortunately, it does not seem possible to verify the HTTP status with `TestResponse`)

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/UninstallPluginsWsAction.java
```diff
@@ -0,0 +1,88 @@
...
+  @Override
+  public void define(WebService.NewController controller) {
+    WebService.NewAction action = controller.createAction("uninstall")
+      .setPost(true)
+      .setDescription("Uninstalls the plugin specified by its key.")
+      .setHandler(this);
+
+    action.createParam(PARAM_KEY)
+      .setDescription("The key identifying the plugin to uninstall")
+      .setRequired(true);
```
jblievremont: Last minor feedback: can you provide an example value?

sns-seb: initially thought about providing an example value, but for a plugin key, I didn't see the value

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/UninstallPluginsWsAction.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  @Override
+  public void handle(Request request, Response response) throws Exception {
```
simonbrandhof: Security check - user must be a system administrator.

simonbrandhof:
Added a blocker comment.

sns-seb:
PR updated to take blocker comment into account and rebase on master

simonbrandhof:
Added comment about description, then LGTM

PR #245, opened 2015-04-22
SONAR-6256 Add dependencies in batch report

julienlancelot:


sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/FileStructure.java
```diff
@@ -38,6 +38,8 @@
     COVERAGES("coverages-", Domain.PB),
     TESTS("tests-", Domain.PB),
     COVERAGE_DETAILS("coverage-details-", Domain.PB),
+    FILE_DEPENDENCIES("file-dependencies-", Domain.PB),
+    MODULE_DEPENDENCIES("module-dependencies-", Domain.PB),
```
teryk: where is the writer/reader for module dependencies ?

sonar-batch-protocol/src/main/protobuf/batch_report.proto
```diff
@@ -251,3 +251,19 @@ message CoverageDetail {
     repeated int32 covered_line = 2 [packed = true];
   }
 }
+
+message FileDependency {
+  optional int32 to_file_ref = 1;
+  optional int32 weight = 2;
+}
+
+message ModuleDependencies {
```
teryk: no need for streaming ?

sonar-batch-protocol/src/main/java/org/sonar/batch/protocol/output/FileStructure.java
julienlancelot: Oups, I've forgot it...

sonar-batch-protocol/src/main/protobuf/batch_report.proto
julienlancelot: It's not possible to do some streaming as we need to read all data in memory to be able to compute the module dependencies.

teryk:
apply my feedback and ...
lgtm-pulp-fiction

PR #246, opened 2015-04-22
SONAR-6366 improve handling of really simultaneous db migration calls

sns-seb:
this will also make unit test stable

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigration.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigration.java
@@ -19,6 +19,7 @@
  */
 package org.sonar.server.db.migrations;
 
+import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 import org.sonar.server.ruby.RubyBridge;
 
@@ -32,6 +33,8 @@ import java.util.concurrent.locks.ReentrantLock;
  * Handles concurrency to make sure only one DB migration can run at a time.
  */
 public class PlatformDatabaseMigration implements DatabaseMigration {
+
+  private static final Logger LOGGER = Loggers.get(PlatformDatabaseMigration.class);
 
   private final RubyBridge rubyBridge;
   /**
@@ -50,7 +53,7 @@ public class PlatformDatabaseMigration implements DatabaseMigration {
    * by the thread executing the db migration.
    * </p>
    */
-  private AtomicBoolean running = new AtomicBoolean(false);
+  private final AtomicBoolean running = new AtomicBoolean(false);
   private Status status = Status.NONE;
   @Nullable
   private Date startDate;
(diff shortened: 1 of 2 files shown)


simonbrandhof:
Out of scope of this commit, but field PlatformDatabaseMigration.running should be `final`.

server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigration.java
```diff
@@ -64,15 +68,34 @@ public PlatformDatabaseMigration(RubyBridge rubyBridge, PlatformDatabaseMigratio
 
   @Override
   public void startIt() {
-    if (lock.isLocked() || this.running.get() /* fail-fast if db migration is running */) {
+    if (lock.isLocked() || this.running.get()) {
+      LOGGER.debug("{}: lock is already taken or process is already running", Thread.currentThread().getName());
```
simonbrandhof: If this log is for developers, then level should be `trace`)

server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigration.java
```diff
@@ -64,15 +68,34 @@ public PlatformDatabaseMigration(RubyBridge rubyBridge, PlatformDatabaseMigratio
 
   @Override
   public void startIt() {
```
simonbrandhof: Not sure to understand the implementation of startIt(). Why don't you do something like the following ?

```
// no condition on lock.isLocked()
if (lock.tryLock()) {
  try {
    startAsynchronousDBMigration();
  } finally {
    lock.unlock();
  }
}
```

server/sonar-server/src/test/java/org/sonar/server/db/migrations/PlatformDatabaseMigrationConcurrentAccessTest.java
```diff
@@ -21,50 +21,58 @@
...
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
-import static org.mockito.internal.verification.VerificationModeFactory.times;
 
 public class PlatformDatabaseMigrationConcurrentAccessTest {
```
simonbrandhof: On tests involving threads, I recommend to define a safe-keeper:
```
@Rule
public TestRule timeout = new DisableOnDebug(Timeout.seconds(5));
```

sns-seb:
> PlatformDatabaseMigration.running should be `final`

indeed

server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigration.java
sns-seb: the condition at the beginning of `startIt` is a fail fast which does not imply any concurrency management.
method `PlatformDatabaseMigration#tryLock()` is actually useless since the only difference is the respect of the fairness policy which we do not use anyway in this class. I'll remove it.

server/sonar-server/src/test/java/org/sonar/server/db/migrations/PlatformDatabaseMigrationConcurrentAccessTest.java
sns-seb: this rule is useless and would be ineffective in debug because test already has a timeout
`pool.awaitTermination(3, TimeUnit.SECONDS);`

good to know this rule exists, though

how does it compare to @Test(timeout = 5)?

sns-seb:
PR updated : 

* feedback applied
* rebased on master

server/sonar-server/src/test/java/org/sonar/server/db/migrations/PlatformDatabaseMigrationConcurrentAccessTest.java
simonbrandhof: The (only ?) difference is the automatic disabling when debugging (thanks to DisableOnDebug)

simonbrandhof:
LGTM

PR #248, opened 2015-04-22
SONAR-6366 threads should be named

sns-seb:
small follow-up PR for SONAR-6366

threads should be named, especially long living one, thread name can be very helpful with reading a thread dump

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigrationExecutorServiceImpl.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigrationExecutorServiceImpl.java
@@ -19,6 +19,7 @@
  */
 package org.sonar.server.db.migrations;
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import org.sonar.server.util.AbstractStoppableExecutorService;
 
 import java.util.concurrent.Executors;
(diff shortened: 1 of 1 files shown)


server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigrationExecutorServiceImpl.java
```diff
@@ -32,6 +33,12 @@
   implements PlatformDatabaseMigrationExecutorService {
 
   public PlatformDatabaseMigrationExecutorServiceImpl() {
-    super(Executors.newSingleThreadExecutor());
+    super(
+      Executors.newSingleThreadExecutor(
+        new ThreadFactoryBuilder()
+          .setDaemon(false)
```
simonbrandhof: Daemon is out of scope of this change. Anyway setting to `false` is correct !

simonbrandhof:
LGTM

server/sonar-server/src/main/java/org/sonar/server/db/migrations/PlatformDatabaseMigrationExecutorServiceImpl.java
sns-seb: the `ThreadFactory` implicitly used when calling `Executors.newSingleThreadExecutor` without arguments sets the `daemon` flag on created threads explicitly. Settings it here enforces the behavior is the same, only the name is changed.

PR #249, opened 2015-04-22
SONAR-6392 Drop measure trends

simonbrandhof:
SONAR-6392 Drop measure trends

Database column PROJECT_MEASURES.TENDENCY is not deleted because such removal can be very long for some database vendors. Meanwhile it is ignored and always null for new rows.


julienlancelot:
LGTM !

PR #250, opened 2015-04-23
fix one line shift while auto-assigning issues in CE

teryk:


simonbrandhof:
does it bug exist in 5.1 ? If yes then a JIRA ticket must be created.

teryk:
it doesn't exist in 5.1. Code was misleading but working fine.

julienlancelot:
LGTM (but I haven't tested it)

PR #252, opened 2015-04-23
SONAR-6428 WS to get component navigation information

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/measure/persistence/MeasureDao.java
```diff
@@ -55,6 +56,10 @@ public void insert(DbSession session, MeasureDto measureDto) {
     mapper(session).insert(measureDto);
   }
 
+  public List<String> selectMetricKeysForSnapshot(DbSession session, Long snapshotId) {
```
simonbrandhof: snapshotId can't be null, so I would enforce that by using a primitive type "long"

server/sonar-server/src/main/java/org/sonar/server/ui/ws/ComponentConfigurationPages.java
```diff
@@ -0,0 +1,156 @@
...
+
+  boolean componentTypeHasProperty(ComponentDto component, String resourceTypeProperty) {
+    ResourceType resourceType = resourceTypes.get(component.qualifier());
+    if (resourceType != null) {
+      return resourceType.getBooleanProperty(resourceTypeProperty);
+    }
+    return false;
+  }
+
+  static class ConfigPage {
```
simonbrandhof: Note that his class and its methods can be private instead of package protected.

server/sonar-server/src/test/java/org/sonar/server/ui/ws/ComponentConfigurationPagesTest.java
```diff
@@ -0,0 +1,123 @@
...
+  public void pages_for_project_with_resource_type_property() throws Exception {
+    String uuid = "abcd";
+    ComponentDto component = ComponentTesting.newProjectDto(uuid);
+    UserSession userSession = MockUserSession.set().setLogin("obiwan").addProjectUuidPermissions(UserRole.ADMIN, uuid);
+    when(resourceTypes.get(component.qualifier())).thenReturn(
+      ResourceType.builder(component.qualifier()).setProperty("configurable", true).build());
+
+    List<ConfigPage> pages = new ComponentConfigurationPages(i18n, resourceTypes).getConfigPages(component, userSession);
+    assertThat(pages).extracting("visible").containsExactly(
+      true, true, true, true, true, true, false, false, false, false);
```
simonbrandhof: not sure to understand what this test does. The assertion is not clear enough.

sonar-core/src/main/resources/org/sonar/core/measure/db/MeasureMapper.xml
```diff
@@ -83,4 +83,11 @@
     )
   </insert>
 
+  <select id="selectMetricKeysForSnapshot" parameterType="long" resultType="string">
+  SELECT m.name
```
simonbrandhof: A snapshot can have multiple measures with the same metric id (key is [snapshot_id, metric_id, characteristic_id, rule_id, person_id]). For this reason SELECT DISTINCT is probably missing here.

simonbrandhof:
LGTM

PR #253, opened 2015-04-23
SONAR-6379 fix response to display many updates per plugin

sns-seb:
this PR modifies the WS code to aggregate PluginUpdate objects from UpdateCenter so that the WS sends out information in the target format (multiple release per plugin object instead of one plugin for each release)

also, it modifies the unit test to test the sample response instead of some other json response specific to the unit test

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PluginUpdateAggregator.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PluginUpdateAggregator.java
@@ -1,0 +1,119 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * SonarQube is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.plugins.ws;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Function;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import org.sonar.updatecenter.common.Plugin;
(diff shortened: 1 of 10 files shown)


server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PluginUpdateAggregator.java
```diff
@@ -0,0 +1,116 @@
...
+  private enum BuilderToPluginUpdateAggregate implements Function<PluginUpdateAggregateBuilder, PluginUpdateAggregate> {
+    INSTANCE;
+
+    @Override
+    public PluginUpdateAggregate apply(@Nonnull PluginUpdateAggregateBuilder input) {
+      return input.build();
+    }
+  }
+
+  public static class PluginUpdateAggregateBuilder {
```
simonbrandhof: this class and its methods do not need to be public.

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PluginUpdateAggregator.java
```diff
@@ -0,0 +1,116 @@
...
+
+    protected PluginUpdateAggregate(PluginUpdateAggregateBuilder builder) {
+      this.plugin = builder.plugin;
+      this.releases = ImmutableList.copyOf(builder.releases);
+    }
+
+    public Plugin getPlugin() {
+      return plugin;
+    }
+    public Collection<PluginUpdate> getReleases() {
```
simonbrandhof: strange to see getReleases() return `PluginUpdate` no ?

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PluginWSCommons.java
```diff
@@ -142,9 +143,44 @@ public void writeArtifact(JsonWriter jsonWriter, Release release) {
     jsonWriter.endObject();
   }
 
+  /**
+   * Write an "update" object to the specified jsonwriter.
+   * <pre>
+   * "update": {
```
simonbrandhof: this documentation does not seem to be up-to-date ("updates" has fields "releases" and "requires")

sns-seb: this method writes the `update` (no **s**) object and so documentation is in sync

did your comment actually apply to another method?

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PluginUpdateAggregator.java
sns-seb: indeed, package protected is enough

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PluginUpdateAggregator.java
sns-seb: indeed, package protected is enough

server/sonar-server/src/main/java/org/sonar/server/plugins/ws/PluginUpdateAggregator.java
sns-seb: renaming property to `updates`

sns-seb:
PE updated to take comments into account and rebase on master

simonbrandhof:
LGTM

PR #254, opened 2015-04-23
SONAR-6429 WS to support settings navigation

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/ui/ws/SettingsNavigationAction.java
```diff
@@ -0,0 +1,82 @@
...
+  public SettingsNavigationAction(Settings settings, Views views, I18n i18n) {
+    this.views = views;
+    this.settings = settings;
+    this.i18n = i18n;
+  }
+
+  @Override
+  public void define(NewController context) {
+    context.createAction("settings")
+      .setDescription("Get information concerning settings navigation for the current user.")
```
julienlancelot: The description is misleading : it makes me think that it's returning all settings pages (pages under Configuration, Security, System) but it's only returning settings pages provided by plugin and it's also saying if the update center page should be displayed or not.

I would have implemented this WS to return all settings pages, but if you keep current implementation please update the description.

julienlancelot:
I would have implement this WS a different way, by returning all settings page. 
But as it's an internal WS, it could be updated in the futur, so LGTM.

PR #256, opened 2015-04-24
SONAR-6385 add WS to list SonarQube upgrades

sns-seb:
I'll investigate why I have imports reodered on my laptop before merging to master

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/platform/ws/UpgradesSystemWsAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/platform/ws/UpgradesSystemWsAction.java
@@ -1,0 +1,156 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * SonarQube is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


simonbrandhof:
Terminology feedback - api/plugins/updates is the similar WS for plugins. For consistency don't you think that  the same action should be used: "updates" or "upgrades" ?

server/sonar-server/src/main/java/org/sonar/server/platform/ws/UpgradesSystemWsAction.java
```diff
@@ -0,0 +1,155 @@
...
+      .setResponseExample(Resources.getResource(this.getClass(), "example-updates_plugins.json"));
+  }
+
+  @Override
+  public void handle(Request request, Response response) throws Exception {
+    JsonWriter jsonWriter = response.newJsonWriter().setSerializeEmptys(false);
+
+    jsonWriter.beginObject();
+    UpdateCenter updateCenter = updateCenterFactory.getUpdateCenter(DO_NOT_FORCE_REFRESH);
+    writeSonarUpgrades(jsonWriter, updateCenter);
```
simonbrandhof: sonar is now sonarqube :-) Method could be renamed "writeSystem*"

simonbrandhof:
else LGTM

sns-seb:
> Terminology feedback - api/plugins/updates is the similar WS for plugins. For consistency don't you think that the same action should be used: "updates" or "upgrades" ?

In plugin WebService, when talking about the plugin compatibility/update status, we distinguish  "plugin updates" from "SQ updates" by calling the later "upgrades". 

In addition, I like the term "upgrade" for the system, since imo the noun implies a bigger modification than a "simple" update. Upgrading SQ may involve updating plugins, it makes sense. The Oxford definition of the nouns kind of goes in my direction (but is subject to interpretation) http://www.oxforddictionaries.com/definition/english/upgrade http://www.oxforddictionaries.com/definition/english/update.

simonbrandhof:
Good, you have a strong opinion. So ok !

PR #257, opened 2015-04-24
SONAR-6493 java.lang.NullPointerException in the PopulateProjectsUuidColumns migration

julienlancelot:
SONAR-6493 java.lang.NullPointerException in the PopulateProjectsUuidColumns migration


agigleux:
Don't you think it would be good to check also if currentComponent.getScope() is NULL ?

julienlancelot:
@agigleux It should never happen. currentComponent can be null in some known cases, but I do not know cases where the scope could be null. 
Do you have some use cases ?

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v50/PopulateProjectsUuidColumnsMigration.java
```diff
@@ -86,98 +91,107 @@ public void handleResult(ResultContext context) {
...
-    StringBuilder moduleUuidPath = new StringBuilder();
-    Component lastModule = null;
-    if (!Strings.isNullOrEmpty(snapshotPath)) {
-      for (String s : Splitter.on(".").omitEmptyStrings().split(snapshotPath)) {
-        Long snapshotId = Long.valueOf(s);
-        Component currentComponent = componentsBySnapshotId.get(snapshotId);
-        if (currentComponent.getScope().equals(Scopes.PROJECT)) {
-          lastModule = currentComponent;
-          moduleUuidPath.append(currentComponent.getUuid()).append(".");
+  private static class MigrationContext {
```
jblievremont: :+1:

server/sonar-server/src/test/resources/org/sonar/server/db/migrations/v50/PopulateProjectsUuidColumnsMigrationTest/not_fail_when_project_has_two_active_snapshots.xml
```diff
@@ -0,0 +1,43 @@
...
+             status="P" islast="[true]" purge_status="[null]"
+             period1_mode="[null]" period1_param="[null]" period1_date="[null]"
+             period2_mode="[null]" period2_param="[null]" period2_date="[null]"
+             period3_mode="[null]" period3_param="[null]" period3_date="[null]"
+             period4_mode="[null]" period4_param="[null]" period4_date="[null]"
+             period5_mode="[null]" period5_param="[null]" period5_date="[null]"
+             depth="[null]" scope="PRJ" qualifier="TRK" created_at="2008-12-02 13:58:00.00" build_date="2008-12-02 13:58:00.00"
+             version="[null]" path=""/>
+
+  <!-- module linked on second active snapshot of th project -->
```
jblievremont: Could it happen to have 2 different modules linked on each snapshot? What is the behavior in this case?

jblievremont:
Little question to answer, then LTGM

server/sonar-server/src/main/java/org/sonar/server/db/migrations/v50/PopulateProjectsUuidColumnsMigration.java
julienlancelot: Indeed, sorry for that, but it was too hard for me to read my own code without this Context class !

server/sonar-server/src/test/resources/org/sonar/server/db/migrations/v50/PopulateProjectsUuidColumnsMigrationTest/not_fail_when_project_has_two_active_snapshots.xml
julienlancelot: Yes it can. One module will be ignored and the other one will be taking into account. This use case is testing the case when the module will be ignored.

PR #258, opened 2015-04-24
SONAR-6430 add plugin source revision to JSON response

sns-seb:
information added to response of WS listing installed pluging and pending operations on plugins

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/platform/ServerComponents.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/platform/ServerComponents.java
@@ -19,7 +19,13 @@
  */
 package org.sonar.server.platform;
 
-import com.google.common.collect.Lists;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.Properties;
+
+import javax.annotation.Nullable;
+
 import org.sonar.api.config.EmailSettings;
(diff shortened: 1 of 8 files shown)


simonbrandhof:
To be consistent with terminology introduced in update center project, "scmRevision" should be renamed to "implementationBuild".
Moreover the JSON example is up-to-date in api/plugins/pending but not in api/plugins/installed.

sns-seb:
> To be consistent with terminology introduced in update center project, "scmRevision" should be renamed to "implementationBuild".

I agree we should be consistent, but "implementationBuild" definitly does not talk to me. I had to look into an existing jar manifest to tell what this field was about and see its value looked a lot like a Git sha1

simonbrandhof:
Yes but in theory this field is not restricted to SCM revision. If we decide to keep "scmRevision", then a ticket should be created in the update center project to keep consistency.

sns-seb:
Lets go with "implementationBuild" to avoid any confusion.

Name is poorly chosen on some aspects but at least it will make clear this JSON field is the same as the one in the manifest in jar files (and since we can not modify the manifest without huge cost, we'll stay with that name)

sns-seb:
PR update:
* rebased on master
* comments taken into account

simonbrandhof:
LGTM

PR #266, opened 2015-04-29
SONAR-6377 apply po feedback

sns-seb:
* rename status in `/plugins/available` and document it
* remove artifact from `/plugins/available` and `/plugins/updates` response (shared code)

PR #267, opened 2015-04-29
add release description and changeLogUrl in JSON response of /plugins/updates

sns-seb:
+ fix broken br tag in action description

removal of artifact from JSON response is done in PR #266


simonbrandhof:
Minor feedback about consistency of spelling : there are both "changeLogUrl" and "changelogUrl" in code and JSON. 
Else LGTM

sns-seb:
> Minor feedback about consistency of spelling : there are both "changeLogUrl" and "changelogUrl" in code and JSON. 

I believe `changeLog` is correct is this is not a single word. References from Gnu world on the wikipedia article seems to go in that direction http://en.wikipedia.org/wiki/Changelog .

PR #269, opened 2015-04-29
SONAR-6376 SONAR-6383 remove artifact from JSON response

sns-seb:
remove artifact from JSON response of WS `/plugins/installed` and `/plugins/pending`


simonbrandhof:
PluginWsCommons#writeArtifact() is unused and should be dropped. Moreover some examples still contain "artifact" field (/available and /updates)

simonbrandhof:
then LGTM

PR #270, opened 2015-04-29
SONAR-6383 leave only key,name,version in JSON response

sns-seb:
this PR will conflict with PR #263, I know, but I will handle that when merging (this PR will have the last word for `/plugins/pending`)


simonbrandhof:
to be discussed with @bellingard but I think this change should not be applied.

sns-seb:
closing, change must not be applied

PR #271, opened 2015-04-29
SONAR-6385 remove plugin description from JSON response

sns-seb:
also rename example file from example-updates_plugins.json to example-upgrades_plugins.json


simonbrandhof:
Same feedback than #270 I think this change should not be applied.

sns-seb:
closing, change must not be applied

PR #272, opened 2015-04-30
PO feedback on navigation WS

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/ui/ViewProxy.java
```diff
@@ -272,6 +274,24 @@ public boolean acceptsAvailableMeasures(String[] availableMeasures) {
...
+    }
+    return authorized;
+  }
+
+  public boolean isUserAuthorized(ComponentDto component) {
+    boolean authorized = userRoles.length == 0;
+    for (String userRole : getUserRoles()) {
+      authorized |= (UserRole.VIEWER.equals(userRole)
+        || UserRole.USER.equals(userRole)
+        || UserSession.get().hasProjectPermissionByUuid(userRole, component.uuid()));
```
simonbrandhof: Links to unauthorized pages must not be displayed to end-users. So the condition must check only the expected permission, not UserRole.VIEWER and USER.

simonbrandhof:
Out of scope of this PR, but I don't understand why line 201 of ComponentNavigationAction uses L10n bundle to get dashboard name. Are dashboards really localized ??
```
i18n.message(locale, String.format("dashboard.%s.name", dashboard.getName())
```

server/sonar-server/src/main/java/org/sonar/server/ui/ViewProxy.java
jblievremont: This condition was translated directly from the equivalent Rails code. After validation with @bellingard I created http://jira.codehaus.org/browse/SONAR-6525 and will fix this before merging.

jblievremont:
Yep, dashboards (at least core dashboards) are localized, see core.properties:

    dashboard.Dashboard.name=Main Dashboard
    dashboard.Hotspots.name=Hotspots
    dashboard.Reviews.name=Reviews
    dashboard.Issues.name=Issues Report
    dashboard.TimeMachine.name=Time Machine

simonbrandhof:
ok, so LGTM

PR #275, opened 2015-05-04
SONAR-6365 consistent properties for plugin in JSON response

sns-seb:
use common code to write plugin properties in JSON response of `/api/system/upgrades`


simonbrandhof:
LGTM

PR #276, opened 2015-05-04
New Java WS to search users

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/es/NewIndex.java
```diff
@@ -175,6 +176,7 @@ public void build() {
     private final NewIndexType indexType;
     private final String fieldName;
     private boolean sortable = false, wordSearch = false, gramSearch = false, docValues = false, disableSearch = false;
+    private SortedMap<String, Object> subFields = Maps.newTreeMap();
```
sns-seb: out of curiosity, why a SortedMap here?

jblievremont: I got the definitive explanation by @simonbrandhof : we want to be able to detect changes in the settings of an index, to drop and rebuild it when necessary. The detection of modifications is based on an order sensitive hashing algorithm (SHA256), so order of settings must be preserved - hence the use of SortedMap. I will enforce it in the ``addSubField`` method.

server/sonar-server/src/main/java/org/sonar/server/user/index/UserIndex.java
```diff
@@ -94,10 +115,38 @@ public UserDoc getByLogin(String login) {
...
+        UserIndexDefinition.FIELD_LOGIN + "." + UserIndexDefinition.SEARCH_SUB_SUFFIX,
+        UserIndexDefinition.FIELD_NAME,
+        UserIndexDefinition.FIELD_NAME + "." + UserIndexDefinition.SEARCH_SUB_SUFFIX)
+        .operator(MatchQueryBuilder.Operator.AND);
+    }
+
+    request.setQuery(QueryBuilders.filteredQuery(query,
+      userFilter));
+
+    return new SearchResult<UserDoc>(request.get(), DOC_CONVERTER);
```
teryk: UserDoc is uneeded, return new SearchResult<> is enough

server/sonar-server/src/main/java/org/sonar/server/user/index/UserIndex.java
```diff
@@ -94,10 +115,38 @@ public UserDoc getByLogin(String login) {
...
+    SearchRequestBuilder request = esClient.prepareSearch(UserIndexDefinition.INDEX)
+      .setTypes(UserIndexDefinition.TYPE_USER)
+      .setSize(options.getLimit())
+      .setFrom(options.getOffset())
+      .addSort(UserIndexDefinition.FIELD_NAME, SortOrder.ASC);
+
+    BoolFilterBuilder userFilter = FilterBuilders.boolFilter()
+      .must(FilterBuilders.termFilter(UserIndexDefinition.FIELD_ACTIVE, true));
+
+    QueryBuilder query = null;
```
teryk: null is uneeded. a personal preference ?

server/sonar-server/src/main/java/org/sonar/server/user/ws/SearchAction.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  private void writeIfNeeded(JsonWriter json, String value, String field, @Nullable List<String> fields) {
+    if (fieldIsWanted(field, fields)) {
+      json.prop(field, value);
+    }
+  }
+
+  private boolean fieldIsWanted(String field, List<String> fields) {
+    return fields == null || fields.isEmpty() || fields.contains(field);
```
teryk: you confirm you want the field if fields == null or fields.isEmpty() ?

teryk:
in UserWs, don't you think it would be better to have only BaseUsersWsAction and no defineSearchAction(controller) & defineDeactivateAction(controller) ?

teryk:
minor feedback, else 
lgtm-kim-jong

server/sonar-server/src/main/java/org/sonar/server/user/index/UserIndex.java
jblievremont: Yes, explicit preferred to implicit.

server/sonar-server/src/main/java/org/sonar/server/user/ws/SearchAction.java
jblievremont: Yep, fields is null when the request parameter is not set.

PR #279, opened 2015-05-04
User modification Web Services

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/user/ws/ChangePasswordAction.java
```diff
@@ -0,0 +1,79 @@
...
+      .setDescription("User login")
+      .setRequired(true)
+      .setExampleValue("myuser");
+
+    action.createParam(PARAM_PASSWORD)
+      .setDescription("New password")
+      .setRequired(true)
+      .setExampleValue("mypassword");
+
+    action.createParam(PARAM_PASSWORD_CONFIRMATION)
```
simonbrandhof: This pattern is pure UI. I don't think it should be applied in WS (I know, it was the case in previous version).

jblievremont: So this check should be done on UI side, right? @stas-vilchik do you agree? In this case we can keep only the ``password`` parameter for this WS.

Note that the Java code which checks that both passwords are equal can only be dropped when the Rails UI is removed.

stas-vilchik: I don't agree, guys. As any other validation, checking the password and its confirmation should be done on server side.

Of course that doesn't mean that we cannot have an additional check in web part.

simonbrandhof: Definitely not.
Password confirmation is a UX pattern used to avoid users typing unwanted passwords. That's all. No link with security concerns in backend.
On the same topic the web service could require the parameter "old password". This is a security point and would make sense on backend. Note that it's not applicable in the current WS as it targets system administrators to edit user passwords that they don't know.

jblievremont: +1 for double input validation only on client side.

I have already discussed the opportunity of allowing non-admin users to use this WS to change their accounts, the subject is a bit more tricky than just a new parameter for the old password and is out of scope for this sprint.

simonbrandhof: @jblievremont yes I know. It was more an example to illustrate the difference between passwd confirmation and old passwd parameters.

simonbrandhof:
LGTM

PR #281, opened 2015-05-05
implement security headers

ustramooner:
would like to make a patch request for allowing various security related (in the sense that security software suggests it).

Implemented are: X-Frame-Options, Xss-Protection and X-Content-Type-Options security headers

here is turned on by default but perhaps better to leave off by default

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/platform/SecureServletFilter.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/platform/SecureServletFilter.java
@@ -1,0 +1,242 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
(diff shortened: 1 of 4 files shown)


simonbrandhof:
Thanks for the contrib. What's the value to allow configuration of these headers ?
FYI there's already a similar filter here: https://github.com/SonarSource/sonarqube/blob/master/server/sonar-server/src/main/java/org/sonar/server/platform/SecurityServletFilter.java

ustramooner:
ok - i didn't see that. probably wasn't in the 4.5 branch.

the other 2 headers are recommended by ZAP which i was using. thought configurabilitiy would benice.

thoughts on forward porting?

simonbrandhof:
According to https://www.owasp.org/index.php/List_of_useful_HTTP_headers it definitely makes sense to define these headers in version 5.2.  No need for configuration IMO.

simonbrandhof:
Being applied to master branch, see https://github.com/SonarSource/sonarqube/pull/295.

PR #282, opened 2015-05-05
CSRF implementaiton

ustramooner:
CSRF implementation throughout

may be too late to put these changes into 4.5 due to the amount of testing that should be done.

i will try and make a patch against master if you are interested?


simonbrandhof:
Thanks for the contribution but it can't be applied:
- only blocker issues are fixed on LTS branch. Such improvements should be implemented on master branch.
- rails is being dropped. CSRF protection should be implemented in JS app. Here is the related JIRA ticket: https://jira.codehaus.org/browse/SONAR-5040

PR #283, opened 2015-05-05
fix deprecated calls

ustramooner:
very small fix to avoid the server logging this deprecation constantly


simonbrandhof:
Rails 2.3 is used, so correct parameter is http_only. See http://stackoverflow.com/a/72621/229031.

simonbrandhof:
in fact the real answer is that RoR is being dropped so only blocker changes are applied to existing code. Thanks.

PR #284, opened 2015-05-05
fonts should return a valid mime type

ustramooner:


simonbrandhof:
Good catch. 
Spec is now a recommendation and is officially application/font-woff. Master has been fixed. Thanks

PR #287, opened 2015-05-05
SONAR-6469 New Java WS to deactivate a user

jblievremont:


server/sonar-server/src/main/java/org/sonar/server/user/ws/DeactivateAction.java
```diff
@@ -0,0 +1,86 @@
...
+    String login = request.mandatoryParam(PARAM_LOGIN);
+    userUpdater.deactivateUserByLogin(login);
+
+    writeResponse(response, login);
+  }
+
+  private void writeResponse(Response response, String login) {
+    UserDoc user = index.getByLogin(login);
+    JsonWriter json = response.newJsonWriter().beginObject();
+    writeUser(json, user);
```
simonbrandhof: not sure that it was needed to return the user informations, but ok, it's done :-)

simonbrandhof:
LGTM

server/sonar-server/src/main/java/org/sonar/server/user/ws/DeactivateAction.java
jblievremont: Actually, the Rails WS returned an empty hash. But at least, it gives a straightforward way to show that the user is indeed not active anymore :)

PR #291, opened 2015-05-05
feature/fix_mssql_role

simonbrandhof:
When permission db_owner is not granted, then two INSERT requests are incomplete and set NULL values instead of booleans.
See https://jira.codehaus.org/browse/SONAR-6529 and https://jira.codehaus.org/browse/SONAR-6530


SonarTech:

julienlancelot:
LGTM

PR #295, opened 2015-05-06
SONAR-6537 Add HTTP security headers

simonbrandhof:
See https://jira.codehaus.org/browse/SONAR-6537


SonarTech:

jblievremont:
It's strange, although I can see the headers added to the response object, they are not written to the response stream. Still digging to understand this behavior.

simonbrandhof:
The reason is that Rack probably resets headers.

SonarTech:

simonbrandhof:
It works well except for static resources like images. I wonder if it can be a vulnerability for mime-sniffing.
Anyway LGTM

PR #296, opened 2015-05-07
Compute DSM measures

julienlancelot:


server/sonar-server/src/main/java/org/sonar/server/computation/design/DsmDataBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+import org.sonar.graph.DsmCell;
+import org.sonar.server.computation.step.ComponentUuidsCache;
+import org.sonar.server.design.db.DsmDb;
+
+public final class DsmDataBuilder {
+
+  private final Dsm<Integer> dsm;
+  private final ComponentUuidsCache uuidByRef;
+
+  private final DsmDb.Data.Builder dsmBuilder;
```
simonbrandhof: Variable should be renamed. It's misleading with DsmDataBuilder. I suggest dbBuilder or something like that. It should show that it relates to DB.

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistReportMeasuresStepTest.java
```diff
@@ -60,12 +64,12 @@
...
   MeasureDao measureDao;
 
-  PersistMeasuresStep sut;
+  PersistReportMeasuresStep sut;
 
   private BatchReport.Component component;
 
   @Before
-  public void setUp() {
+  public void setUp() throws Exception {
```
simonbrandhof: Lots of ```throws Exception``` added.

server/sonar-server/src/main/java/org/sonar/server/design/db/DsmDataEncoder.java
```diff
@@ -0,0 +1,74 @@
...
+  private DsmDataEncoder() {
+    // Only static methods
+  }
+
+  /**
+   * Serialize and compress protobuf message {@link org.sonar.server.design.db.DsmDb.Data
+   */
+  public static byte[] encodeSourceData(DsmDb.Data data) {
+    ByteArrayOutputStream byteOutput = new ByteArrayOutputStream();
+    LZ4BlockOutputStream compressedOutput = new LZ4BlockOutputStream(byteOutput);
```
simonbrandhof: I would not compress in this column (yet).

server/sonar-server/src/main/java/org/sonar/server/design/ws/DsmAction.java
```diff
@@ -0,0 +1,183 @@
...
+      .setResponseExample(Resources.getResource(getClass(), "example-dsm.json"));
+
+    // TODO add component key param
+    action.createParam(PARAM_COMPONENT_UUID)
+      .setDescription("Component uuid (only sub project or directory are allowed for the moment)")
+      .setRequired(true)
+      .setExampleValue("2312cd03-b514-4acc-94f4-5c5e8e0062b2");
+
+    action.createParam(PARAM_DISPLAY_NO_DEP)
+      .setDescription("Display component with no dependencies or not")
```
simonbrandhof: If I correctly understand, by default the components that don't have any dependencies on other files are removed from DSM response. What's the reason to hide them ?

server/sonar-server/src/main/java/org/sonar/server/design/ws/DsmAction.java
```diff
@@ -0,0 +1,183 @@
...
+
+    DbSession session = dbClient.openSession(false);
+    try {
+      ComponentDto component = dbClient.componentDao().getByUuid(session, componentUuid);
+      UserSession.get().checkProjectUuidPermission(UserRole.USER, component.projectUuid());
+
+      // TODO manage old dsm measure
+      DsmDb.Data dsmData = loadDsmData(session, component.key());
+      List<ComponentDto> components;
+      if (!displayComponentWithoutDep) {
```
simonbrandhof: variable can be inlined

server/sonar-server/src/main/java/org/sonar/server/measure/ServerMetrics.java
```diff
@@ -0,0 +1,49 @@
...
+
+package org.sonar.server.measure;
+
+import com.google.common.collect.ImmutableList;
+import org.sonar.api.measures.CoreMetrics;
+import org.sonar.api.measures.Metric;
+
+import java.util.List;
+
+public class ServerMetrics {
```
simonbrandhof: A javadoc is deserved to explain that these metrics are not supposed to be known by API. It's a hack for storage of internal data. Isn't it ?

server/sonar-server/src/main/java/org/sonar/server/util/EmptyIterator.java
```diff
@@ -0,0 +1,34 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.server.util;
+
+public class EmptyIterator<O> extends CloseableIterator<O> {
```
simonbrandhof: A javadoc is deserved here to simply explain that it iterates over an empty collection.

server/sonar-server/src/main/java/org/sonar/server/computation/design/DsmDataBuilder.java
julienlancelot: Indeed

server/sonar-server/src/main/java/org/sonar/server/util/cache/DiskCacheById.java
```diff
@@ -0,0 +1,96 @@
...
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.ObjectOutputStream;
+import java.io.Serializable;
+
+/**
+ * Serialize and deserialize objects on disk in different files identified by an id. No search capabilities, only traversal (full scan).
+ */
+public class DiskCacheById<O extends Serializable> {
```
simonbrandhof: I have the feeling that DiskCache and DiskCacheById will have to be replaced later in the sprint. Let's keep this implementation for now.

server/sonar-server/src/test/java/org/sonar/server/computation/ReportTester.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Rule;
+import org.junit.rules.ExternalResource;
+import org.junit.rules.TemporaryFolder;
+import org.sonar.batch.protocol.Constants;
+import org.sonar.batch.protocol.output.BatchReport;
+import org.sonar.batch.protocol.output.BatchReportWriter;
+
+import java.io.File;
+
+public class ReportTester extends ExternalResource {
```
simonbrandhof: this class is not used and can be dropped.

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistReportMeasuresStepTest.java
julienlancelot: Yes, it's strange, I do not know how they appear ! I'll remove them.

server/sonar-server/src/main/java/org/sonar/server/design/db/DsmDataEncoder.java
julienlancelot: Oh, I wasn't aware of this method ! 
Then indeed, this class can be dropped.

server/sonar-server/src/main/resources/org/sonar/server/design/ws/example-dsm.json
```diff
@@ -0,0 +1,14 @@
+{
```
simonbrandhof: this example deserves at least two or three components, else it's hard to understand the concept

server/sonar-server/src/main/java/org/sonar/server/design/ws/DsmAction.java
julienlancelot: Because displaying them makes the DSM unreadable when there's a lof of directories / files. Currently in the UI there's a checkbox to display them.

server/sonar-server/src/main/java/org/sonar/server/design/ws/DsmAction.java
julienlancelot: ok

server/sonar-server/src/main/java/org/sonar/server/measure/ServerMetrics.java
julienlancelot: Indeed, the goal is to not been able to compute measures related on these metrics. I do not understand why you talk of hack ?
If we put this metric in CoreMetrics, everyone will be able to use it.

server/sonar-server/src/main/java/org/sonar/server/util/EmptyIterator.java
julienlancelot: Ok

server/sonar-server/src/main/java/org/sonar/server/util/cache/DiskCacheById.java
julienlancelot: Ok, no pb

server/sonar-server/src/test/java/org/sonar/server/computation/ReportTester.java
julienlancelot: Good catch !

server/sonar-server/src/main/resources/org/sonar/server/design/ws/example-dsm.json
julienlancelot: exact

server/sonar-server/src/test/java/org/sonar/server/computation/design/DsmDataBuilderTest.java
```diff
@@ -0,0 +1,89 @@
...
+    assertThat(result.getUuidCount()).isEqualTo(0);
+    assertThat(result.getCellCount()).isEqualTo(0);
+  }
+
+  /**
+   * / | A | B | C |
+   * A |   | 2 | 4 |
+   * B | 1 |   |   |
+   * C |   |   |   |
+   */
```
simonbrandhof: this comment is very useful !

simonbrandhof:
You should check that batch report does not declare dependencies with weight<1

server/sonar-server/src/main/java/org/sonar/server/measure/ServerMetrics.java
simonbrandhof: The term "hack" was a too strong. In fact it should be named "InternalMetrics"

simonbrandhof: and it should contain the deprecated CoreMetrics#DUPLICATIONS_DATA

PR #297, opened 2015-05-07
SONAR-6470 New Java WS to show a user's groups

jblievremont:


julienlancelot:
No pipeline ?

server/sonar-server/src/main/java/org/sonar/server/user/ws/GroupsAction.java
```diff
@@ -0,0 +1,140 @@
...
+      .setDescription("If specified, only show groups whose name contains the query.")
+      .setExampleValue("user");
+
+    action.addPagingParams(25);
+  }
+
+  @Override
+  public void handle(Request request, Response response) throws Exception {
+    String login = request.mandatoryParam(PARAM_LOGIN);
+    UserDto user = dbClient.userDao().selectActiveUserByLogin(login);
```
julienlancelot: Why not use selectByLogin() ? It will prevent you to do the nullable check. And if an none active user is selected, no groups will be returned.

julienlancelot: I think you should also use method using a session in parameter, it will prevent opening a new session on each SQL query

server/sonar-server/src/main/java/org/sonar/server/user/ws/GroupsAction.java
```diff
@@ -0,0 +1,140 @@
...
+      .membership(getMembership(selected))
+      .pageIndex(page)
+      .pageSize(pageSize)
+      .build();
+
+    int total = dbClient.groupMembershipDao().countGroups(query, user.getId());
+
+    Paging paging = Paging.create(pageSize, page, total);
+
+    List<GroupMembershipDto> groups = dbClient.groupMembershipDao().selectGroups(query, user.getId(), paging.offset(), pageSize);
```
julienlancelot: You could use org.sonar.server.component.ws.SearchOptions

server/sonar-server/src/main/java/org/sonar/server/user/ws/GroupsAction.java
```diff
@@ -0,0 +1,140 @@
...
+      json.beginObject()
+        .prop("name", group.getName())
+        .prop("description", group.getDescription())
+        .prop("selected", group.getUserId() != null)
+        .endObject();
+    }
+    json.endArray();
+  }
+
+  private void writePaging(JsonWriter json, Paging paging) {
```
julienlancelot: It's already provided by org.sonar.server.component.ws.SearchOptions

server/sonar-server/src/test/java/org/sonar/server/user/ws/GroupsActionTest.java
```diff
@@ -0,0 +1,202 @@
...
+import org.sonar.server.user.db.GroupDao;
+import org.sonar.server.user.db.UserDao;
+import org.sonar.server.user.db.UserGroupDao;
+import org.sonar.server.ws.WsTester;
+
+@RunWith(MockitoJUnitRunner.class)
+public class GroupsActionTest {
+
+  @ClassRule
+  public static final DbTester dbTester = new DbTester();
```
julienlancelot: :+1:

sonar-core/src/main/java/org/sonar/core/user/GroupMembershipDao.java
```diff
@@ -47,6 +48,16 @@ public GroupMembershipDao(MyBatis mybatis) {
     }
   }
 
+  public int countGroups(GroupMembershipQuery query, Long userId) {
```
julienlancelot: I would prefer to add the session in parameter. If all methods of all daos were using session, we could remove the dependency on MyBatis from Daos.

julienlancelot:
Apply my feedback, then LGTM !

server/sonar-server/src/main/java/org/sonar/server/user/ws/GroupsAction.java
jblievremont: You mean org.sonar.server.es.SearchOptions? I don't like the idea of using an "ES" oriented component for RDBMS search, but maybe it can be refactored in some way.

julienlancelot: Indeed, wrong package ! And indeed, forget it, we won't add a dependency to an es package

PR #298, opened 2015-05-08
WS api/projects/provisioned - SONAR-6526

teryk:


server/sonar-server/src/main/java/org/sonar/server/component/ws/ProvisionedProjectsAction.java
```diff
@@ -0,0 +1,129 @@
...
+      request.mandatoryParamAsInt(Param.PAGE_SIZE)
+      );
+    List<String> desiredFields = desiredFields(request);
+    String query = request.param(Param.TEXT_QUERY);
+
+    DbSession dbSession = dbClient.openSession(false);
+    List<ComponentDto> projects;
+    int nbOfProjects;
+    try {
+      projects = dbClient.componentDao().selectProvisionedProjects(dbSession, options, query);
```
sns-seb: query parameter here come directly from request and is used as is as a SQL parameter
*SQL injection Warning*
One must ask itself whether parameters are sanitized by ibatis before using such pattern?

server/sonar-server/src/main/java/org/sonar/server/component/ws/ProvisionedProjectsAction.java
```diff
@@ -0,0 +1,129 @@
...
+import org.sonar.server.user.UserSession;
+
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+
+public class ProvisionedProjectsAction implements ProjectsAction {
+  private final DbClient dbClient;
+
+  private static final String[] POSSIBLE_FIELDS = new String[] {"uuid", "key", "name", "creationDate"};
```
sns-seb: what's the point of storing an array in a constant when its only used as a list?

server/sonar-server/src/main/java/org/sonar/server/component/ws/ProvisionedProjectsAction.java
```diff
@@ -0,0 +1,129 @@
...
+    UserSession.get().checkGlobalPermission(UserRole.ADMIN, "You need admin rights.");
+    SearchOptions options = new SearchOptions().setPage(
+      request.mandatoryParamAsInt(Param.PAGE),
+      request.mandatoryParamAsInt(Param.PAGE_SIZE)
+      );
+    List<String> desiredFields = desiredFields(request);
+    String query = request.param(Param.TEXT_QUERY);
+
+    DbSession dbSession = dbClient.openSession(false);
+    List<ComponentDto> projects;
```
sns-seb: having uninitialized variables is a code smell imo
use a method for all code after the finally and call that method from inside the try, you will save lines of code and avoid writing ambiguous code

server/sonar-server/src/main/java/org/sonar/server/component/ws/ProvisionedProjectsAction.java
teryk: yep, and my assumption was that it was.

server/sonar-server/src/main/java/org/sonar/server/component/ws/ProvisionedProjectsAction.java
teryk: no point at all, I replace it

server/sonar-server/src/main/java/org/sonar/server/component/ws/ProvisionedProjectsAction.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  private List<String> desiredFields(Request request) {
+    List<String> desiredFields = request.paramAsStrings(Param.FIELDS);
+    if (desiredFields == null) {
+      desiredFields = Arrays.asList(POSSIBLE_FIELDS);
+    }
+
+    return desiredFields;
```
sns-seb: ok, so If I understand correctly, when field parameter exists in the request, there is no validation nor indirection on the field names?

This is hardly desirable:
1. this WS being public, we are not free anymore to change column names as we like
2. client can specify any column name, even if its internal, apparently only the supported ones are written to the response but those unsupported will still be retrieved from DB

server/sonar-server/src/main/java/org/sonar/server/language/ws/ListAction.java
```diff
@@ -52,7 +52,7 @@ public ListAction(Languages languages) {
 
   @Override
   public void handle(Request request, Response response) throws Exception {
-    String query = request.param("q");
+    String query = request.param(WebService.Param.TEXT_QUERY);
```
sns-seb: missing import of WebService.PARAM here

server/sonar-server/src/main/java/org/sonar/server/component/ws/ProvisionedProjectsAction.java
teryk: +1 for saving lines and being less ambiguous. I'm not sure it would apply if the computation after finally was costly. Anyway, I apply your feedback.

server/sonar-server/src/main/java/org/sonar/server/component/ws/ProvisionedProjectsAction.java
teryk: nope, it's not the behavior. Look at the writeProjects method and you'll see that only fields specified internally are writen in the json response.
Regarding the DB, all fields are retrieved, no partial loading. It's in the json response that there's a filter.

server/sonar-server/src/main/java/org/sonar/server/language/ws/ListAction.java
teryk: ok, updated everywhere it's called

server/sonar-server/src/test/java/org/sonar/server/component/db/ComponentDaoTest.java
```diff
@@ -84,9 +88,9 @@ public void get_by_uuid() {
 
   @Test
   public void get_by_uuid_on_technical_project_copy() {
-    setupData("shared");
+    db.prepareDbUnit(getClass(), "shared.xml");
```
sns-seb: lots of duplication could be saved by moving this statement at the end of the `@Before` method

sns-seb: forget it, just noticed that test methods with various initial dbunit setup are mixed in this class
my comment does not apply

teryk: I haven't moved in the @Before method as it's not used in all the test. I have extract a method "loadBasicDataInDatabase()"

server/sonar-server/src/main/java/org/sonar/server/component/db/ComponentDao.java
```diff
@@ -172,4 +177,22 @@ protected void doDeleteByKey(DbSession session, String key) {
     return mapper(session).selectProjectsFromView("%." + viewUuid + ".%", projectViewUuid);
   }
 
+  public List<ComponentDto> selectProvisionedProjects(DbSession session, SearchOptions searchOptions, @Nullable String query) {
+    Map<String, String> parameters = new HashMap<>();
+    parameters.put("qualifier", Qualifiers.PROJECT);
+    if (query != null) {
+      parameters.put("query", "%" + query + "%");
+    }
+    return session.selectList("org.sonar.core.component.db.ComponentMapper.selectProvisionedProjects", parameters,
```
sns-seb: is there any way to avoid having such constants in code? 
Can't you use  `ComponentMapper#selectProvisionedProjects`?

server/sonar-server/src/main/java/org/sonar/server/component/ws/ProvisionedProjectsAction.java
sns-seb: you're right, nothing of my comment applies, I mixed the query parameter and the desiredField parameter, sorry

server/sonar-server/src/test/java/org/sonar/server/component/db/ComponentDaoTest.java
sns-seb: since `ComponentDao` exposes quite a lot of methods which can be semantically grouped, a good refactoring could be to split the class in smaller ones or at least split the unit test. In both cases, it will then be possible to move the `db.prepareUnit` call to `@Before` method

sns-seb: but it's out of scope of this PR

server/sonar-server/src/test/java/org/sonar/server/component/ws/ProvisionedProjectsActionTest.java
```diff
@@ -0,0 +1,138 @@
...
+    dbSession = dbClient.openSession(false);
+    db.truncateTables();
+    ws = new WsTester(new ProjectsWs(new ProvisionedProjectsAction(dbClient)));
+  }
+
+  @Test
+  public void all_provisioned_projects_without_analyzed_projects() throws Exception {
+    MockUserSession.set().setGlobalPermissions(UserRole.ADMIN);
+    ComponentDto analyzedProject = ComponentTesting.newProjectDto("analyzed-uuid-1");
+    dbClient.componentDao().insert(dbSession, newProvisionedProject("1"), newProvisionedProject("2"));
```
sns-seb: (duplication) dbClient.componentDao can be moved to a field

server/sonar-server/src/test/java/org/sonar/server/component/ws/ProvisionedProjectsActionTest.java
```diff
@@ -0,0 +1,138 @@
...
+    analyzedProject = dbClient.componentDao().insert(dbSession, analyzedProject);
+    SnapshotDto snapshot = SnapshotTesting.createForProject(analyzedProject);
+    dbClient.snapshotDao().insert(dbSession, snapshot);
```
sns-seb: (duplication) dbClient.snapshotDao can be moved to a field

server/sonar-server/src/main/java/org/sonar/server/component/db/ComponentDao.java
teryk: quite surprisingly for me, it works if you have a map and a RowBounds object as a parameter. I thought it wouldn't work. I was wrong. Thanks !

server/sonar-server/src/test/java/org/sonar/server/component/ws/ProvisionedProjectsActionTest.java
```diff
@@ -0,0 +1,138 @@
...
+import org.sonar.server.component.SnapshotTesting;
+import org.sonar.server.component.db.ComponentDao;
+import org.sonar.server.component.db.SnapshotDao;
+import org.sonar.server.db.DbClient;
+import org.sonar.server.user.MockUserSession;
+import org.sonar.server.ws.WsTester;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ProvisionedProjectsActionTest {
```
sns-seb: I believe every unit test of a WS action should have a test using its json example as expected JSON.
this practice enforces that the example is consistent and up to date with code

sns-seb: for this PR, it would replace with benefit the "all_provisioned_projects_without_analyzed_projects.json" file uses in the first test

server/sonar-server/src/test/java/org/sonar/server/component/ws/ProvisionedProjectsActionTest.java
teryk: I think you're talking about componentDao and yes, it could be moved to a field

sonar-core/src/main/resources/org/sonar/core/component/db/ComponentMapper.xml
```diff
@@ -227,5 +227,30 @@
...
+  <select id="countProvisionedProjects" parameterType="map" resultType="int">
+    select count(p.id)
+    from projects p
+    <include refid="provisionClauses"/>
+  </select>
+
+  <sql id="provisionClauses">
+    left join snapshots s on s.project_id=p.id
+    where
+    s.id is null
```
sns-seb: shouldn't "s.is is null" be moved to the join statement for clarity?
(not sure, I'm not very much used to the old/stantard SQL join syntax)

server/sonar-server/src/main/java/org/sonar/server/component/ws/ProvisionedProjectsAction.java
sns-seb: after verification, sanitization is applied if `#{}` syntax is used for parameters, which is the case here => all good

sonar-core/src/main/resources/org/sonar/core/component/db/ComponentMapper.xml
teryk: not so sure about it. There's pro's and con's. In the join statements you're supposed to put join conditions. In the where clause, the filtering occurs. That's how I separate responsabilities

sns-seb:
please add a unit test that uses the WS action's example, then LGTM

server/sonar-server/src/test/java/org/sonar/server/component/ws/ProvisionedProjectsActionTest.java
teryk: I think it's a good idea, I can see how fields related to date can be cumbersome to test though. I give it a try and see how it goes.

teryk: done

PR #299, opened 2015-05-11
make UserSession non static

sns-seb:
* split `UserSession` into `ThreadLocalUserSession` and `ServerUserSession`, former is responsible of using a ThreadLocal to retrieve the current session, later is responsible for the dao related logic, both implement `UserSession` interface
* reworked `MockUserSession` to drop dao mocks and introduced `AnonumousMockUserSession`
* introduced `UserSessionRule` (itself a `UserSession`) with support for `ServerTest` to integrate nicely with the `ThreadLocalUserSession` instance in the pico container


server/sonar-server/src/test/java/org/sonar/server/tester/UserSessionRule.java
```diff
@@ -0,0 +1,318 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * {@code UserSessionRule} is intended to be used as a {@link org.junit.Rule} to easily manage {@link UserSession} in
+ * unit tests.
+ * <p>
+ * It can be used as a {@link org.junit.ClassRule} but be careful not to modify its states from inside tests methods
+ * unless you purposily want to have side effects between each tests.
```
simonbrandhof: typo: "purposely" !

server/sonar-server/src/test/java/org/sonar/server/tester/UserSessionRule.java
```diff
@@ -0,0 +1,318 @@
...
+  @CheckForNull
+  private final ServerTester serverTester;
+  private UserSession currentUserSession;
+
+  private UserSessionRule(@Nullable ServerTester serverTester) {
+    this.serverTester = serverTester;
+    anonymous();
+  }
+
+  public static UserSessionRule standalone() {
```
simonbrandhof: The name standalone() could be improved. IMO it does not highlight the fact that it's a factory method. What about create() ?

server/sonar-server/src/test/java/org/sonar/server/tester/UserSessionRule.java
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  public static UserSessionRule forServerTester(ServerTester serverTester) {
+    return new UserSessionRule(Preconditions.checkNotNull(serverTester));
+  }
+
+  /**
+   * Log on with the default login {@link #DEFAULT_LOGIN}
+   */
+  public UserSessionRule logon() {
```
simonbrandhof: I suppose that this method was added according to all the tests you touched ? Else I'm not sure about the value of this shortcut regarding to logon(String)

server/sonar-server/src/test/java/org/sonar/server/tester/UserSessionRule.java
```diff
@@ -0,0 +1,318 @@
...
+    return logon(DEFAULT_LOGIN);
+  }
+
+  /**
+   * Log on with the specified login
+   */
+  public UserSessionRule logon(String login) {
```
simonbrandhof: Minor - I would naturally prefer the verb "log in" instead of "log on". That's the term used in the webapp.

sns-seb: I would prefer that too, but someone decided that getter convention wasn't right and decided to use `login` instead of `getLogin` as the name of the method to get the login of the current `UserSession`. I used `logon` to avoid unsolvable conflict and avoid confusion.

btw: isn't there a rule about method named exactly the same as a property? or is that only in findbug and/or disabled on SQ?

simonbrandhof: no pb for renaming ```login()``` to ```getLogin()```.
This convention was initially introduced for APIs but there are too many frequent debates about that.

server/sonar-server/src/test/java/org/sonar/server/tester/UserSessionRule.java
sns-seb: my intent is the following:
1. most code do not care about the actual login of the user therefor this avoids having to specify a String in tones of unit tests where it will only be noise (and constants will often be missing there for adding code duplication, ...)
2. currently, isLoggedIn() implementation is based on the fact that a login is present or not but this is an implementation detail. What most tests do is verify the user is logged in, not that is has a login (which would be wrong, btw). Tests should express that.

server/sonar-server/src/test/java/org/sonar/server/tester/UserSessionRule.java
sns-seb: why not `create`

sns-seb: renaming `standalone()` to `create()` makes the javadoc of `UserSessionRule` hard to write
`standalone` is a convenient way to designate the usage opposed to the one with `ServerTester`

PR #301, opened 2015-05-12
Fix issue bulk change modal permission check

jblievremont:


sns-seb:
fix the import grouping, then LGTM

PR #315, opened 2015-05-20
add module concept

sns-seb:
this PR introduces the concept of module as a group of components
this will allow reducing length of the ComponentLevel classes and in the near future implement lazy start of components and/or description of implicit component dependencies
this PR also includes a few module to demo usage and as a proof-of-concept

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/batch/BatchWsModule.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/batch/BatchWsModule.java
@@ -17,35 +17,22 @@
  * along with this program; if not, write to the Free Software Foundation,
  * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
-package org.sonar.core.component;
+package org.sonar.server.batch;
 
-import org.sonar.core.platform.ComponentContainer;
+import org.sonar.core.component.Module;
+import org.sonar.server.computation.ws.SubmitReportAction;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
-public abstract class Module {
-  private ComponentContainer container;
-
-  public Module configure(ComponentContainer container) {
-    this.container = checkNotNull(container);
-
-    configureModule();
-
-    return this;
+public class BatchWsModule extends Module {
+  @Override
+  protected void configureModule() {
+    add(
+      BatchIndex.class,
+      GlobalAction.class,
+      ProjectAction.class,
+      ProjectRepositoryLoader.class,
+      SubmitReportAction.class,
+      IssuesAction.class,
+      UsersAction.class,
+      BatchWs.class);
   }
-
-  protected abstract void configureModule();
-
-  protected <T> T getComponentByType(Class<T> tClass) {
-    return container.getComponentByType(tClass);
-  }
-
-  protected void add(Object... objects) {
-    for (Object object : objects) {
-      if (object != null) {
-        container.addComponent(object, true);
-      }
-    }
-  }
-
 }
--- a/server/sonar-server/src/main/java/org/sonar/server/db/migrations/MigrationStepModule.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/db/migrations/MigrationStepModule.java
@@ -19,7 +19,7 @@
  */
 package org.sonar.server.db.migrations;
 
-import com.google.common.collect.ImmutableList;
+import org.sonar.core.component.Module;
 import org.sonar.server.db.migrations.v36.ViolationMigrationStep;
 import org.sonar.server.db.migrations.v42.CompleteIssueMessageMigrationStep;
 import org.sonar.server.db.migrations.v42.PackageKeysMigrationStep;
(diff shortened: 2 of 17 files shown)


server/sonar-server/src/main/java/org/sonar/server/batch/BatchWsModule.java
```diff
@@ -0,0 +1,38 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.batch;
+
+import org.sonar.core.component.Module;
+import org.sonar.server.computation.ws.SubmitReportAction;
+
+public class BatchWsModule extends Module {
+  @Override
+  protected void configureModule() {
```
SonarTech: :new: MAJORProtected method 'configureModule(...)' is never used.[](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AUnusedProtectedMethod)

server/sonar-server/src/main/java/org/sonar/server/search/EsSearchModule.java
```diff
@@ -0,0 +1,33 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.search;
+
+import org.sonar.core.component.Module;
+import org.sonar.server.es.EsClient;
+
+public class EsSearchModule extends Module {
+  @Override
+  protected void configureModule() {
```
SonarTech: :new: MAJORProtected method 'configureModule(...)' is never used.[](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AUnusedProtectedMethod)

server/sonar-server/src/main/java/org/sonar/server/platform/platformlevel/PlatformLevel4.java
```diff
@@ -306,7 +299,7 @@ public PlatformLevel4(PlatformLevel parent, List<Object> level4AddedComponents)
   }
 
   @Override
-  public PlatformLevel configure() {
+  protected void configureLevel() {
```
SonarTech: :new: MAJORThis method has 387 lines, which is greater than the 100 lines authorized. Split it into smaller methods.[](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS138)

server/sonar-server/src/main/java/org/sonar/server/db/migrations/MigrationStepModule.java
```diff
@@ -73,78 +73,77 @@
 import org.sonar.server.db.migrations.v52.FeedProjectLinksComponentUuid;
 import org.sonar.server.db.migrations.v52.MoveProjectProfileAssociation;
 
-import java.util.List;
+public class MigrationStepModule extends Module {
+  @Override
+  protected void configureModule() {
```
SonarTech: :new: MAJORProtected method 'configureModule(...)' is never used.[](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AUnusedProtectedMethod)

SonarTech:
* 4 major

Details:
* [Protected method 'configureModule(...)' is never used.](https://github.com/SonarSource/sonarqube/blob/283df66342aa71314a090323fe65a59927fc0b05/server/sonar-server/src/main/java/org/sonar/server/batch/BatchWsModule.java#L27)[](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AUnusedProtectedMethod)
* [Protected method 'configureModule(...)' is never used.](https://github.com/SonarSource/sonarqube/blob/283df66342aa71314a090323fe65a59927fc0b05/server/sonar-server/src/main/java/org/sonar/server/db/migrations/MigrationStepModule.java#L78)[](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AUnusedProtectedMethod)
* [This method has 387 lines, which is greater than the 100 lines authorized. Split it into smaller methods.](https://github.com/SonarSource/sonarqube/blob/283df66342aa71314a090323fe65a59927fc0b05/server/sonar-server/src/main/java/org/sonar/server/platform/platformlevel/PlatformLevel4.java#L302)[](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS138)
* [Protected method 'configureModule(...)' is never used.](https://github.com/SonarSource/sonarqube/blob/283df66342aa71314a090323fe65a59927fc0b05/server/sonar-server/src/main/java/org/sonar/server/search/EsSearchModule.java#L27)[](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AUnusedProtectedMethod)

simonbrandhof:
There's a conflict in terminology. SQ also deals with "components" and "modules" when inspecting source code. 
Unfortunately I don't have any proposal.

sns-seb:
ComponentModule?

simonbrandhof:
You're talking about which module of components ? :-(

julienlancelot:
What about using 'aggregator' ?

simonbrandhof:
Let's merge and improve naming later.

sns-seb:
I don't think there is any way around it:
* "component" is a very generic name and is used in tones of contexts, so does "module"
* here, we are in the IOC context, both terms are quite common (see Dagger, Guice for "module"), I don't see how we could benefit from inventing words here

Overall and from a practical point of view, I don't think this is a problem anyway.
"Component" and "Module" will be used as part of class or interface names, I don't see how it could be confused with a component or a module as an SQ source inspection unit (why would an interface or class be used to represent the latter). Regarding other usage of "component" or "module" (in code, methods names, ...), those will occur only in precisely located places (ie. in the code and package that handles IOC)

sonar-core/src/main/java/org/sonar/core/component/Module.java
```diff
@@ -0,0 +1,63 @@
...
+    if (object != null) {
+      container.addComponent(object, singleton);
+    }
+  }
+
+  protected <T> T getComponentByType(Class<T> tClass) {
+    return container.getComponentByType(tClass);
+  }
+
+  protected void add(Object... objects) {
```
simonbrandhof: Sounds misleading to have both add(Object, boolean) and add(Object...) that implicitly sets the boolean to true. 
Moreover don't you think that these protected methods aim to duplicate ComponentContainer ? An alternative could be: ```protected abstract void configureModule(ComponentContainer container)```

sns-seb: `add(Object, boolean)` is currently never used. It created it only to provide support for the current use of `ComponentContainer.add(Object, boolean)` in `PlatformLevel`. Btw, the `ComponentContainer` method is used only once with a false parameter and I don't think it is even appropriate https://github.com/SonarSource/sonarqube/blob/master/server/sonar-server/src/main/java/org/sonar/server/platform/platformlevel/PlatformLevel4.java#L424.

I'm not very happy with the idea of `protected abstract void configureModule(ComponentContainer container)`.

First, it creates a non necessary coupling between subclasses of `Module` and the class `ComponentContainer`. Implementation only cares to add classes or objects to the module. How this objects are in turn added to the container is implementation specific. Side-benefit: if implementation is someday not based on `ComponentContainer` but, eg. on PicoContainer directly, Module subclasses will not change.

Second, `ComponentContainer` exposes to many methods to add a component to the container. You note that methods `add` of module are confusing, this is worse for `ComponentContainer` imo (look at all the add* methods). In addition, some methods have nothing to do with adding a component and they should not be called from a module.

simonbrandhof: If the parameter ComponentContainer is not added to ```configureModule()```, then ```add(Object, boolean)``` and its single usage for MeasureDao should be dropped.

sns-seb:
PR updated:
* rebased on master
* unused protected methods in `Module` removed
* fix `MeasuresDao` not being a singleton

simonbrandhof:
LGTM

PR #329, opened 2015-05-26
SONAR-6576 WS api/metrics/domains list all the available metric domains available

teryk:


SonarTech:

server/sonar-server/src/main/java/org/sonar/server/metric/ws/DomainsAction.java
```diff
@@ -0,0 +1,73 @@
...
+      json.close();
+    } finally {
+      MyBatis.closeQuietly(dbSession);
+    }
+  }
+
+  private void writeDomains(JsonWriter json, List<String> domains) {
+    json.name("domains");
+    json.beginArray();
+    for (String domain : domains) {
```
jblievremont: You could have used ``json.values(domains);`` instead.

jblievremont:
Very small feedback, and LGTM :+1:

server/sonar-server/src/main/java/org/sonar/server/metric/ws/DomainsAction.java
teryk: thanks, I didn't know about it

PR #331, opened 2015-05-26
Remove projectKey and projectSettings from ComputationContext

julienlancelot:


server/sonar-server/src/main/java/org/sonar/server/computation/component/ComponentImpl.java
```diff
@@ -83,6 +89,27 @@ public int getRef() {
     return component.getRef();
   }
 
+  public String getUuid() {
+    return uuid;
+  }
+
+  public ComponentImpl setUuid(String uuid) {
```
sns-seb: too bad the current version is mutable... but in this case, the method from interface is missing a `@CheckForNull`

server/sonar-server/src/main/java/org/sonar/server/computation/component/ComponentImpl.java
```diff
@@ -83,6 +89,27 @@ public int getRef() {
...
+    this.uuid = uuid;
+    return this;
+  }
+
+  public String getKey() {
+    if (key == null) {
+      throw new UnsupportedOperationException(String.format("Component key of ref '%s' has not be fed yet", getRef()));
```
sns-seb: if the method throws an exception when the value is not set, this should be documented in javadoc of the method in the `Component` interface

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedComponentUuidsStep.java
```diff
@@ -0,0 +1,137 @@
...
+import org.sonar.server.db.DbClient;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Read all components from the batch report and set components uuid and key
+ */
+public class FeedComponentUuidsStep implements ComputationStep {
```
sns-seb: Populate should be more appropriate than Feed, imo

server/sonar-server/src/main/java/org/sonar/server/computation/component/ComponentImpl.java
julienlancelot: I agree with you that we should use a immutable version ! But adding a @CheckForNull will force every caller to do some check and fail if null.
What about failing in the getUuid() if uuid is null ?

server/sonar-server/src/main/java/org/sonar/server/computation/component/ComponentImpl.java
julienlancelot: Indeed

server/sonar-server/src/main/java/org/sonar/server/computation/component/ComponentImpl.java
sns-seb: since you fail for `getKey` I was expecting you would do the same for `getUuid`

julienlancelot: Indeed !

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedComponentUuidsStep.java
```diff
@@ -0,0 +1,137 @@
...
+
+  private final DbClient dbClient;
+
+  public FeedComponentUuidsStep(DbClient dbClient) {
+    this.dbClient = dbClient;
+  }
+
+  @Override
+  public void execute(final ComputationContext context) {
```
sns-seb: `context` does not need to be final

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedComponentUuidsStep.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  @Override
+  public String getDescription() {
+    return "Feed components uuid";
+  }
+
+  private class ComponentDepthTraversalTypeAwareVisitor extends DepthTraversalTypeAwareVisitor {
+
+    private BatchReportReader reportReader;
```
sns-seb: should be final

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedComponentUuidsStep.java
```diff
@@ -0,0 +1,137 @@
...
+    private Map<String, String> componentUuidByKey;
```
sns-seb: should be final

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedComponentUuidsStep.java
```diff
@@ -0,0 +1,137 @@
...
+    private String branch;
```
sns-seb: should be final and annotated `@Nullable`

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedComponentUuidsStep.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private void executeForDirectoryAndFile(Component component) {
+      BatchReport.Component batchComponent = reportReader.readComponent(component.getRef());
+      // TODO fail if path is null
+      String componentKey = ComponentKeys.createEffectiveKey(nearestModule.getKey(), batchComponent.getPath());
+      feedComponent((ComponentImpl) component, componentKey);
+    }
+
+    private void feedComponent(ComponentImpl projectImpl, String componentKey) {
```
sns-seb: variable projectImpl could actually be component, right ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/ParseReportStep.java
```diff
@@ -31,39 +32,24 @@
...
     this.issueComputation = issueComputation;
-    this.computeComponentsRefCache = computeComponentsRefCache;
   }
 
   @Override
-  public void execute(ComputationContext context) {
-    int rootComponentRef = context.getReportMetadata().getRootComponentRef();
-    recursivelyProcessComponent(context, rootComponentRef);
-    processDeletedComponents(context);
+  public void execute(final ComputationContext context) {
```
sns-seb: no need to be final

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistComponentsStep.java
```diff
@@ -58,189 +53,176 @@ public PersistComponentsStep(DbClient dbClient, ComputeComponentsRefCache comput
...
+      }
+      if (!ObjectUtils.equals(existingComponent.parentProjectId(), newComponent.parentProjectId())) {
+        isUpdated = true;
+      }
+      return isUpdated;
     }
   }
 
+  private static String getFileQualifier(BatchReport.Component reportComponent) {
+    return !reportComponent.getIsTest() ? Qualifiers.FILE : Qualifiers.UNIT_TEST_FILE;
```
sns-seb: please invert test to avoid an error prone `!`

server/sonar-server/src/test/java/org/sonar/server/computation/step/PersistTestsStepTest.java
```diff
@@ -43,15 +40,20 @@
...
-import org.sonar.server.computation.language.LanguageRepository;
 import org.sonar.server.computation.component.DbComponentsRefCache;
 import org.sonar.server.computation.component.DbComponentsRefCache.DbComponent;
+import org.sonar.server.computation.component.DumbComponent;
+import org.sonar.server.computation.language.LanguageRepository;
 import org.sonar.server.db.DbClient;
 import org.sonar.server.source.db.FileSourceDao;
 import org.sonar.server.source.db.FileSourceDb;
 
+import java.io.File;
```
sns-seb: we do not group imports anymore

server/sonar-server/src/test/java/org/sonar/server/computation/step/PurgeDatastoresStepTest.java
```diff
@@ -29,15 +30,15 @@
...
 import org.sonar.server.computation.ComputationContext;
 import org.sonar.server.computation.component.ComponentTreeBuilders;
+import org.sonar.server.computation.component.DbComponentsRefCache;
 import org.sonar.server.computation.component.DumbComponent;
+import org.sonar.server.computation.component.ProjectSettingsRepository;
 import org.sonar.server.computation.language.LanguageRepository;
-import org.sonar.server.computation.component.DbComponentsRefCache;
 import org.sonar.server.db.DbClient;
 
 import java.io.File;
```
sns-seb: we do not group imports anymore

sns-seb:
too bad you didn't keep the patch for ComponentVisitor I gave you as a separate commit

sns-seb:
smal name change for PopulateComponentUuidsStep and then LGTM

PR #332, opened 2015-05-26
Remove transitive dependencies from sonar-plugin-api

simonbrandhof:
This is another step before supporting guava 18.


sonar-duplications/src/main/java/net/sourceforge/pmd/cpd/SourceCode.java
```diff
@@ -59,9 +62,7 @@
     protected abstract Reader getReader() throws Exception;
 
     protected List<String> load() {
-      LineNumberReader lnr = null;
-      try {
-        lnr = new LineNumberReader(getReader());
+      try (LineNumberReader lnr = new LineNumberReader(getReader())) {
```
julienlancelot: Jacoco will not be happy with that...

julienlancelot:
On my side it's LGTM. I let @sns-seb give its own LGTM.

sonar-duplications/src/main/java/net/sourceforge/pmd/cpd/SourceCode.java
simonbrandhof: Indeed but on purpose.

julienlancelot: Could you just tell me why then ?

sns-seb: what? Jacoco does not support try-with-resource?
That would be the reason I haven't many of them in code yet.

julienlancelot: Indeed, Jacoco does not currently "understand" try-with-resource, it badly detect 8 branches to cover.
See https://github.com/jacoco/jacoco/issues/15 for the whole story.

sns-seb:
well, I gave it a high altitude eye and LGTM

PR #336, opened 2015-05-27
refacto ce step 1

sns-seb:
this PR:
* introduces a CE specific interface `BatchReportReader` (see https://jira.codehaus.org/browse/SONAR-6590)
    * added a `BatchReportReaderRule` to easily write unit tests
    * unit tests do not create files at all when they use `BatchReportReader`
* moves `BatchReportReader` as a dependency of steps that uses it
    * this implies that `ComputationThread`, `ComputationThreadLauncher` and `ComputationService` (amount others) have been modified to create a `PicoContainer` specific for each `Queue.Item`
    * remove static call to `Platform.getInstance()` in CE code (static is evil, CE failed with a Pico error when an item was available to process the first time the CE thread ran)


server/sonar-server/src/main/java/org/sonar/server/computation/batch/CEBatchReportReader.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.List;
+import javax.annotation.CheckForNull;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.LineIterator;
+import org.sonar.batch.protocol.output.BatchReport;
+import org.sonar.server.computation.ReportQueue;
+import org.sonar.server.util.CloseableIterator;
+
+public class CEBatchReportReader implements BatchReportReader {
```
julienlancelot: Why there's no Unit Test ? Because of pending work ?

julienlancelot:
Apply my feedback and LGTM !

PR #340, opened 2015-05-29
SONAR-6586 Fix user login length check

jblievremont:


server/sonar-server/src/test/java/org/sonar/server/user/UserUpdaterTest.java
```diff
@@ -251,7 +251,7 @@ public void fail_to_create_user_with_too_short_login() {
         .setPasswordConfirmation("password"));
       fail();
     } catch (BadRequestException e) {
-      assertThat(e.errors().messages()).containsOnly(Message.of(Validation.IS_TOO_SHORT_MESSAGE, "Login", 2));
+      assertThat(e.errors().messages()).containsOnly(Message.of(Validation.IS_TOO_SHORT_MESSAGE, "Login", 3));
```
teryk: shouldn't we use the ExpectedException rule ?

server/sonar-server/src/main/java/org/sonar/server/user/UserUpdater.java
```diff
@@ -211,10 +214,10 @@ private static void checkNotEmptyParam(@Nullable String value, String param, Lis
   private static void validateLoginFormat(@Nullable String login, List<Message> messages) {
     checkNotEmptyParam(login, LOGIN_PARAM, messages);
     if (!Strings.isNullOrEmpty(login)) {
```
teryk: you already check that the login is null or empty. It would be more logical to check that you have no message. WDYT ?

server/sonar-server/src/test/java/org/sonar/server/user/UserUpdaterTest.java
teryk: same at https://github.com/SonarSource/sonarqube/pull/340/files#diff-83a2772a2e470bbeb519026e14605c32R239

server/sonar-server/src/main/java/org/sonar/server/user/UserUpdater.java
jblievremont: Well, that was not in the code I modified :)

Moreover, there is first the check for null or empty (which eventually adds a message to messages), then if login is not null or empty, other checks. So I would not change anything here.

server/sonar-server/src/test/java/org/sonar/server/user/UserUpdaterTest.java
jblievremont: Good point.

teryk:
LGTM

PR #341, opened 2015-05-29
SONAR-6259 Grant project permission on the compute engine

julienlancelot:


sns-seb:
some feedback on unit tests then LGTM

PR #352, opened 2015-06-03
Support unzipped sonarqube-application

dgageot:
mci -Ddev \
  -Dassembly.format=dir \
  -Dassembly.includeBaseDirectory=false \
  -Dassembly.checksum=false

### DIFF
--- a/sonar-application/assembly.xml
+++ b/sonar-application/assembly.xml
@@ -2,9 +2,9 @@
 <assembly>
   <id>distribution</id>
   <formats>
-    <format>zip</format>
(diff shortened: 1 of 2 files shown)


dgageot:
@simonbrandhof Can you take a look?

I didn't want to add yet another profile nor use these settings in `dev` profile because I'm afraid of the impacts. However, it makes sense for a quick dev build to not create a zip, nor compute checksums

dgageot:
Changed the `assembly.checksum` property to `checksum.failOnError`

PR #353, opened 2015-06-03
do not create ES indexes for every ServerTester instances

sns-seb:
now, to use `ServerTester` with ES indexes, one must call `ServerTester.withEsIndexes()`, only few medium test actually require ES indexes

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/es/BaseIndexer.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/es/BaseIndexer.java
@@ -36,7 +36,7 @@ public abstract class BaseIndexer implements Startable {
   private final ThreadPoolExecutor executor;
   private final String indexName, typeName, dateFieldName;
   protected final EsClient esClient;
-  private volatile long lastUpdatedAt = 0L;
+  private volatile long lastUpdatedAt = -1L;
 
   /**
    * Indexers are disabled during server startup, to avoid too many consecutive refreshes of the same index
@@ -65,6 +65,9 @@ public abstract class BaseIndexer implements Startable {
       Future submit = executor.submit(new Runnable() {
         @Override
         public void run() {
+          if (lastUpdatedAt == -1L) {
+            lastUpdatedAt = esClient.getMaxFieldValue(indexName, typeName, dateFieldName);
+          }
           if (requestedAt > lastUpdatedAt) {
             long l = task.index(lastUpdatedAt);
             // l can be 0 if no documents were indexed
@@ -98,7 +101,7 @@ public abstract class BaseIndexer implements Startable {
 
   @Override
   public void start() {
-    lastUpdatedAt = esClient.getMaxFieldValue(indexName, typeName, dateFieldName);
+    // nothing to do at startup
   }
 
   @Override
--- a/server/sonar-server/src/main/java/org/sonar/server/es/IndexDefinitions.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/es/IndexDefinitions.java
@@ -21,11 +21,10 @@ package org.sonar.server.es;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
+import java.util.Map;
 import org.elasticsearch.common.settings.Settings;
 import org.picocontainer.Startable;
 import org.sonar.api.server.ServerSide;
-
-import java.util.Map;
 
 /**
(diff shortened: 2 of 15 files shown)


server/sonar-server/src/main/java/org/sonar/server/es/BaseIndexer.java
```diff
@@ -98,7 +100,9 @@ public BaseIndexer setEnabled(boolean b) {
 
   @Override
   public void start() {
-    lastUpdatedAt = esClient.getMaxFieldValue(indexName, typeName, dateFieldName);
+    if (!settings.getBoolean("sonar.internal.es.disableIndexes")) {
```
julienlancelot: Instead of using this property here, you should use the property 'enabled' -> 
if (enabled) {
   ...
}

server/sonar-server/src/main/java/org/sonar/server/activity/index/ActivityIndexer.java
```diff
@@ -36,8 +37,8 @@
 
   private final DbClient dbClient;
 
-  public ActivityIndexer(DbClient dbClient, EsClient esClient) {
-    super(esClient, 0L, ActivityIndexDefinition.INDEX, ActivityIndexDefinition.TYPE, ActivityIndexDefinition.FIELD_CREATED_AT);
+  public ActivityIndexer(DbClient dbClient, EsClient esClient, Settings settings) {
```
julienlancelot: I think this dependency is useless

server/sonar-server/src/main/java/org/sonar/server/es/IndexDefinitions.java
```diff
@@ -101,12 +102,15 @@ public IndexDefinitions(IndexDefinition[] defs) {
   public void start() {
     // collect definitions
     IndexDefinition.IndexDefinitionContext context = new IndexDefinition.IndexDefinitionContext();
-    for (IndexDefinition definition : defs) {
-      definition.define(context);
-    }
 
-    for (Map.Entry<String, NewIndex> entry : context.getIndices().entrySet()) {
-      byKey.put(entry.getKey(), new Index(entry.getValue()));
+    if (!settings.getBoolean("sonar.internal.es.disableIndexes")) {
```
julienlancelot: Are you sure creating the definition of index cost something ?

server/sonar-server/src/main/java/org/sonar/server/search/IndexSynchronizer.java
```diff
@@ -68,6 +68,7 @@ public IndexSynchronizer(DbClient db, IndexClient index, SourceLineIndexer sourc
     this.userIndexer = userIndexer;
     this.viewIndexer = viewIndexer;
     this.activityIndexer = activityIndexer;
+    this.settings = settings;
   }
 
   public void executeDeprecated() {
```
julienlancelot: You forgot to add the check on the property here

julienlancelot:
Apply my feedback, then LGTM !

server/sonar-server/src/main/java/org/sonar/server/es/BaseIndexer.java
sns-seb: removing the Settings dependency would indeed be good, I'll look into that

server/sonar-server/src/main/java/org/sonar/server/es/IndexDefinitions.java
sns-seb: this makes the `IndexDefinitions` empty which in turn makes sure no index is created

I like the fact that if there should be no index created into ES anyway, then there is no `IndexDefinition` in this class too

server/sonar-server/src/main/java/org/sonar/server/es/BaseIndexer.java
sns-seb: `enabled` is not enough for me, I need to disable what's done in the start method too

server/sonar-server/src/main/java/org/sonar/server/es/IndexDefinitions.java
julienlancelot: Ok

server/sonar-server/src/main/java/org/sonar/server/es/BaseIndexer.java
julienlancelot: As discussed, let's move the code from the start() in index()

PR #360, opened 2015-06-09
add storage to MeasureRepository, add a MetricRepository and add qualitygate to ce

sns-seb:
this PR adds the following:
* a CE specific interface Metric and a MetricRepository
* adds the QualityGate to the CE container via the QualityGateHolder
* adds possibility to stored new Measure into MeasureRepository
    * (WARNING) Measure added to MeasureRepository are not yet persisted into DB

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/measure/BatchMeasureToMeasure.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/measure/BatchMeasureToMeasure.java
@@ -34,7 +34,7 @@ public class BatchMeasureToMeasure {
     }
 
     String data = batchMeasure.hasStringValue() ? batchMeasure.getStringValue() : null;
-    switch (metric.getMetricType().getValueType()) {
+    switch (metric.getType().getValueType()) {
       case INT:
         return toIntegerMeasure(batchMeasure, data);
       case LONG:
@@ -50,7 +50,7 @@ public class BatchMeasureToMeasure {
       case NO_VALUE:
         return toNoValueMeasure(batchMeasure);
       default:
-        throw new IllegalArgumentException("Unsupported Measure.ValueType " + metric.getMetricType().getValueType());
+        throw new IllegalArgumentException("Unsupported Measure.ValueType " + metric.getType().getValueType());
     }
   }
 
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/measure/BatchMeasureToMeasure.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/measure/BatchMeasureToMeasure.java
@@ -1,0 +1,118 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * SonarQube is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.computation.measure;
+
+import com.google.common.base.Optional;
+import java.util.Objects;
+import javax.annotation.Nullable;
+import org.sonar.batch.protocol.output.BatchReport;
+import org.sonar.server.computation.metric.Metric;
+
+public class BatchMeasureToMeasure {
+
+  public Optional<Measure> toMeasure(@Nullable BatchReport.Measure batchMeasure, Metric metric) {
+    Objects.requireNonNull(metric);
+    if (batchMeasure == null) {
+      return Optional.absent();
+    }
+
+    String data = batchMeasure.hasStringValue() ? batchMeasure.getStringValue() : null;
+    switch (metric.getMetricType().getValueType()) {
+      case INT:
+        return toIntegerMeasure(batchMeasure, data);
+      case LONG:
+        return toLongMeasure(batchMeasure, data);
+      case DOUBLE:
+        return toDoubleMeasure(batchMeasure, data);
+      case BOOLEAN:
+        return toBooleanMeasure(batchMeasure, data);
+      case STRING:
+        return toStringMeasure(batchMeasure);
+      case LEVEL:
+        return toLevelMeasure(batchMeasure);
+      case NO_VALUE:
+        return toNoValueMeasure(batchMeasure);
+      default:
+        throw new IllegalArgumentException("Unsupported Measure.ValueType " + metric.getMetricType().getValueType());
+    }
+  }
+
+  private static Optional<Measure> toIntegerMeasure(BatchReport.Measure batchMeasure, @Nullable String data) {
+    if (!batchMeasure.hasIntValue()) {
+      return toMeasure(MeasureImpl.createNoValue(), batchMeasure);
+    }
+    return toMeasure(MeasureImpl.create(batchMeasure.getIntValue(), data), batchMeasure);
+  }
+
+  private static Optional<Measure> toLongMeasure(BatchReport.Measure batchMeasure, @Nullable String data) {
+    if (!batchMeasure.hasLongValue()) {
+      return toMeasure(MeasureImpl.createNoValue(), batchMeasure);
+    }
+    return toMeasure(MeasureImpl.create(batchMeasure.getLongValue(), data), batchMeasure);
+  }
+
+  private static Optional<Measure> toDoubleMeasure(BatchReport.Measure batchMeasure, @Nullable String data) {
+    if (!batchMeasure.hasDoubleValue()) {
+      return toMeasure(MeasureImpl.createNoValue(), batchMeasure);
+    }
+    return toMeasure(MeasureImpl.create(batchMeasure.getDoubleValue(), data), batchMeasure);
+  }
+
+  private static Optional<Measure> toBooleanMeasure(BatchReport.Measure batchMeasure, @Nullable String data) {
+    if (!batchMeasure.hasBooleanValue()) {
+      return toMeasure(MeasureImpl.createNoValue(), batchMeasure);
+    }
+    return toMeasure(MeasureImpl.create(batchMeasure.getBooleanValue(), data), batchMeasure);
+  }
+
+  private static Optional<Measure> toStringMeasure(BatchReport.Measure batchMeasure) {
+    if (!batchMeasure.hasStringValue()) {
+      return toMeasure(MeasureImpl.createNoValue(), batchMeasure);
+    }
+    return toMeasure(MeasureImpl.create(batchMeasure.getStringValue()), batchMeasure);
+  }
+  
+  private static Optional<Measure> toLevelMeasure(BatchReport.Measure batchMeasure) {
+    if (!batchMeasure.hasStringValue()) {
+      return toMeasure(MeasureImpl.createNoValue(), batchMeasure);
+    }
+    Optional<Measure.Level> level = Measure.Level.toLevel(batchMeasure.getStringValue());
+    if (!level.isPresent()) {
+      return toMeasure(MeasureImpl.createNoValue(), batchMeasure);
+    }
+    return toMeasure(MeasureImpl.create(level.get()), batchMeasure);
+  }
+
+  private static Optional<Measure> toNoValueMeasure(BatchReport.Measure batchMeasure) {
+    return toMeasure(MeasureImpl.createNoValue(), batchMeasure);
+  }
+
+  private static Optional<Measure> toMeasure(MeasureImpl measure, BatchReport.Measure batchMeasure) {
+    if (batchMeasure.hasAlertStatus() && !measure.hasQualityGateStatus()) {
+      Optional<Measure.Level> qualityGateStatus = Measure.Level.toLevel(batchMeasure.getAlertStatus());
+      if (qualityGateStatus.isPresent()) {
+        String text = batchMeasure.hasAlertText() ? batchMeasure.getAlertText() : null;
+        measure.setQualityGateStatus(new Measure.QualityGateStatus(qualityGateStatus.get(), text));
+      }
+    }
+    return Optional.of((Measure) measure);
+  }
+
+}
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
@@ -1,0 +1,176 @@
+/*
+ * SonarQube, open source software quality management tool.
+ * Copyright (C) 2008-2014 SonarSource
+ * mailto:contact AT sonarsource DOT com
+ *
+ * SonarQube is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * SonarQube is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.computation.measure;
+
+import com.google.common.base.Objects;
+import com.google.common.base.Optional;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+import static java.util.Objects.requireNonNull;
+
+public interface Measure {
+
+  enum ValueType {
+    NO_VALUE, BOOLEAN, INT, LONG, DOUBLE, STRING, LEVEL
+  }
+
+  enum Level {
+    OK("Green"),
+    WARN("Orange"),
+    ERROR("Red");
+
+    private final String colorName;
+
+    Level(String colorName) {
+      this.colorName = colorName;
+    }
+
+    public String getColorName() {
+      return colorName;
+    }
+
+    public static Optional<Level> toLevel(@Nullable String level) {
+      if (level == null) {
+        return Optional.absent();
+      }
+      try {
+        return Optional.of(Level.valueOf(level));
+      } catch (IllegalArgumentException e) {
+        return Optional.absent();
+      }
+    }
+  }
+
+  /**
+   * A QualityGate status has a level and an optional describing text.
+   */
+  @Immutable
+  final class QualityGateStatus {
+    private final Level status;
+    @CheckForNull
+    private final String text;
+
+    /**
+     * Creates a QualityGateStatus without a text.
+     */
+    public QualityGateStatus(Level status) {
+      this(status, null);
+    }
+
+    /**
+     * Creates a QualityGateStatus with a optional text.
+     */
+    public QualityGateStatus(Level status, @Nullable String text) {
+      this.status = requireNonNull(status);
+      this.text = text;
+    }
+
+    public Level getStatus() {
+      return status;
+    }
+
+    @CheckForNull
+    public String getText() {
+      return text;
+    }
+
+    @Override
+    public String toString() {
+      return Objects.toStringHelper(this)
+          .add("status", status)
+          .add("text", text)
+          .toString();
+    }
+  }
+
+  /**
+   * The type of value stored in the measure.
+   */
+  ValueType getValueType();
+
+  /**
+   * The value of this measure as a boolean if the type is {@link Measure.ValueType#BOOLEAN}.
+   *
+   * @throws IllegalStateException if the value type of the measure is not {@link Measure.ValueType#BOOLEAN}
+   */
+  boolean getBooleanValue();
+
+  /**
+   * The value of this measure as a int if the type is {@link Measure.ValueType#INT}.
+   *
+   * @throws IllegalStateException if the value type of the measure is not {@link Measure.ValueType#INT}
+   */
+  int getIntValue();
+
+  /**
+   * The value of this measure as a long if the type is {@link Measure.ValueType#LONG}.
+   *
+   * @throws IllegalStateException if the value type of the measure is not {@link Measure.ValueType#LONG}
+   */
+  long getLongValue();
+
+  /**
+   * The value of this measure as a double if the type is {@link Measure.ValueType#DOUBLE}.
+   *
+   * @throws IllegalStateException if the value type of the measure is not {@link Measure.ValueType#DOUBLE}
+   */
+  double getDoubleValue();
+
+  /**
+   * The value of this measure as a String if the type is {@link Measure.ValueType#STRING}.
+   *
+   * @throws IllegalStateException if the value type of the measure is not {@link Measure.ValueType#STRING}
+   */
+  String getStringValue();
+
+  /**
+   * The value of this measure as a Level if the type is {@link Measure.ValueType#LEVEL}.
+   *
+   * @throws IllegalStateException if the value type of the measure is not {@link Measure.ValueType#LEVEL}
+   */
+  Level getLevelValue();
+
+  /**
+   * The data of this measure if it exists.
+   * <p>
+   * If the measure type is {@link Measure.ValueType#STRING}, the value returned by this function is the same as {@link #getStringValue()}.
+   * </p>
+   *
+   * @throws IllegalStateException if the value type of the measure is not {@link Measure.ValueType#BOOLEAN}
+   */
+  @CheckForNull
+  String getData();
+
+  /**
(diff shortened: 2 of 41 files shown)


server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  enum AlertStatus {
+    OK("Green"),
+    WARN("Orange"),
+    ERROR("Red");
+
+    private final String colorName;
+
+    AlertStatus(String colorName) {
```
simonbrandhof: The term "alert" should be removed. I suggest "QualityGateStatus" or "GateStatus"

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
```diff
@@ -0,0 +1,140 @@
...
+   * The type of value stored in the measure.
+   */
+  ValueType getValueType();
+
+  /**
+   * The value of this measure as a boolean if the type is {@link Measure.ValueType#BOOLEAN}.
+   *
+   * @throws IllegalStateException if the value type of the measure is not {@link Measure.ValueType#BOOLEAN}
+   */
+  boolean getBooleanValue();
```
simonbrandhof: What happens if the value is not set ? Returns ```false``` or throws an exception ?

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
```diff
@@ -0,0 +1,140 @@
...
+   */
+  String getStringValue();
+
+  /**
+   * The data of this measure if it exists.
+   * <p>
+   * If the measure type is {@link Measure.ValueType#STRING}, the value returned by this function is the same as {@link #getStringValue()}.
+   * </p>
+   *
+   * @throws IllegalStateException if the value type of the measure is not {@link Measure.ValueType#BOOLEAN}
```
simonbrandhof: Bad copy/paste of BOOLEAN.

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
```diff
@@ -0,0 +1,140 @@
...
+   */
+  @CheckForNull
+  String getData();
```
simonbrandhof: I wonder if we could merge ```getData()``` with ```getStringValue()```.

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * The alert text for this measure.
+   * <strong>Don't call this method unless you've checked the result of {@link #hasAlertText()} first</strong>
+   *
+   * @throws IllegalStateException if the measure has no alert text
+   */
+  String getAlertText();
+
+  Integer getSeverityIndex();
```
simonbrandhof: This is a db-specific value isn't it ? In this case it should not be here.

server/sonar-server/src/main/java/org/sonar/server/computation/metric/Metric.java
```diff
@@ -0,0 +1,57 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.computation.metric;
+
+import org.sonar.server.computation.measure.Measure;
+
+public interface Metric {
+  /**
+   * The Metric's key is its domain identifier.
```
simonbrandhof: What do you mean by "domain identifier" ?

server/sonar-server/src/main/java/org/sonar/server/computation/metric/Metric.java
```diff
@@ -0,0 +1,57 @@
...
+   */
+  String getKey();
+
+  String getName();
+
+  MetricType getMetricType();
```
simonbrandhof: Could be simply renamed ```getType()```.

server/sonar-server/src/main/java/org/sonar/server/computation/metric/Metric.java
```diff
@@ -0,0 +1,57 @@
...
+    MILLISEC(Measure.ValueType.INT),
+    RATING(Measure.ValueType.INT),
+    WORK_DUR(Measure.ValueType.LONG),
+    FLOAT(Measure.ValueType.DOUBLE),
+    PERCENT(Measure.ValueType.DOUBLE),
+    BOOL(Measure.ValueType.BOOLEAN),
+    STRING(Measure.ValueType.STRING),
+    DISTRIB(Measure.ValueType.STRING),
+    DATA(Measure.ValueType.STRING),
+    LEVEL(Measure.ValueType.NO_VALUE);
```
simonbrandhof: Why NO_VALUE ? This should be a STRING. Valid values are "OK", "WARN" and "ERROR".

server/sonar-server/src/main/java/org/sonar/server/computation/metric/MetricImpl.java
```diff
@@ -0,0 +1,82 @@
...
+  public int hashCode() {
+    return Objects.hash(key, metricType);
+  }
+
+  @Override
+  public String toString() {
+    return com.google.common.base.Objects.toStringHelper(this)
+        .add("key", key)
+        .add("name", name)
+        .add("metricType", metricType)
```
simonbrandhof: "type" ? :-)

server/sonar-server/src/main/java/org/sonar/server/computation/metric/MetricRepositoryImpl.java
```diff
@@ -0,0 +1,56 @@
...
+import org.sonar.core.metric.db.MetricDto;
+import org.sonar.core.persistence.DbSession;
+import org.sonar.server.db.DbClient;
+import org.sonar.server.metric.persistence.MetricDao;
+
+import static java.util.Objects.requireNonNull;
+
+public class MetricRepositoryImpl implements MetricRepository {
+  private final DbClient dbClient;
+  private final MetricDao metricDao;
```
simonbrandhof: This field can be removed. It is already available through DbClient.

server/sonar-server/src/main/java/org/sonar/server/computation/metric/MetricRepositoryImpl.java
```diff
@@ -0,0 +1,56 @@
...
+  public MetricRepositoryImpl(DbClient dbClient, MetricDao metricDao) {
+    this.dbClient = dbClient;
+    this.metricDao = metricDao;
+  }
+
+  @Override
+  public Metric getByKey(String key) {
+    requireNonNull(key);
+
+    try (DbSession dbSession = dbClient.openSession(false)) {
```
simonbrandhof: A SQL request is executed each time compute engine gets a metric. Metrics are a central piece. I suggest to simply load all of them once and to keep a cache during integration of report.

server/sonar-server/src/main/java/org/sonar/server/computation/metric/MetricImpl.java
```diff
@@ -0,0 +1,82 @@
...
+  public boolean equals(@Nullable Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    MetricImpl metric = (MetricImpl) o;
+    return Objects.equals(key, metric.key) &&
+        Objects.equals(metricType, metric.metricType);
```
simonbrandhof: The key is the key :-) The comparison of type must be dropped from equals() and hashCode().

server/sonar-server/src/main/java/org/sonar/server/computation/qualitygate/ResolvedCondition.java
```diff
@@ -0,0 +1,111 @@
...
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    ResolvedCondition that = (ResolvedCondition) o;
+    return java.util.Objects.equals(id, that.id)
+      && java.util.Objects.equals(metric, that.metric)
+      && java.util.Objects.equals(operator, that.operator)
+      && java.util.Objects.equals(warningThreshold, that.warningThreshold)
+      && java.util.Objects.equals(errorThreshold, that.errorThreshold)
+      && java.util.Objects.equals(period, that.period);
```
simonbrandhof: Condition id is metric + period.

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputationSteps.java
```diff
@@ -47,6 +47,8 @@
       // Read report
       ParseReportStep.class,
 
+      QualityGateStep.class,
```
simonbrandhof: Don't you think that ```Step``` implementations should be named with verbs like ```RetrieveQualityGateStep``` or nouns  like ```QualityGateLoaderStep``` ? Current name ```QualityGateStep" does not help to understand the responsibility.

server/sonar-server/src/main/java/org/sonar/server/computation/measure/MeasureRepository.java
```diff
@@ -19,17 +19,39 @@
...
+   *
+   * @throws NullPointerException if either argument is {@code null}
+   */
+  Optional<Measure> getBaseMeasure(Component component, Metric metric);
+
+  /**
+   * Retrieves the measure created during the current analysis for the specified {@link Component} for the specified
+   * {@link MetricImpl} if it exists (ie. one created by the Compute Engine or the Batch).
+   */
+  Optional<Measure> getRawMeasure(Component component, Metric metric);
```
simonbrandhof: The term "raw" is implicit so it could be dropped (```getMeasure()``` and ```getBaseMeasure()```). WDYT ?

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
sns-seb: if the Measure has no value, its ValueType will be NO_VALUE and as per the description of the method it will indeed throw an exception

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
sns-seb: I'm going for `QualityGateStatus`

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
sns-seb: good catch
there is no throw at all on this method (hence the `@CheckForNull`)

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
sns-seb: nop, some measure have another type and data still
seen in batch's code

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
sns-seb: left over, removing it, it's not even implemented

server/sonar-server/src/main/java/org/sonar/server/computation/measure/MeasureRepository.java
sns-seb: given the documentation of this method, yes, raw could be dropped

but when you suggested to use "base" and "raw", this paradigms seemed to be used in other places

we should just be consistent

server/sonar-server/src/main/java/org/sonar/server/computation/metric/Metric.java
sns-seb: see design of Measure (the legacy and the new one)
"level" is not stored as a the value of the Measure but in a specific field (which all measures, whichever their type may have populated)
hence, for consistency, the Measure has no value but an always populated QualityGateStatus

Please note that the LEVEL is badly named or at least misleading (its really a metric of type QualityGate), but this is the value stored in DB (and also exposed by WebService I reckon)

server/sonar-server/src/main/java/org/sonar/server/computation/metric/MetricImpl.java
sns-seb: yes, can be renamed, shorter is better here

server/sonar-server/src/main/java/org/sonar/server/computation/metric/MetricImpl.java
sns-seb: ok

server/sonar-server/src/main/java/org/sonar/server/computation/metric/MetricRepositoryImpl.java
sns-seb: I haven't made any assumption on how this should be optimized yet since I don't know how it is used (only use I have yet can really work with this simple implementation)

server/sonar-server/src/main/java/org/sonar/server/computation/qualitygate/ResolvedCondition.java
```diff
@@ -0,0 +1,111 @@
...
+
+  public Metric getMetric() {
+    return metric;
+  }
+
+  public String getOperator() {
+    return operator;
+  }
+
+  public String getWarningThreshold() {
```
sns-seb: I realize annotations are missing on the getters of properties annotated with `@CheckForNull`, IDEA did not generate them

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputationSteps.java
sns-seb: I agree with your comment

WDYT of `QualityGateLoadingStep`?

server/sonar-server/src/main/java/org/sonar/server/computation/qualitygate/ResolvedCondition.java
sns-seb: could I drop the "id" field?
It's DB specific and I don't think we do anything with it in batch code

server/sonar-server/src/main/java/org/sonar/server/computation/metric/Metric.java
sns-seb: I wanted to describe this getter and emphasis this is not the DB identifier. Also, I added this because just "The metric's key" was a useless javadoc

server/sonar-server/src/main/java/org/sonar/server/computation/step/QualityGateStep.java
```diff
@@ -0,0 +1,100 @@
...
+    String qualityGateSetting = settings.getString(PROPERTY_QUALITY_GATE);
+
+    if (qualityGateSetting == null || StringUtils.isBlank(qualityGateSetting)) {
+      LOGGER.debug("No quality gate is configured for project " + projectKey);
+      qualityGateHolder.setQualityGate(DEFAULT_QUALITY_GATE);
+      return;
+    }
+
+    try {
+      long qualityGateId = Long.valueOf(qualityGateSetting.trim());
```
SonarTech: MAJOR Use "Long.parseLong" for this string-to-long conversion. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS2130)

server/sonar-server/src/main/java/org/sonar/server/computation/measure/BatchMeasureToMeasure.java
```diff
@@ -0,0 +1,114 @@
...
+
+import com.google.common.base.Optional;
+import java.util.Objects;
+import javax.annotation.Nullable;
+import org.sonar.batch.protocol.output.BatchReport;
+import org.sonar.server.computation.metric.Metric;
+
+public class BatchMeasureToMeasure {
+
+  public Optional<Measure> toMeasure(@Nullable BatchReport.Measure batchMeasure, Metric metric) {
```
SonarTech: MAJOR The Cyclomatic Complexity of this method "toMeasure" is 17 which is greater than 10 authorized. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)
MAJOR Reduce the number of returns of this method 7, down to the maximum allowed 5. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1142)

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
```diff
@@ -0,0 +1,140 @@
...
+      return colorName;
+    }
+
+    public static Optional<Measure.AlertStatus> toAlertStatus(@Nullable String alertStatus) {
+      if (alertStatus == null) {
+        return Optional.absent();
+      }
+      try {
+        return Optional.of(Measure.AlertStatus.valueOf(alertStatus));
+      } catch (IllegalArgumentException e) {
```
SonarTech: CRITICAL Either log or rethrow this exception. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1166)

server/sonar-server/src/main/java/org/sonar/server/computation/measure/MeasureDtoToMeasure.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.Objects;
+import javax.annotation.Nullable;
+import org.sonar.core.measure.db.MeasureDto;
+import org.sonar.server.computation.metric.Metric;
+
+import static org.sonar.server.computation.measure.Measure.AlertStatus.toAlertStatus;
+
+public class MeasureDtoToMeasure {
+
+  public Optional<Measure> toMeasure(@Nullable MeasureDto measureDto, Metric metric) {
```
SonarTech: MAJOR Reduce the number of returns of this method 7, down to the maximum allowed 5. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1142)
MAJOR The Cyclomatic Complexity of this method "toMeasure" is 16 which is greater than 10 authorized. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

server/sonar-server/src/main/java/org/sonar/server/computation/measure/MeasureDtoToMeasure.java
```diff
@@ -0,0 +1,111 @@
...
+      return toMeasure(MeasureImpl.createNoValue(), measureDto);
+    }
+    return toMeasure(MeasureImpl.create(value.doubleValue(), data), measureDto);
+  }
+
+  private static Optional<Measure> toBooleanMeasure(MeasureDto measureDto, @Nullable Double value, String data) {
+    if (value == null) {
+      return toMeasure(MeasureImpl.createNoValue(), measureDto);
+    }
+    return toMeasure(MeasureImpl.create(value == 1.0d, data), measureDto);
```
SonarTech: CRITICAL Equality tests should not be made with floating point values. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1244)

server/sonar-server/src/main/java/org/sonar/server/computation/qualitygate/ResolvedCondition.java
```diff
@@ -0,0 +1,111 @@
...
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    ResolvedCondition that = (ResolvedCondition) o;
+    return java.util.Objects.equals(id, that.id)
```
SonarTech: MAJOR Reduce the number of conditional operators (5) used in the expression (maximum allowed 3). [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1067)

server/sonar-server/src/main/java/org/sonar/server/computation/qualitygate/ResolvedCondition.java
```diff
@@ -0,0 +1,111 @@
...
+  public String getErrorThreshold() {
+    return errorThreshold;
+  }
+
+  public Integer getPeriod() {
+    return period;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
SonarTech: MAJOR The Cyclomatic Complexity of this method "equals" is 11 which is greater than 10 authorized. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AMethodCyclomaticComplexity)

server/sonar-server/src/main/java/org/sonar/server/computation/measure/MeasureImpl.java
```diff
@@ -0,0 +1,158 @@
...
+
+  @Override
+  public ValueType getValueType() {
+    return valueType;
+  }
+
+  @Override
+  public boolean getBooleanValue() {
+    checkValueType(ValueType.BOOLEAN);
+    return value == 1.0d;
```
SonarTech: CRITICAL Equality tests should not be made with floating point values. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1244)

server/sonar-server/src/main/java/org/sonar/server/computation/measure/MeasureImpl.java
```diff
@@ -0,0 +1,158 @@
...
+  @Nullable
+  private String alertText;
+
+  protected MeasureImpl(ValueType valueType, @Nullable Double value, @Nullable String data) {
+    this.valueType = valueType;
+    this.value = value;
+    this.data = data;
+  }
+
+  public static MeasureImpl create(boolean value, @Nullable String data) {
```
SonarTech: MAJOR Provide multiple methods instead of using "value" to determine which action to take. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS2301)

SonarTech:
SonarQube analysis reported 11 new issues:
* 3 critical
* 8 major

server/sonar-server/src/main/java/org/sonar/server/computation/metric/Metric.java
simonbrandhof: No, level is a metric type. Related measures have a value. You mix with alert_status.

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputationSteps.java
simonbrandhof: ok

server/sonar-server/src/main/java/org/sonar/server/computation/qualitygate/ResolvedCondition.java
simonbrandhof: exact. Id must be dropped.

server/sonar-server/src/main/java/org/sonar/server/computation/metric/Metric.java
sns-seb: it will be

server/sonar-server/src/main/java/org/sonar/server/computation/metric/Metric.java
sns-seb: the enum for Level is the same as the enum for the alertStatus, it is really confusing

and the only usage in CoreMetrics of the MetricType LEVEL is the ALERT_STATUS which does not use the data field of the legacy metric but the alertStatus

I'm missing something, lets talk about it

server/sonar-server/src/main/java/org/sonar/server/computation/qualitygate/ResolvedCondition.java
sns-seb: fixed

sns-seb:
PR updated
* rebased on master
   * handled conflicts on several steps
* Measure now have a Level ValueType
* Measure now have an object to represent qualitygate status and text (renamed from alertStatus and alertText)

simonbrandhof:
I'm not sure that supporting measures without any value is needed (```ValueType.NO_VALUE```). Do you have an example of usage ?

server/sonar-server/src/main/java/org/sonar/server/computation/measure/MeasureRepositoryImpl.java
```diff
@@ -19,39 +19,81 @@
...
+  private final Map<Integer, Map<String, Measure>> measures = new HashMap<>();
 
-  public MeasureRepositoryImpl(DbClient dbClient, BatchReportReader reportReader) {
+  public MeasureRepositoryImpl(DbClient dbClient, BatchReportReader reportReader,
+    MeasureDtoToMeasure measureDtoToMeasure, final BatchMeasureToMeasure batchMeasureToMeasure,
+    final MetricRepository metricRepository) {
     this.dbClient = dbClient;
     this.reportReader = reportReader;
+    this.measureDtoToMeasure = measureDtoToMeasure;
+    this.batchMeasureToMeasure = batchMeasureToMeasure;
```
simonbrandhof: What's the value to inject this object instead of using a static method like ```BatchMeasureToMeasure.toMeasure()``` ?

simonbrandhof:
2 comments and LGTM

sns-seb:
>I'm not sure that supporting measures without any value is needed (ValueType.NO_VALUE). Do you have an example of usage ?

This handles the case where, for any reason, db data can not be read into a Measure object. The other solution is to raise an exception, not sure we want to fail the whole process for one bad value in DB...

server/sonar-server/src/main/java/org/sonar/server/computation/measure/MeasureRepositoryImpl.java
sns-seb: >What's the value for injecting this object instead of using a static method like BatchMeasureToMeasure.toMeasure() ?

none, removing the constructor parameters

> Note that QualityGateServiceImpl is slightly different: the field conditionDtoToBean is instantiated directly in the constructor and is not injected by container.

`conditionDtoToBean` is a `Function`, it can not be injected. But it depends on the injected `MetricRepository` (so, must be stored in a property to avoid useless multiple instanciations). Since the `MetricRepository` is used only in this `Function` we do not create a useless `MetricRepository` property and the `Function` is created in the constructor

PR #361, opened 2015-06-09
SONAR-6567 send notifications on QGate failures

simonbrandhof:
Note: deprecated DumbComponent constructor is still there and should be dropped.


sns-seb:
commit message is too long

sns-seb:
too bad the renaming of `PopulateComponentsUuidAndKeyStep` isn't isolated into a specific commit. Hard to tell there was any change to code

sns-seb:
since `ComponentImpl.name` is now set in `ComponentImpl`'s constructor, is there any other changfe in  `FillComponentsStep` than renaming the class now?
asking because I wonder why Git does not seem to identify the file renaming (and see an addition and a separate deletion)

sns-seb:
please improve `DumbComponent`'s builder and then LGTM

PR #362, opened 2015-06-10
SONAR-6615 add in table 'manual_measures' column 'compunent_uuid'

teryk:
tested on dory (migration + ui with manual_measures creation and update)


simonbrandhof:
LGTM

PR #364, opened 2015-06-10
SONAR-6557 Clean design stuff in database

julienlancelot:


simonbrandhof:
Some SQL requests in ```DefaultTimeMachine``` and ```PastSnapshotFinder*``` have conditions on ```qualifier<>'LIB'```. These classes aim to be moved to compute engine, but the conditions can already be dropped.
Else LGTM.

julienlancelot:
You're right, I saw then but I didn't want to touch it as it will be moved to the CE, but I can indeed already drop it.

julienlancelot:
Feedback apply

PR #368, opened 2015-06-11
Start refactoring on issue JSON serialization

jblievremont:
Initial goal is to use the same objects to write Issue and User objects across Web Services.


simonbrandhof:
LGTM

PR #369, opened 2015-06-11
rename field reportReader in SourceLinesCache

teryk:


simonbrandhof:
no need to do a PR for that !

teryk:
it's assigned to me. Just wanted to check if it generates issues

PR #372, opened 2015-06-11
Create a DebtModelHolder

julienlancelot:


server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
```diff
@@ -0,0 +1,52 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.server.computation.debt;
+
+import com.google.common.base.Preconditions;
+import org.apache.commons.lang.builder.ToStringBuilder;
+import org.apache.commons.lang.builder.ToStringStyle;
+
+public class CharacteristicImpl implements Characteristic {
```
sns-seb: You could add the `@Immutable` annotation to this class, it's only informative (unless we have rules triggered by the presence of this annotation) but a good practice imho

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
```diff
@@ -0,0 +1,52 @@
...
+
+  private int id;
```
sns-seb: missing final on properties

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
```diff
@@ -0,0 +1,52 @@
...
+  private String key;
+
+  public CharacteristicImpl(int id, String key) {
```
sns-seb: I guess its missing `@Nullable` on key parameter since there is no null check on it

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    Preconditions.checkNotNull(key, "key cannot be null");
```
sns-seb: you should use `Objects.requireNonNull` since this method is the same as Guava's but provided by the JDK

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @Override
+  public String key() {
+    return key;
+  }
+
+  @Override
+  public String toString() {
+    return ToStringBuilder.reflectionToString(this, ToStringStyle.SHORT_PREFIX_STYLE);
```
sns-seb: for dumb bean like that, don't use reflection, that's stupid waste of CPU
there is convenient alternative (check the options provided by IDEA's toString generator, `Objects.toStringHelper` from Guava is nice enough but it will never beat a simple StringBuffer performance wise)

as a general practice, reflexion should never be used unless nothing can really be done otherwise
but in toString() more than anywhere else
one can not know how many times an Object's toString method may be called, directly or as part of wrapping objects
in addition, toString is one of these tricky cases where enabling log can actually make things worse in production

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    this.key = key;
+  }
+
+  @Override
+  public int id() {
+    return id;
+  }
+
+  @Override
+  public String key() {
```
sns-seb: missing `@CheckForNull`

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolder.java
```diff
@@ -0,0 +1,51 @@
...
+
+public interface DebtModelHolder {
+
+  /**
+   * Return a characteristic by its key
+   *
+   * @throws IllegalStateException if the holder is empty
+   */
+  @CheckForNull
+  Characteristic characteristicByKey(String key);
```
sns-seb: why not using a get or find? (I prefer the former) functions need verbs to be clear about their intent

you could use `Optional<Characteristic>` to enforce the fact that the method may not find the Characteristic and avoid returning null and the `@CheckForNull` annotation

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolder.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Return the list of root characteristics
+   *
+   * @throws IllegalStateException if the holder is empty
+   */
+  Collection<Characteristic> rootCharacteristics();
```
sns-seb: same thing here, method name lacks a verb

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolderImpl.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.common.collect.Multimap;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DebtModelHolderImpl implements MutableDebtModelHolder {
+
+  private Multimap<Characteristic, Characteristic> subCharacteristicsByRootCharacteristic = ArrayListMultimap.create();
```
sns-seb: missing `final`

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolderImpl.java
```diff
@@ -0,0 +1,77 @@
...
+  private Map<String, Characteristic> characteristicByKey = new HashMap<>();
```
sns-seb: missing `final`

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
simonbrandhof: I completed http://confluence.internal.sonarsource.com/display/PLAT/Custom+Dory+Rules for adding a rule on Dory

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolderImpl.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Override
+  public void addCharacteristics(Characteristic rootCharacteristic, List<? extends Characteristic> subCharacteristics) {
+    Preconditions.checkNotNull(rootCharacteristic, "rootCharacteristic cannot be null");
+    Preconditions.checkNotNull(subCharacteristics, "subCharacteristics cannot be null");
+    Preconditions.checkState(!subCharacteristics.isEmpty(), "subCharacteristics cannot be empty");
```
sns-seb: this is not an `IllegalStateException`, this is an `IllegalArgumentException`

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolderImpl.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Map;
+
+public class DebtModelHolderImpl implements MutableDebtModelHolder {
+
+  private Multimap<Characteristic, Characteristic> subCharacteristicsByRootCharacteristic = ArrayListMultimap.create();
+  private Map<String, Characteristic> characteristicByKey = new HashMap<>();
+
+  @Override
+  public void addCharacteristics(Characteristic rootCharacteristic, List<? extends Characteristic> subCharacteristics) {
+    Preconditions.checkNotNull(rootCharacteristic, "rootCharacteristic cannot be null");
```
sns-seb: use `Objects.requireNonNull()` and maybe a static import for clarity

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolderImpl.java
```diff
@@ -0,0 +1,77 @@
...
+    return ImmutableList.copyOf(subCharacteristicsByRootCharacteristic.get(rootCharacteristic));
+  }
+
+  @Override
+  public Characteristic characteristicByKey(String key) {
+    checkCharacteristicsAreInitialized();
+    return characteristicByKey.get(key);
+  }
+
+  private void checkCharacteristicsAreInitialized() {
```
sns-seb: missing `static`

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
sns-seb: this bean lacks a `equals` and `hashcode` implementation

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolderImpl.java
```diff
@@ -0,0 +1,77 @@
...
+    subCharacteristicsByRootCharacteristic.putAll(rootCharacteristic, subCharacteristics);
+
+    characteristicByKey.put(rootCharacteristic.key(), rootCharacteristic);
+    for (Characteristic characteristic : subCharacteristics) {
+      characteristicByKey.put(characteristic.key(), characteristic);
+    }
+  }
+
+  @Override
+  public Collection<Characteristic> rootCharacteristics() {
```
sns-seb: why not returning a `Set`?
it enforces the fact that there is no duplicate in the returned collection

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolderImpl.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  @Override
+  public Collection<Characteristic> subCharacteristicsByRootKey(String rootCharacteristicKey) {
+    checkCharacteristicsAreInitialized();
+    Characteristic rootCharacteristic = characteristicByKey.get(rootCharacteristicKey);
+    if (rootCharacteristic == null) {
+      return Collections.emptyList();
+    }
+    return ImmutableList.copyOf(subCharacteristicsByRootCharacteristic.get(rootCharacteristic));
```
sns-seb: is the order of `Characteristic` objects in the returned collection relevant?
If so, why not return a `List`?

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolder.java
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Return a characteristic by its key
+   *
+   * @throws IllegalStateException if the holder is empty
+   */
+  @CheckForNull
+  Characteristic characteristicByKey(String key);
+
+  /**
+   * Return the list of root characteristics
```
sns-seb: if its a "List" why return a `Collection`?

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedDebtModelStep.java
```diff
@@ -0,0 +1,112 @@
...
+      public Characteristic apply(@Nonnull CharacteristicDto characteristicDto) {
+        return toCharacteristic(characteristicDto);
+      }
+    };
+    this.dtoToParentIdFunction = new Function<CharacteristicDto, Integer>() {
+      @Nullable
+      @Override
+      public Integer apply(@Nonnull CharacteristicDto characteristicDto) {
+        Integer parentId = characteristicDto.getParentId();
+        return parentId != null ? parentId : null;
```
sns-seb: invert test for clarity

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedDebtModelStep.java
```diff
@@ -0,0 +1,112 @@
...
+    this.dbClient = dbClient;
+    this.mutableDebtModelHolder = mutableDebtModelHolder;
+    this.dtoToCharFunction = new Function<CharacteristicDto, Characteristic>() {
+      @Nullable
+      @Override
+      public Characteristic apply(@Nonnull CharacteristicDto characteristicDto) {
+        return toCharacteristic(characteristicDto);
+      }
+    };
+    this.dtoToParentIdFunction = new Function<CharacteristicDto, Integer>() {
```
sns-seb: this function does not reference any property of the class, it should be a singleton enum

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedDebtModelStep.java
```diff
@@ -0,0 +1,112 @@
...
+      @Nullable
+      @Override
+      public Integer apply(@Nonnull CharacteristicDto characteristicDto) {
+        Integer parentId = characteristicDto.getParentId();
+        return parentId != null ? parentId : null;
+      }
+    };
+    this.isRootPredicate = new Predicate<CharacteristicDto>() {
```
sns-seb: this function does not reference any property of the class, it should be a singleton enum

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedDebtModelStep.java
```diff
@@ -0,0 +1,112 @@
...
+  private final DbClient dbClient;
+  private final MutableDebtModelHolder mutableDebtModelHolder;
+  private final Function<CharacteristicDto, Characteristic> dtoToCharFunction;
+  private final Function<CharacteristicDto, Integer> dtoToParentIdFunction;
+  private final Predicate<CharacteristicDto> isRootPredicate;
+
+  public FeedDebtModelStep(DbClient dbClient, MutableDebtModelHolder mutableDebtModelHolder) {
+    this.dbClient = dbClient;
+    this.mutableDebtModelHolder = mutableDebtModelHolder;
+    this.dtoToCharFunction = new Function<CharacteristicDto, Characteristic>() {
```
sns-seb: this function does not reference any property of the class, it should be a singleton enum

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedDebtModelStep.java
```diff
@@ -0,0 +1,112 @@
...
+      .filter(Predicates.not(isRootPredicate))
+      .index(dtoToParentIdFunction);
+    for (CharacteristicDto rootCharacteristicDto : rootCharacteristicDtos) {
+      Collection<CharacteristicDto> subCharacteristicDtos = subCharacteristicDtosByParentId.get(rootCharacteristicDto.getId());
+
+      Characteristic rootCharacteristic = toCharacteristic(rootCharacteristicDto);
+      List<Characteristic> subCharacteristic = FluentIterable.from(subCharacteristicDtos)
+        .transform(dtoToCharFunction)
+        .toList();
+      mutableDebtModelHolder.addCharacteristics(rootCharacteristic, subCharacteristic);
```
sns-seb: modify `addCharacteristics` to take an `Iterable` instead of a List (unless the order of the elements is actually relevant) and do not use the `toList` of `FluentIterable`. You will save the useless creation of an intermediary list passed to the holder and just read through

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedDebtModelStep.java
```diff
@@ -0,0 +1,112 @@
...
+    try {
+      feedDebtModel(session);
+    } finally {
+      session.close();
+    }
+  }
+
+  private void feedDebtModel(DbSession session) {
+    List<CharacteristicDto> characteristicDtos = dbClient.debtCharacteristicDao().selectEnabledCharacteristics(session);
+    List<CharacteristicDto> rootCharacteristicDtos = FluentIterable.from(characteristicDtos)
```
sns-seb: you can save the creation of a `List` object by looping directly on the result of the `filter` method

server/sonar-server/src/test/java/org/sonar/server/computation/debt/CharacteristicImplTest.java
```diff
@@ -0,0 +1,53 @@
...
+  @Test
+  public void test_getter_and_setter() throws Exception {
+    CharacteristicImpl characteristic = new CharacteristicImpl(1, "PORTABILITY");
+    assertThat(characteristic.id()).isEqualTo(1);
+    assertThat(characteristic.key()).isEqualTo("PORTABILITY");
+  }
+
+  @Test
+  public void test_to_string() throws Exception {
+    assertThat(new CharacteristicImpl(1, "PORTABILITY").toString()).isNotEmpty();
```
sns-seb: you should test the result of toString, you could be invoking the `Object.toString()` method as far as this test checks

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedDebtModelStep.java
```diff
@@ -0,0 +1,112 @@
...
+      .filter(isRootPredicate).toList();
+    Multimap<Integer, CharacteristicDto> subCharacteristicDtosByParentId = FluentIterable.from(characteristicDtos)
+      .filter(Predicates.not(isRootPredicate))
```
sns-seb: the result of `Predicates.not` could also be stored in a constant, it actually creates a new Object which can be spared

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedDebtModelStep.java
```diff
@@ -0,0 +1,112 @@
...
+  public void execute() {
+    DbSession session = dbClient.openSession(false);
+    try {
+      feedDebtModel(session);
+    } finally {
+      session.close();
+    }
+  }
+
+  private void feedDebtModel(DbSession session) {
```
sns-seb: from what I see, his method can be written differently and optimized

let's talk about that

server/sonar-server/src/test/java/org/sonar/server/computation/debt/DebtModelHolderImplTest.java
```diff
@@ -0,0 +1,122 @@
...
+    assertThat(sut.rootCharacteristics()).hasSize(2);
+    assertThat(sut.subCharacteristicsByRootKey("PORTABILITY")).hasSize(2);
+  }
+
+  @Test
+  public void add_characteristics_fail_with_a_NPE_if_root_characteristic_is_null() throws Exception {
+    thrown.expect(NullPointerException.class);
+    thrown.expectMessage("rootCharacteristic cannot be null");
+
+    sut.addCharacteristics(null, Collections.singletonList(new CharacteristicImpl(2, "COMPILER_RELATED_PORTABILITY")));
```
sns-seb: what about a static import of `singletonList` to improve readability?

server/sonar-server/src/test/java/org/sonar/server/computation/debt/DebtModelHolderImplTest.java
```diff
@@ -0,0 +1,122 @@
...
+  public void get_characteristic_by_key() throws Exception {
+    sut.addCharacteristics(new CharacteristicImpl(1, "PORTABILITY"), Collections.singletonList(new CharacteristicImpl(2, "COMPILER_RELATED_PORTABILITY")));
+
+    assertThat(sut.characteristicByKey("PORTABILITY")).isNotNull();
+    assertThat(sut.characteristicByKey("COMPILER_RELATED_PORTABILITY")).isNotNull();
+    assertThat(sut.characteristicByKey("UNKNOWN")).isNull();
+  }
+
+  @Test
+  public void get_characteristic_by_key_throws_a_ISE_when_not_initialized() throws Exception {
```
sns-seb: lets be very picky about English: either write "_an_ISE_" or just "_ISE_" for short

server/sonar-server/src/test/java/org/sonar/server/computation/debt/DebtModelHolderImplTest.java
```diff
@@ -0,0 +1,122 @@
...
+    thrown.expect(IllegalStateException.class);
+    thrown.expectMessage("Characteristics have not been initialized yet");
+
+    sut.rootCharacteristics();
+  }
+
+  @Test
+  public void get_sub_characteristics_by_root_key() throws Exception {
+    sut.addCharacteristics(new CharacteristicImpl(1, "PORTABILITY"),
+      Arrays.asList(new CharacteristicImpl(2, "COMPILER_RELATED_PORTABILITY"), new CharacteristicImpl(4, "READABILITY")));
```
sns-seb: you could use constants for `new CharacteristicImpl(1, "PORTABILITY")`, `new CharacteristicImpl(2, "COMPILER_RELATED_PORTABILITY")`, etc...

server/sonar-server/src/test/java/org/sonar/server/computation/step/FeedDebtModelStepTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.sonar.server.computation.debt.Characteristic;
+import org.sonar.server.computation.debt.DebtModelHolderImpl;
+import org.sonar.server.computation.debt.MutableDebtModelHolder;
+import org.sonar.server.db.DbClient;
+import org.sonar.test.DbTests;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Category(DbTests.class)
+public class FeedDebtModelStepTest extends BaseStepTest {
```
sns-seb: the value of `BaseStepTest` is very limited and the `step()` method little meaningful... What about dropping this abstract class?

server/sonar-server/src/main/java/org/sonar/server/computation/debt/Characteristic.java
```diff
@@ -0,0 +1,29 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.server.computation.debt;
+
+public interface Characteristic {
```
sns-seb: I wonder what is the value of having an interface here. This is a dumb simple bean, defines not behavior, we might be better off with a final class

sns-seb:
I've flooded this PR with comments :) @julienlancelot its now yours

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
julienlancelot: Ok, I was not aware of the @Immutable interface, Im adding it

julienlancelot: Why would we always add equals and hashcode if it's not used ?

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
julienlancelot: Good catch !

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
julienlancelot: Ok

server/sonar-server/src/main/java/org/sonar/server/computation/debt/CharacteristicImpl.java
julienlancelot: Ok !

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolder.java
julienlancelot: Indeed, it's more explicit to add a prefix. 
And I could also use an Optional, I'm still not used to use it.

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolderImpl.java
julienlancelot: No, the order is not relevant

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolderImpl.java
julienlancelot: No, because characteristicByKey is a class variable

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedDebtModelStep.java
julienlancelot: Ok

server/sonar-server/src/test/java/org/sonar/server/computation/step/FeedDebtModelStepTest.java
julienlancelot: Yes, it's a good idea !

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedDebtModelStep.java
julienlancelot: Ok

server/sonar-server/src/main/java/org/sonar/server/computation/debt/Characteristic.java
julienlancelot: I was wondering the same thing in fact... So let's remove the interface !

julienlancelot:
Thanks for your feedbacks, they're applied @sns-seb !

sns-seb:
> Why would we always add equals and hashcode if it's not used ?

since I suggested to put CharacteristicImpl objects into a Set, equals and hashcode are needed
more over, you may want to control how objects will behave when added to a Set or as a key in a Map when you design the object

server/sonar-server/src/main/java/org/sonar/server/computation/debt/DebtModelHolderImpl.java
sns-seb: but, is the fact that a `Characteristic` can appear only once in the returned collection relevant?
If so, the returned type should be a `Set`

julienlancelot:
Feedback applied !

sns-seb:
well, LGTM then!

PR #376, opened 2015-06-15
add update method to MeasureRepository

sns-seb:
this PR:
* makes `Measure` class immutable
* adds full support for `Measure` associated to a rule or a characteristic
* adds the `update` method
* fix some quality flaws encountered on the way

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
@@ -76,13 +76,13 @@ public final class Measure {
   @CheckForNull
   private final String description;
   @CheckForNull
-  private QualityGateStatus qualityGateStatus;
+  private final QualityGateStatus qualityGateStatus;
   @CheckForNull
-  private MeasureVariations variations;
+  private final MeasureVariations variations;
 
   private Measure(ValueType valueType, @Nullable Integer ruleId, @Nullable Integer characteristicId,
     @Nullable Double value, @Nullable String data, @Nullable Level dataLevel,
-    @Nullable String description) {
+    @Nullable String description, @Nullable QualityGateStatus qualityGateStatus, @Nullable MeasureVariations variations) {
     this.valueType = valueType;
     this.ruleId = ruleId;
     this.characteristicId = characteristicId;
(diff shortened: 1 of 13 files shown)


server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
```diff
@@ -137,37 +146,93 @@ public Builder forCharacteristic(int characteristicId) {
...
+    }
+
+    /**
+     * Sets the QualityGateStatus of the updated Measure to create.
+     *
+     * @throws NullPointerException if the specified {@link QualityGateStatus} is {@code null}
+     * @throws UnsupportedOperationException if the source measure already has a {@link QualityGateStatus}
+     */
+    public UpdateMeasureBuilder setQualityGateStatus(QualityGateStatus qualityGateStatus) {
+      if (source.qualityGateStatus != null) {
```
julienlancelot: Why not use a Preconditions.checkState ?

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
```diff
@@ -137,37 +146,93 @@ public Builder forCharacteristic(int characteristicId) {
...
+    }
+
+    /**
+     * Sets the MeasureVariations of the updated Measure to create.
+     *
+     * @throws NullPointerException if the specified {@link MeasureVariations} is {@code null}
+     * @throws UnsupportedOperationException if the source measure already has a {@link MeasureVariations}
+     */
+    public UpdateMeasureBuilder setVariations(MeasureVariations variations) {
+      if (source.variations != null) {
```
julienlancelot: Why not use Preconditions.checkState ?

server/sonar-server/src/main/java/org/sonar/server/computation/measure/MeasureRepositoryImpl.java
```diff
@@ -173,4 +237,55 @@ public String apply(@Nonnull BatchReport.Measure input) {
       return input.getMetricKey();
     }
   }
+
+  @Immutable
+  private static final class MeasureKey {
+    private static final int DEFAULT_INT_VALUE = -6253;
```
julienlancelot: What is that ? 
Why not use a Object for ruleId and characteristicId, and set default value to null ?

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
sns-seb: Because I think an `UnsupportedOperationException` is a more appropriate exception here. The state of the builder isn't incorrect since we refuse the change that would lead to it, the operation the consumer is trying to attemp is.

julienlancelot: The state could be considered as incorrect as the caller is using a builder in a wrong state as the quality gate is already be defined, but it's a matter of point of view.

server/sonar-server/src/main/java/org/sonar/server/computation/measure/MeasureRepositoryImpl.java
sns-seb: two reasons:
* `int` uses less memory than `Integer`
* saves on handling the case where property is null (slighly faster)

minor optimization for an object which is used as a map key and which equals and hashcode methods are potentially highly used

julienlancelot: Ok

julienlancelot:
LGTM

PR #378, opened 2015-06-17
now fully cache measures from batch in MeasureRepository

sns-seb:
side effect: Measure returned by getRawMeasures can now be updated


julienlancelot:
Tested, LGTM !

PR #379, opened 2015-06-17
SONAR-6643 Move computation of differential measures to Compute Engine

julienlancelot:


server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
```diff
@@ -28,7 +28,7 @@
 import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
 
-public final class Measure {
+public class Measure {
```
sns-seb: final should not be removed on this class. Its contracts is complexe enough that to be guaranteed, the possibility to subclass it must be disabled

server/sonar-server/src/main/java/org/sonar/server/computation/metric/MetricImpl.java
```diff
@@ -60,6 +61,13 @@ public MetricType getType() {
     return type;
   }
 
+  public boolean isNumericType() {
```
sns-seb: This function returns `true` when valuetype is `BOOLEAN` which is not understandable to anyone. This function is more `isStoredAsADouble` or `isConvertibleToDouble`.

But Storage is : 
1. no concern to this Metric interface as a element of the Compute Engine. Abstracting away the storing concerns is the point of having interfaces and repositories in the CE
2. is related to the Measure and not to the Metric

In addition, this method is used in a single step. This is internal mechanic of this step. It should not be moved to the Metric interface (nor the Measure class).

Finally, it misses the `@Override` annotation

server/sonar-server/src/main/java/org/sonar/server/computation/metric/MetricRepositoryImpl.java
```diff
@@ -83,6 +84,11 @@ public Metric getById(long id) {
     return res;
   }
 
+  @Override
+  public Iterable<Metric> getAll() {
+    return FluentIterable.from(metricsByKey.values());
```
sns-seb: Just using `FluentIterable.from` like that is either not necessary or not what you intended.

Either returns `metricsByKey.values()` directly because it is a `Set` which implements `Iterable` or use `toSet` or `ToList` of `FluentIterable` to create a copy of the `Set` and avoid having the internal map potentially modified externally

(go for the second option!)

server/sonar-server/src/main/java/org/sonar/server/computation/period/Period.java
```diff
@@ -22,19 +22,22 @@
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
-import org.apache.commons.lang.builder.ReflectionToStringBuilder;
+import javax.annotation.concurrent.Immutable;
 
+@Immutable
 public class Period {
```
sns-seb: I've got a cleaner version of this class in PR https://github.com/SonarSource/sonarqube/pull/384, you should rebase on it to add this snapshotId property

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedPeriodsStep.java
```diff
@@ -165,7 +165,7 @@ private Period findByDate(int index, Date date) {
         return null;
       }
       LOG.debug(String.format("Compare to date %s (analysis of %s)", formatDate(date.getTime()), formatDate(snapshot.getCreatedAt())));
```
sns-seb: you don't need `String.format` here. `debug` method supports parameters

server/sonar-server/src/main/java/org/sonar/server/computation/measure/Measure.java
julienlancelot: Indeed, I can re-add it, I thought I need to create a sub class but I was wrong

server/sonar-server/src/main/java/org/sonar/server/computation/period/Period.java
sns-seb: in addition, you should not add a `snapshotId` property but a `Snapshot` class with id and date properties

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
```diff
@@ -0,0 +1,234 @@
...
+          Double pastValue = pastMeasure.getValue();
+          if (pastValue != null) {
+            Metric metric = metricRepository.getByKey(metricKey);
+            measuresWithVariationRepository.add(metric, measure, period, computeVariation(measure, pastValue));
+          }
+        }
+      }
+    }
+
+    private double computeVariation(Measure measure, Double pastValue) {
```
sns-seb: should be static

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
```diff
@@ -0,0 +1,234 @@
...
+          return (measure.getBooleanValue() ? 1d : 0d) - pastValue;
+        default:
+          throw new IllegalArgumentException("Unsupported Measure.ValueType " + measure.getValueType());
+      }
+    }
+  }
+
+  private static final class MeasuresWithVariationRepository {
+
+    private Map<MeasureKey, MeasureWithVariations> measuresWithVariations = new HashMap<>();
```
sns-seb: should be final

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
```diff
@@ -0,0 +1,234 @@
...
+    public Collection<MeasureWithVariations> measures() {
+      return measuresWithVariations.values();
+    }
+  }
+
+  private static final class MeasureWithVariations {
+    private final Metric metric;
+    private final Measure measure;
+
+    private Double[] variations = new Double[5];
```
sns-seb: should be final

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
```diff
@@ -0,0 +1,234 @@
...
+        return new MeasureKey(metric.getKey(), input.getCharacteristicId(), input.getRuleId());
+      }
+    };
+  }
+
+  @Override
+  public void execute() {
+    final DbSession dbSession = dbClient.openSession(false);
+    try {
+      final Iterable<Metric> metrics = FluentIterable.from(metricRepository.getAll()).filter(NumericMetric.INSTANCE);
```
sns-seb: final is not required here

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
```diff
@@ -0,0 +1,234 @@
...
+      final List<Integer> metricIds = FluentIterable.from(metrics).transform(MetricDtoToMetricId.INSTANCE).toList();
```
sns-seb: final is not required here

sns-seb: you should use filter and transform in a single chain. After all, this is what FluentIterable is for, chaining

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
```diff
@@ -0,0 +1,234 @@
...
+      new VariationMeasuresVisitor(dbSession, metricIds).visit(treeRootHolder.getRoot());
+    } finally {
+      dbSession.close();
+    }
+  }
+
+  private class VariationMeasuresVisitor extends DepthTraversalTypeAwareVisitor {
+
+    private final DbSession session;
+    private final List<Integer> metricIds;
```
sns-seb: why don't use use a Map<String, Metric> of Metric by key next to this List of metrid ids. This would save having to call `MeasureRepository.getByKey` in `setVariationMeasure`

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
```diff
@@ -0,0 +1,234 @@
...
+      }
+    }
+
+    private void setVariationMeasures(Component component, List<PastMeasureDto> pastMeasures, int period, MeasuresWithVariationRepository measuresWithVariationRepository) {
+      Map<MeasureKey, PastMeasureDto> pastMeasuresByMeasureKey = FluentIterable.from(pastMeasures).uniqueIndex(pastMeasureToMeasureKey);
+      for (Map.Entry<String, Measure> entry : measureRepository.getRawMeasures(component).entries()) {
+        String metricKey = entry.getKey();
+        Measure measure = entry.getValue();
+        PastMeasureDto pastMeasure = pastMeasuresByMeasureKey.get(new MeasureKey(metricKey, measure.getCharacteristicId(), measure.getRuleId()));
+        if (pastMeasure != null) {
```
sns-seb: this two imbricated `!= null` tests are not very nice to look at

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
```diff
@@ -0,0 +1,234 @@
...
+    @Override
+    public void visitAny(Component component) {
+      MeasuresWithVariationRepository measuresWithVariationRepository = new MeasuresWithVariationRepository();
+      for (Period period : periodsHolder.getPeriods()) {
+        List<PastMeasureDto> pastMeasures = dbClient.measureDao().selectByComponentUuidAndProjectSnapshotIdAndMetricIds(session, component.getUuid(), period.getSnapshotId(),
+          metricIds);
+        setVariationMeasures(component, pastMeasures, period.getIndex(), measuresWithVariationRepository);
+      }
+
+      for (MeasureWithVariations measureWithVariations : measuresWithVariationRepository.measures()) {
```
sns-seb: side effet of having Measure be an immutable object, it forced you to separate computing the variations from storing them
even better would be to use two private method to emphasis this separation of concern

server/sonar-server/src/main/java/org/sonar/server/measure/persistence/MeasureDao.java
```diff
@@ -53,6 +55,18 @@ public MeasureDto findByComponentKeyAndMetricKey(DbSession session, String compo
     });
   }
 
+  public List<PastMeasureDto> selectByComponentUuidAndProjectSnapshotIdAndMetricIds(final DbSession session, final String componentUuid,
+    final long projectSnapshotId,
+    List<Integer> metricIds) {
```
sns-seb: should you take a Set rather than a list?
you don't care about the order of ids but you really need to not request twice (or more) for a specified metric id, hence a Set

server/sonar-server/src/main/java/org/sonar/server/computation/metric/MetricImpl.java
julienlancelot: Indeed this method should be kept private to the step

server/sonar-server/src/test/resources/org/sonar/server/measure/persistence/MeasureDaoTest/past_measures_with_characteristic_id.xml
```diff
@@ -0,0 +1,45 @@
...
+                   created_at="2013-11-20" updated_at="2013-11-22"/>
+
+  <!-- project -->
+  <projects long_name="[null]" id="1" scope="PRJ" qualifier="TRK" kee="project" name="project"
+            root_id="[null]" uuid="ABCD" project_uuid="ABCD" module_uuid="[null]" module_uuid_path=".ABCD."
+            description="[null]"
+            enabled="true" language="java" copy_resource_id="[null]"/>
+
+  <!-- snapshots -->
+  <snapshots purge_status="[null]" period1_mode="[null]" period1_param="[null]" period1_date="[null]" period2_mode="[null]" period2_param="[null]" period2_date="[null]"
```
sns-seb: do you really need to define all these (mostly null) properties on snapshot?

sns-seb:
@julienlancelot another code review comment flood for you ;)

server/sonar-server/src/main/java/org/sonar/server/computation/period/Period.java
julienlancelot: I disagree with your second pint, adding a Snapshot class add a stronger dependency to Snapshot when we only need 2 fields...

julienlancelot: For the first point, the first who merged into master win :)

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
julienlancelot: Indeed ! At the begin, I was needed the 2 lists, but as now I only need one, it's indeed better to chain

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
julienlancelot: Good point !

server/sonar-server/src/main/java/org/sonar/server/computation/period/Period.java
sns-seb: I diidn't mean to use the any existing Snapshot class, but specific one

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
julienlancelot: Indeed, I've added a hasValue() in the dto

server/sonar-server/src/main/java/org/sonar/server/measure/persistence/MeasureDao.java
julienlancelot: exact

server/sonar-server/src/main/java/org/sonar/server/computation/period/Period.java
julienlancelot: A internal class of Period do you mean ? Is it worth it just to remove one parameter ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/FeedPeriodsStep.java
julienlancelot: good to know !

server/sonar-server/src/main/java/org/sonar/server/computation/period/Period.java
sns-seb: a snapshot is something the period is associated to, so having a object for that is relevant
but it saves only one property... for now it may not be worth it

server/sonar-server/src/test/resources/org/sonar/server/measure/persistence/MeasureDaoTest/past_measures_with_characteristic_id.xml
julienlancelot: No :)

julienlancelot:
@seb feedback applied !

server/sonar-server/src/main/java/org/sonar/server/computation/period/Period.java
julienlancelot: ok

julienlancelot:
@sns-seb Is it ok for you ?

julienlancelot:
@seb ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
```diff
@@ -0,0 +1,253 @@
...
+
+  private final DbClient dbClient;
+  private final TreeRootHolder treeRootHolder;
+  private final PeriodsHolder periodsHolder;
+  private final MetricRepository metricRepository;
+  private final MeasureRepository measureRepository;
+
+  private final Function<PastMeasureDto, MeasureKey> pastMeasureToMeasureKey;
+
+  public FillMeasuresWithVariationsStep(DbClient dbClient, TreeRootHolder treeRootHolder, PeriodsHolder periodsHolder, final MetricRepository metricRepository,
```
sns-seb: you can save the final keyword on the constructor argument, by using `this.metricRepository` in your function, which in turns means that you can initialize the `pastMeasureToMeasureKey` where it is declared (therefor having a smaller constructor)

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
```diff
@@ -0,0 +1,253 @@
...
+      return metric.getKey();
+    }
+  }
+
+  private enum NumericMetric implements Predicate<Metric> {
+    INSTANCE;
+
+    @Override
+    public boolean apply(@Nonnull Metric metric) {
+      return Measure.ValueType.INT.equals(metric.getType().getValueType())
```
sns-seb: you could use a variable to store `metric.getType().getValueType()`

sns-seb:
some minor new comments and then LGTM

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
julienlancelot: But then the function will be re created on each call ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/FillMeasuresWithVariationsStep.java
julienlancelot: Ok

PR #380, opened 2015-06-17
fix platform language repository + rename it

sns-seb:
fix platform language repository injection when there is no language plugin in SQ + rename it


simonbrandhof:
LGTM

PR #384, opened 2015-06-18
SONAR-6620 compute QualityGate metrics in CE

sns-seb:
this PR:
* adds a MeasureAssert to be used by any test
* adds a step to compute QualityGate measure and update measures for the metrics configured in the conditions of the QualityGate
* update and improve PeriodHolder

the removal of the batch side computation of the measures moved to CE will be added to this PR as a new and separate commit


sns-seb:
PR updated:
* fix various bug which broke IT (some not related to my code, property sonar.qualitygate wasn't updated when changed by WS)
* removed code from batch
* rebased on master

server/sonar-server/src/main/java/org/sonar/server/computation/qualitygate/ConditionEvaluator.java
```diff
@@ -0,0 +1,190 @@
...
+
+  private static Optional<Double> getPeriodValue(Measure measure, int period) {
+    if (!measure.hasVariations()) {
+      return Optional.absent();
+    }
+
+    MeasureVariations variations = measure.getVariations();
+    switch (period) {
+      case 1:
+        return variations.hasVariation1() ? Optional.of(variations.getVariation1()) : Optional.<Double>absent();
```
julienlancelot: I think it would be much better to use variations.getVariation(in i)

server/sonar-server/src/main/java/org/sonar/server/computation/qualitygate/QualityGateHolderImpl.java
```diff
@@ -19,27 +19,45 @@
...
+      throw new IllegalStateException("QualityGateHolder can be initialized only once");
     }
-    this.qualityGate = qualityGate;
   }
 
   @Override
-  public QualityGate getQualityGate() {
-    if (qualityGate == null) {
+  public Optional<QualityGate> getQualityGate() {
+    if (!initialized) {
```
julienlancelot: Why not use Preconditions.checkState() ?

server/sonar-server/src/main/java/org/sonar/server/computation/step/QualityGateLoadingStep.java
```diff
@@ -77,16 +75,22 @@ private void executeForProject(Component project) {
...
+      if (qualityGate.isPresent()) {
+        qualityGateHolder.setQualityGate(qualityGate.get());
+      }
+      else {
+        qualityGateHolder.setNoQualityGate();
+      }
     } catch (NumberFormatException e) {
       LOGGER.error(String.format("Unsupported value in property %s, using empty Quality Gate", PROPERTY_QUALITY_GATE));
-      qualityGateHolder.setQualityGate(DEFAULT_QUALITY_GATE);
+      qualityGateHolder.setNoQualityGate();
```
julienlancelot: Why not fail the process ? Logging an error in server console will never be seen by users...

server/sonar-server/src/test/java/org/sonar/server/computation/measure/MeasureAssert.java
```diff
@@ -0,0 +1,333 @@
...
+    isNotNull();
+
+    if (actual.hasVariations()) {
+      failWithMessage("Expected Measure to have a no Variations but it did");
+    }
+
+    return this;
+  }
+
+  public MeasureAssert hasVariation1(double expected) {
```
julienlancelot: I would have create a single hasVariation(int period, double expected)

server/sonar-server/src/test/java/org/sonar/server/computation/step/QualityGateMeasuresStepTest.java
```diff
@@ -0,0 +1,263 @@
...
+    ArgumentCaptor<Measure> equals1WarningConditionMeasureCaptor = ArgumentCaptor.forClass(Measure.class);
+
+    verify(measureRepository).getRawMeasure(PROJECT_COMPONENT, INT_METRIC_1);
+    verify(measureRepository).getRawMeasure(PROJECT_COMPONENT, INT_METRIC_2);
+    verify(measureRepository).update(same(PROJECT_COMPONENT), same(INT_METRIC_1), equals1ErrorConditionMeasureCaptor.capture());
+    verify(measureRepository).update(same(PROJECT_COMPONENT), same(INT_METRIC_2), equals1WarningConditionMeasureCaptor.capture());
+    verify(measureRepository).add(same(PROJECT_COMPONENT), same(ALERT_STATUS_METRIC), alertStatusMeasureCaptor.capture());
+    verify(measureRepository).add(same(PROJECT_COMPONENT), same(QUALITY_GATE_DETAILS_METRIC), qgDetailsMeasureCaptor.capture());
+    verifyNoMoreInteractions(measureRepository);
+
```
julienlancelot: A lot of verify here... Wouldn't have it been better to use a fake implementation of MeasureRepository (or waiting for that use new MeasureRepositoryImpl(null, reportReader, mock(MetricRepository.class), null)) ?
And then just that the repository is containing what you've added / updated ?

julienlancelot:
@seb, it's in your hand !

server/sonar-server/src/main/java/org/sonar/server/computation/step/QualityGateLoadingStep.java
sns-seb: you're right, I'll throw an exception, this can not happen unless user modified the DB directly or something like that

server/sonar-server/src/test/java/org/sonar/server/computation/step/QualityGateMeasuresStepTest.java
sns-seb: as the `verifyNoMoreInteractions` highlights, I want to verify the implementation (I know, it's a tight unit test). It is not supposed to do any else than call these specific add, update and getRawMeasure methods. If anyone changes that, I want the test to break.

server/sonar-server/src/test/java/org/sonar/server/computation/measure/MeasureAssert.java
sns-seb: `hasVariation1(1d)` is shorter and much clearer than `hasVariation(1d, 1)`
first version does not require to know what the second parameter is and there is no wondering which parameter is to be asserted

sns-seb:
feedback applied

server/sonar-server/src/main/java/org/sonar/server/computation/step/QualityGateLoadingStep.java
```diff
@@ -77,16 +75,23 @@ private void executeForProject(Component project) {
...
+      if (qualityGate.isPresent()) {
+        qualityGateHolder.setQualityGate(qualityGate.get());
+      }
+      else {
+        qualityGateHolder.setNoQualityGate();
+      }
     } catch (NumberFormatException e) {
-      LOGGER.error(String.format("Unsupported value in property %s, using empty Quality Gate", PROPERTY_QUALITY_GATE));
-      qualityGateHolder.setQualityGate(DEFAULT_QUALITY_GATE);
+      throw new IllegalStateException(
```
julienlancelot: Yeah, fail fast !

server/sonar-server/src/test/java/org/sonar/server/computation/step/QualityGateMeasuresStepTest.java
julienlancelot: I understand, but it's a lof of verify to check the implementation. If tomorrow the implementation change but with keeping the same behaviour, you'll have to update the test.

server/sonar-server/src/test/java/org/sonar/server/computation/measure/MeasureAssert.java
julienlancelot: Hum, that a question of point of view as it duplicates some code but ok.

julienlancelot:
LGTM

PR #392, opened 2015-06-22
SONAR-6659 ProgressMonitor during file indexation is not properly sto…

dbmeneses:
…pped in case of error


henryju:
LGTM, you could maybe propose a pull request on sslr-squid-bridge or at least report issue to language team. https://github.com/SonarSource/sslr-squid-bridge/blob/master/src/main/java/org/sonar/squidbridge/ProgressReport.java (I took this code from there).

PR #410, opened 2015-07-07
Adding support for loading XML rules into an extended repository

bernardo-macedo:
This changes the class RulesDefinitionXmlLoader in order to receive a NewExtendedRepository as parameter, instead of its subclass NewRepository.  This way, the rules loaded from the XML file not only can be loaded into a new repository but also into an existing repository.

This is particularly useful to in order to add custom checkstyle checks to be analysed by SonarQube in conjunction with Checkstyle's SonarQube plugin. This plugin reads the rules from a repository named "checkstyle". Therefore, in order for custom checks to be analysed they need to be inserted together into the same repository.


sns-seb:
Hi,

Thank you for contributing.
Unfortunately, I can not review your Pull Request in its current state and I will have to close it.
Can you create a

PR #423, opened 2015-07-20
ws system/db_migration_status return the status of a database

teryk:


sns-seb:
some comments then LGTM

PR #428, opened 2015-07-22
Fix closing of resovled manual issue

simonbrandhof:
Fix closing of resolved manual issue

### DIFF
--- a/sonar-core/src/main/java/org/sonar/core/issue/workflow/IssueWorkflow.java
+++ b/sonar-core/src/main/java/org/sonar/core/issue/workflow/IssueWorkflow.java
@@ -91,7 +91,7 @@ public class IssueWorkflow implements Startable {
         .functions(new SetResolution(null), new SetCloseDate(false))
         .build())
 
-      // resolve as false-positive
+    // resolve as false-positive
       .transition(Transition.builder(DefaultTransitions.FALSE_POSITIVE)
         .from(Issue.STATUS_OPEN).to(Issue.STATUS_RESOLVED)
         .functions(new SetResolution(Issue.RESOLUTION_FALSE_POSITIVE), UnsetAssignee.INSTANCE)
@@ -108,7 +108,7 @@ public class IssueWorkflow implements Startable {
         .requiredProjectPermission(UserRole.ISSUE_ADMIN)
         .build())
 
-      // resolve as won't fix
+    // resolve as won't fix
       .transition(Transition.builder(DefaultTransitions.WONT_FIX)
         .from(Issue.STATUS_OPEN).to(Issue.STATUS_RESOLVED)
         .functions(new SetResolution(Issue.RESOLUTION_WONT_FIX), UnsetAssignee.INSTANCE)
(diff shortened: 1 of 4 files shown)


sonar-core/src/main/java/org/sonar/core/issue/workflow/IssueWorkflow.java
```diff
@@ -151,26 +150,19 @@ private void buildAutomaticTransitions(StateMachine.Builder builder) {
...
       .transition(Transition.builder("automaticclose")
         .from(Issue.STATUS_RESOLVED).to(Issue.STATUS_CLOSED)
-        .conditions(IsBeingClosed.INSTANCE)
-        .functions(SetClosed.INSTANCE, new SetCloseDate(true))
-        .automatic()
-        .build())
-      .transition(Transition.builder("automaticclosemanual")
-        .from(Issue.STATUS_RESOLVED).to(Issue.STATUS_CLOSED)
-        .conditions(IsManual.INSTANCE)
+        .conditions(new OrCondition(IsBeingClosed.INSTANCE, IsManual.INSTANCE))
```
sns-seb: was the bug that having two transitions with the same "from" and "to", only the last one was kept?
if so, some check might be wise to avoid future similar bug, WDYT?

sonar-core/src/test/java/org/sonar/core/issue/workflow/IssueWorkflowTest.java
```diff
@@ -139,7 +139,7 @@ public void fail_if_unknown_status_when_listing_transitions() {
   }
 
   @Test
-  public void do_automatic_transition() {
+  public void automatically_close_resolved_issue() {
```
sns-seb: I find it odd to see only formatting changes in the unit test.
There does not seem to be a test covering closing an issue for a manual rule (which might explain why the bug existed, if I understood correctly the fix).

sns-seb:
@simonbrandhof back to you

sonar-core/src/main/java/org/sonar/core/issue/workflow/IssueWorkflow.java
simonbrandhof: that's why ```State.outAutomaticTransition()``` fails if there are multiple possible transitions.

sonar-core/src/test/java/org/sonar/core/issue/workflow/IssueWorkflowTest.java
simonbrandhof: the test was badly named but it does not relate to the bug being fixed. Another test is missing.

sonar-core/src/main/java/org/sonar/core/issue/workflow/IssueWorkflow.java
sns-seb: `State.outAutomaticTransition()` is it something new? If not, why did the previous code not fail?
(trying to understand this fix)

sns-seb:
@simonbrandhof LGTM
the question about `State.outAutomaticTransition()` bothers me though. Either it is not called, or it does not work (or I missed something)

simonbrandhof:
Let's discuss your last comment offline.

PR #434, opened 2015-07-23
SONAR-6605 Move computation of formula measures to Compute Engine

julienlancelot:


server/sonar-server/src/test/java/org/sonar/server/computation/step/ComplexityMeasuresStepTest.java
```diff
@@ -0,0 +1,224 @@
...
+    measureRepository.addRawMeasure(FILE_1_REF, metricKey, newMeasureBuilder().create(10));
+    measureRepository.addRawMeasure(FILE_2_REF, metricKey, newMeasureBuilder().create(40));
+
+    underTest.execute();
+
+    assertThat(measureRepository.getAddedRawMeasure(FILE_1_REF, metricKey)).isAbsent();
+    assertThat(measureRepository.getAddedRawMeasure(FILE_2_REF, metricKey)).isAbsent();
+
+    int expectedNonFileValue = 50;
+    assertThat(toEntries(measureRepository.getAddedRawMeasures(DIRECTORY_REF))).contains(entryOf(metricKey, newMeasureBuilder().create(expectedNonFileValue)));
```
sns-seb: could not use `containsOnly` here? too bad, test is weaker with `contains`

julienlancelot: No, because I only want to test to aggregation of one specific measure. As another measure is generated, it's not possible to use containsOnly. Otherwise it will test 2 different measures, and it's not what I want.

sns-seb:
I'm a little overwhelmed by the this PR but LGTM

PR #442, opened 2015-07-27
SONAR-6726 Authentication freeze with LDAP and Dev Cockpit

julienlancelot:


server/sonar-web/src/main/webapp/WEB-INF/lib/need_authentication.rb
```diff
@@ -153,13 +153,15 @@ def synchronize(username, password, details)
...
+
+      # This save is to used to set a lock on the users table
       user.save(false)
 
       synchronize_groups(user)
-      user.notify_creation_handlers
-      user
+      user.save(false)
     end
+    user.notify_creation_handlers
```
simonbrandhof: The point of this patch is that ```user.notify_creation_handlers``` is outside the RoR transaction. A comment should be added, with a link to JIRA ticket.

simonbrandhof:
Comment is missing then LGTM

PR #458, opened 2015-08-04
Update sonar.properties

PierreR:
Fix default `-Xmx` value for `sonar.search.javaOpts` in comment

### DIFF
--- a/sonar-application/src/main/assembly/conf/sonar.properties
+++ b/sonar-application/src/main/assembly/conf/sonar.properties
@@ -206,7 +206,7 @@
 #    Set min and max memory (respectively -Xms and -Xmx) to the same value to prevent heap
 #    from resizing at runtime.
 #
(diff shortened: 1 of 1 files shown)


simonbrandhof:
I don't get your point. The effective default value is 1G, see https://github.com/SonarSource/sonarqube/blob/feature/sb/issue-location-ws/server/sonar-process/src/main/java/org/sonar/process/ProcessProperties.java#L110-110.

PierreR:
It is actually `-Xmx256m` in the LTS version ... Given the above comment (in the config file)
```
Set min and max memory (respectively -Xms and -Xmx) to the same value to prevent heap from resizing at runtime.
```
I didn't assume it would have been changed that way ... Maybe the comment should be removed ?

simonbrandhof:
You're right. This comment was removed from master. Thanks.

PR #464, opened 2015-08-05
SONAR-6749 views in 5.2

sns-seb:
this PR adds:
* the ability to run views the old way plugin were run: as part of the WebServer
* the ViewsBridge interface which is the entry point from the WebServer into Views
* some DAO changes to be able to run basic code from views

### DIFF
--- a/sonar-db/src/main/java/org/sonar/db/component/ComponentMapper.java
+++ b/sonar-db/src/main/java/org/sonar/db/component/ComponentMapper.java
@@ -53,12 +53,23 @@ public interface ComponentMapper {
 
   List<ComponentDto> selectByUuids(@Param("uuids") Collection<String> uuids);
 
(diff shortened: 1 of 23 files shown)


sonar-db/src/main/java/org/sonar/db/component/ComponentMapper.java
```diff
@@ -53,6 +53,10 @@
 
   List<ComponentDto> selectByUuids(@Param("uuids") Collection<String> uuids);
 
+  List<ComponentDto> selectByProjectUuid(@Param("projectUuid") String projectUuid);
+
+  List<ComponentDto> selectByCustomMeasure(@Param("metricKey") String metricKey, @Param("metricValue") String metricValue);
```
teryk: views creates custom measures ?

teryk:
not a server admin expert but it LGTM

PR #478, opened 2015-08-21
Allow two different submodules to have the same name.

abbeyj:
This also covers the case where a submodule has the same name as its parent.

This addresses https://jira.sonarsource.com/browse/SONAR-6665
and https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/sonarqube/9FnKTyp0r5E/pc8YVM20AgAJ


abbeyj:
This keeps getting out of date because of changes made on top of https://github.com/SonarSource/sonarqube/commit/a805e4796f8826acc73f2078e7374f1956402dfd .  IMO, that change should be reverted since this change solves the underlying problem instead of just erroring out on perfectly valid directory structures.

The longer this goes on the worse the problem gets.  At this point I'm getting a bit tired of rebasing this over and over again.  Could somebody take a look at this sooner rather than later?

henryju:
Hi @abbeyj . Sorry for the delay. I'll take care of that either today or on Monday.

henryju:
@abbeyj It was a bit hard to merge your P/R because of the number of different commits. We usually tend to prefer P/R with a single commit to preserve a clean history. So I have manually applied your fix in https://github.com/SonarSource/sonarqube/commit/2e24cfd57b9ffa406a95018f23bc5034567a761a. Anyway many thanks for your contribution!

PR #496, opened 2015-09-07
SONAR-6749 Add query to get custom measures from a metric key and a text value

julienlancelot:
SONAR-6749 Add query to get custom measures from a metric key and a text value


sns-seb:
LGTM

PR #506, opened 2015-09-14
Simplify deletion of a project in PurgeDao

julienlancelot:
Simplify deletion of a project in PurgeDao

- Rename deleteResourceTree to deleteProject
- deleteProject only takes a uuid as paramter (instead of a IdUUidPair)


sns-seb:
small doc comments, LGTM

PR #507, opened 2015-09-15
No op for highlighting and symbol references in issues mode

henryju:


sonar-batch/src/main/java/org/sonar/batch/source/DefaultHighlightable.java
```diff
@@ -32,15 +33,17 @@
...
-  public DefaultHighlightable(DefaultInputFile inputFile, SensorStorage sensorStorage) {
+  public DefaultHighlightable(DefaultInputFile inputFile, SensorStorage sensorStorage, AnalysisMode analysisMode) {
     this.inputFile = inputFile;
     this.sensorStorage = sensorStorage;
+    this.analysisMode = analysisMode;
   }
 
   @Override
   public HighlightingBuilder newHighlighting() {
-    DefaultHighlighting defaultHighlighting = new DefaultHighlighting(sensorStorage);
```
dbmeneses: Wouldn't it be better if the builder builds a no-op object (NoOpHighlighting), leaving DefaultHighlighting unchanged?

dbmeneses:
LGTM appart of the comment, that also applies to SymbolTable.

henryju:
@dbmeneses Good idea, please have a look at new implementation.

dbmeneses:
LGTM

PR #512, opened 2015-09-16
SONAR-6845 Add index on PROJECTS.QUALIFIER

julienlancelot:


sns-seb:
from what I've seen in the mapper, this index will very likely improve performance on many requests

LGTM

PR #514, opened 2015-09-18
SONAR-6840 Allow measure values of integers with more than 10 digits in database

julienlancelot:
The migration has been updated to be able to group migration that are made on the same table


sonar-db/src/main/java/org/sonar/db/version/UpdateColumnsBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+import org.sonar.db.dialect.PostgreSql;
+
+import static com.google.common.collect.Lists.newArrayList;
+
+/**
+ * Generate SQL queries to update multiple column types from a table.
+ *
+ * Note that it's not possible to change the nullable state of a column (Because on Oracle, sending a query to set a column to be nullable when it's already nullable will fail).
+ */
+public class UpdateColumnsBuilder {
```
simonbrandhof: I suggest to rename this class ```AlterTableBuilder``` as "alter" is the standard term.

sonar-db/src/main/java/org/sonar/db/version/DecimalColumnDef.java
```diff
@@ -0,0 +1,117 @@
...
+  public int getScale() {
+    return scale;
+  }
+
+  @Override
+  public String generateSqlType(Dialect dialect) {
+    switch (dialect.getId()) {
+      case PostgreSql.ID :
+      case Oracle.ID :
+        return "NUMERIC (" + Integer.toString(precision)  + "," + Integer.toString(scale) + ")";
```
simonbrandhof: Be ready for the issue about ```String.format()``` :-)

sonar-db/src/main/java/org/sonar/db/version/DecimalColumnDef.java
```diff
@@ -0,0 +1,117 @@
...
+      case MySql.ID :
+      case MsSql.ID :
+        return "DECIMAL (" + Integer.toString(precision)  + "," + Integer.toString(scale) + ")";
+      default :
```
simonbrandhof: I suggest to add the H2 case and fail unsupported dialects in ```default```

sonar-db/src/main/java/org/sonar/db/version/DecimalColumnDef.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  public static class Builder {
+    @CheckForNull
+    private String columnName;
+    @CheckForNull
+    private Integer precision;
+    @CheckForNull
+    private Integer scale;
+    private boolean isNullable;
```
simonbrandhof: As the default value is false if ```setIsNullable()``` is not called , the field deserves to be explicitly initialized to false.

sonar-db/src/main/java/org/sonar/db/version/DecimalColumnDef.java
```diff
@@ -0,0 +1,117 @@
...
+      return this;
+    }
+
+    public Builder setScale(int scale) {
+      this.scale = scale;
+      return this;
+    }
+
+    public DecimalColumnDef build() {
+      validateColumnName(columnName);
```
simonbrandhof: Just for my information, why using static methods with field as parameter ? Why not object methods ```checkPrecision()``` and ```checkScale()``` without parameters ?

sonar-db/src/main/java/org/sonar/db/version/UpdateColumnsBuilder.java
simonbrandhof: Oh in fact you make the difference because of ```DropColumnsBuilder``` ?

sonar-db/src/main/java/org/sonar/db/version/v52/IncreasePrecisionOfNumerics.java
```diff
@@ -0,0 +1,74 @@
...
+  private void executeSql(DdlChange.Context context, String table, String... columns) throws SQLException {
+    for (String sql : generateSql(table, columns)) {
+      context.execute(sql);
+    }
+  }
+
+  private List<String> generateSql(String table, String... columns) {
+    UpdateColumnsBuilder columnsBuilder = new UpdateColumnsBuilder(getDatabase().getDialect(), table);
+    for (String column : columns) {
+      columnsBuilder.updateColumn(newDecimalColumnDefBuilder().setColumnName(column).setPrecision(38).setScale(20).build());
```
simonbrandhof: The values 38 and 20 are supposed to be the standard precision and scale to be used for future new decimal columns, so they deserve to be defined in DecimalColumnDef (constants or default field values in builder).

sonar-db/src/main/java/org/sonar/db/version/UpdateColumnsBuilder.java
julienlancelot: No you're right AlterColumnsBuilder should be better

sonar-db/src/main/java/org/sonar/db/version/DecimalColumnDef.java
julienlancelot: Right !

sonar-db/src/main/java/org/sonar/db/version/DecimalColumnDef.java
julienlancelot: Ok

sonar-db/src/main/java/org/sonar/db/version/DecimalColumnDef.java
julienlancelot: ok

sonar-db/src/main/java/org/sonar/db/version/DecimalColumnDef.java
julienlancelot: In order to use it in the setter and in the builder (see line 82)

sonar-db/src/main/java/org/sonar/db/version/v52/IncreasePrecisionOfNumerics.java
julienlancelot: Oh yes, so in fact fields setPrecision et setScale should be removed. Right ?

simonbrandhof: No, they're valuable and must be kept, but they should be optional. By default the fields of builder must be initialized to 38 and 20

julienlancelot: ok

julienlancelot:
@simon Feedbacks applied !

PR #516, opened 2015-09-18
Improve management of compute engine queue

simonbrandhof:
- SONAR-6834 Add search and filtering capabilities to the "Project Computation" page
- SONAR-6630 delete past reports older than 6 months
- SONAR-6851 Analysis of new projects requires the "Provision Project" permission
- SONAR-6861 Drop the property "sonar.preventAutoProjectCreation"


sns-seb:
stopping review unfinished. We need to talk @simonbrandhof :)

PR #517, opened 2015-09-18
SONAR-6347 Increase max value of duplications_index id column

julienlancelot:
SONAR-6347 Increase max value of duplications_index id column


simonbrandhof:
The PR should rebase master as db migration 933 is already reserved.
Else LGTM

PR #522, opened 2015-09-21
SONAR-6843 Read file sources from the report, then from the db (if report contains no file source)

julienlancelot:


server/sonar-server/src/main/java/org/sonar/server/computation/source/FileSourceRepository.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.sonar.server.computation.source;
+
+import org.sonar.core.util.CloseableIterator;
+import org.sonar.server.computation.component.Component;
+
+public interface FileSourceRepository {
+
+  /**
+   * Return lines from a given component. If file sources is not in the report then we read it the database.
```
simonbrandhof: "read it from the database"

server/sonar-server/src/main/java/org/sonar/server/computation/source/FileSourceRepositoryImpl.java
```diff
@@ -0,0 +1,85 @@
...
+
+    Optional<CloseableIterator<String>> linesIteratorOptional = reportReader.readFileSource(component.getReportAttributes().getRef());
+    if (linesIteratorOptional.isPresent()) {
+      return linesIteratorOptional.get();
+    }
+    DbSession session = dbClient.openSession(false);
+    try {
+      return readLinesFromDb(session, component);
+    } finally {
+      session.close();
```
simonbrandhof: Could be replaced by ```dbClient.closeSession(session)``` to not re-throw errors.

server/sonar-server/src/main/java/org/sonar/server/computation/source/FileSourceRepositoryImpl.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  private CloseableIterator<String> readLinesFromDb(DbSession session, Component component){
```
simonbrandhof: It would be great to replace this method by ```SourceService.getLinesAsRawText(dbSession, componentUuid, 1, Integer.MAX_VALUE)```, which returns ```Optional<Iterable<String>>```, but not a ```CloseableIterator<String>```

server/sonar-server/src/main/java/org/sonar/server/computation/source/FileSourceRepositoryImpl.java
julienlancelot: Indeed !

server/sonar-server/src/main/java/org/sonar/server/computation/source/FileSourceRepository.java
julienlancelot: ok

server/sonar-server/src/main/java/org/sonar/server/computation/source/FileSourceRepositoryImpl.java
julienlancelot: Good point !

PR #523, opened 2015-09-21
SONAR-5770 add WS api/ce/project

simonbrandhof:
This web service is used to get the instant status of a project
in compute engine:

- tasks in queue (pending/in progress)
- the lastly executed task


SonarTech:
* MAJOR 1 major

* MAJOR [Make "buildComponent" a "static" method.](https://github.com/SonarSource/sonarqube/blob/c60690251a87a3b707aafbeea5477ff80932dd5c/server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeWsTaskFormatter.java#L116) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS2325)

server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeActivityWsAction.java
```diff
@@ -115,7 +115,7 @@ private CeActivityQuery readQuery(Request wsRequest) {
     if (componentId == null) {
       userSession.checkGlobalPermission(UserRole.ADMIN);
     } else {
-      userSession.checkProjectUuidPermission(UserRole.ADMIN, componentId);
+      userSession.checkProjectUuidPermission(UserRole.USER, componentId);
```
teryk: I would name the variable componentUuid

server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeProjectWsAction.java
```diff
@@ -0,0 +1,93 @@
...
+import org.sonar.db.DbSession;
+import org.sonar.db.ce.CeActivityDto;
+import org.sonar.db.ce.CeActivityQuery;
+import org.sonar.db.ce.CeQueueDto;
+import org.sonar.server.user.UserSession;
+import org.sonar.server.ws.WsUtils;
+import org.sonarqube.ws.WsCe;
+
+/**
+ * GET api/ce/project
```
teryk: You think this javadoc is useful ? I ask because I never document my WS actions. Should I ?

server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeProjectWsAction.java
```diff
@@ -0,0 +1,93 @@
...
+ * <p>Get the pending and last executed tasks of a given project</p>
+ */
+public class CeProjectWsAction implements CeWsAction {
+
+  public static final String PARAM_COMPONENT_ID = "componentId";
```
teryk: I would rename PARAM_COMPONENT_ID to PARAM_COMPONENT_UUID

server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeProjectWsAction.java
```diff
@@ -0,0 +1,93 @@
...
+
+  private final UserSession userSession;
+  private final DbClient dbClient;
+  private final CeWsTaskFormatter formatter;
+
+  public CeProjectWsAction(UserSession userSession, DbClient dbClient, CeWsTaskFormatter formatter) {
```
teryk: We decided to name action by their name, in this case it should be renamed ProjectAction. It would consistent over the whole stack

server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeProjectWsAction.java
```diff
@@ -0,0 +1,93 @@
...
+    this.dbClient = dbClient;
+    this.formatter = formatter;
+  }
+
+  @Override
+  public void define(WebService.NewController controller) {
+    WebService.NewAction action = controller.createAction("project")
+      .setDescription("Get the pending and last executed tasks of a given project")
+      .setInternal(true)
+      .setResponseExample(getClass().getResource("CeProjectWsAction/example.json"))
```
teryk: usually the example are in the same path than the class and named by their name. In this case it should be project-example.json

server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeProjectWsAction.java
```diff
@@ -0,0 +1,93 @@
...
+      .setHandler(this);
+
+    action.createParam(PARAM_COMPONENT_ID)
+      .setRequired(true)
+      .setExampleValue(Uuids.UUID_EXAMPLE_01);
+  }
+
+  @Override
+  public void handle(Request wsRequest, Response wsResponse) throws Exception {
+    String componentUuid = wsRequest.mandatoryParam(PARAM_COMPONENT_ID);
```
teryk: not linked to this PR but I wonder why we have to specify the mandatoryParam when it's already declared in the define method. Any good reason behind this ?
I would prefer a wsRequest.param(String) without making the difference between mandatory and non mandatory in the handle method. It should be already checked depending on the declaration

server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeProjectWsAction.java
```diff
@@ -0,0 +1,93 @@
...
+
+    DbSession dbSession = dbClient.openSession(false);
+    try {
+      List<CeQueueDto> queueDtos = dbClient.ceQueueDao().selectByComponentUuid(dbSession, componentUuid);
+      CeActivityQuery activityQuery = new CeActivityQuery()
+        .setComponentUuid(componentUuid)
+        .setOnlyCurrents(true);
+      List<CeActivityDto> activityDtos = dbClient.ceActivityDao().selectByQuery(dbSession, activityQuery, new RowBounds(0, 1));
+
+      WsCe.ProjectResponse.Builder wsResponseBuilder = WsCe.ProjectResponse.newBuilder();
```
teryk: in the context of the class, you can import the WsCe

server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeProjectWsAction.java
```diff
@@ -0,0 +1,93 @@
...
+      wsResponseBuilder.addAllQueue(formatter.formatQueue(dbSession, queueDtos));
+      if (activityDtos.size() == 1) {
```
teryk: what happen if there are several activities at the same time ?

teryk: shouldn't we already manage when there'll be several analysis at the same time ?
or we should:
- throw an exception
- clean the db to have only one current activity
- another solution I haven't thought of

server/sonar-server/src/main/java/org/sonar/server/platform/platformlevel/PlatformLevel4.java
```diff
@@ -727,6 +728,7 @@ protected void configureLevel() {
       CeWs.class,
       CeWsTaskFormatter.class,
       CeTaskWsAction.class,
+      CeProjectWsAction.class,
```
teryk: Maybe it's time for a CeWsModule class ? : )

server/sonar-server/src/main/resources/org/sonar/server/computation/ws/CeProjectWsAction/example.json
```diff
@@ -0,0 +1,25 @@
+{
+  "queue": [
+    {
+      "id": "AU_w84A6gAS1Hm6h4_ih",
```
teryk: I would put at least 2 (3 would be perfect) elements in the queue example to show there can be several elements.

server/sonar-server/src/test/java/org/sonar/server/computation/ws/CeProjectWsActionTest.java
```diff
@@ -0,0 +1,111 @@
...
+      .execute();
+
+    WsCe.ProjectResponse response = Protobuf.read(wsResponse.getInputStream(), WsCe.ProjectResponse.PARSER);
+    assertThat(response.getQueueCount()).isEqualTo(0);
+    assertThat(response.hasCurrent()).isFalse();
+  }
+
+  @Test
+  public void project_tasks() {
+    userSession.addProjectUuidPermissions(UserRole.USER, "PROJECT_1");
```
teryk: I prefer to set the user session in the setup method. You change it if it's needed

server/sonar-server/src/test/java/org/sonar/server/computation/ws/CeProjectWsActionTest.java
```diff
@@ -0,0 +1,111 @@
...
+  @Test
+  public void empty_queue_and_empty_activity() {
+    userSession.addProjectUuidPermissions(UserRole.USER, "PROJECT_1");
+
+    TestResponse wsResponse = tester.newRequest()
+      .setParam("componentId", "PROJECT_1")
+      .setMediaType(MimeTypes.PROTOBUF)
+      .execute();
+
+    WsCe.ProjectResponse response = Protobuf.read(wsResponse.getInputStream(), WsCe.ProjectResponse.PARSER);
```
teryk: I love testing Protobuf Message directly, I should have done it before!

server/sonar-server/src/test/java/org/sonar/server/computation/ws/CeProjectWsActionTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+  CeWsTaskFormatter formatter = new CeWsTaskFormatter(dbTester.getDbClient());
+  CeProjectWsAction underTest = new CeProjectWsAction(userSession, dbTester.getDbClient(), formatter);
+  WsActionTester tester = new WsActionTester(underTest);
+
+  @Test
+  public void empty_queue_and_empty_activity() {
+    userSession.addProjectUuidPermissions(UserRole.USER, "PROJECT_1");
+
+    TestResponse wsResponse = tester.newRequest()
```
teryk: you can extract the method with the componentId parameter and the mime type

server/sonar-server/src/test/java/org/sonar/server/computation/ws/CeProjectWsActionTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.sonar.db.ce.CeTaskTypes;
+import org.sonar.server.plugins.MimeTypes;
+import org.sonar.server.tester.UserSessionRule;
+import org.sonar.server.ws.TestResponse;
+import org.sonar.server.ws.WsActionTester;
+import org.sonarqube.ws.WsCe;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class CeProjectWsActionTest {
```
teryk: You need to test the json example. If you want to ignore a field, a great developer has already provided an easy way to do that: JsonAssert.ignoreFields. You shouldn't need it as you don't have db ids.

teryk:
Apply my feedbacks and LGTM

server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeProjectWsAction.java
simonbrandhof: param() returns a nullable string whereas mandatoryParam() returns a non-null string. These methods could be renamed though.

server/sonar-server/src/main/java/org/sonar/server/computation/ws/CeProjectWsAction.java
simonbrandhof: It's not possible to have more than 1 activity because of ```RowBound(0,1)```.

server/sonar-server/src/test/java/org/sonar/server/computation/ws/CeProjectWsActionTest.java
simonbrandhof: not me, because I like reading my test in a single shot. Moreover it prevents from writing a test (now or later) that does not need such permission.

teryk: I understand. It's ok here but when you have dozens of tests and in 90% of the cases you don't bother about permissions, it's simpler IMO

PR #525, opened 2015-09-22
SONAR-6548 Log a message on deprecated projects

julienlancelot:
SONAR-6548 Log a message on deprecated projects


sns-seb:
seems to do the job

small comment on method naming then LGTM

PR #529, opened 2015-09-22
SONAR-6143 add new_sqale_debt_ratio to default quality gate

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/qualitygate/RegisterQualityGates.java
```diff
@@ -59,6 +59,7 @@ private void createBuiltinQualityGate() {
     QualityGateDto builtin = qualityGates.create(BUILTIN_QUALITY_GATE);
     qualityGates.createCondition(builtin.getId(), CoreMetrics.BLOCKER_VIOLATIONS_KEY, QualityGateConditionDto.OPERATOR_GREATER_THAN, null, "0", null);
     qualityGates.createCondition(builtin.getId(), CoreMetrics.CRITICAL_VIOLATIONS_KEY, QualityGateConditionDto.OPERATOR_GREATER_THAN, null, "0", 3);
+    qualityGates.createCondition(builtin.getId(), CoreMetrics.NEW_SQALE_DEBT_RATIO_KEY, QualityGateConditionDto.OPERATOR_GREATER_THAN, null, "5", 3);
```
simonbrandhof: I didn't test that limit is 5 (%) but not 0.05.

simonbrandhof:
LGTM if you verified that threshold is 5 but not 0.05.

server/sonar-server/src/main/java/org/sonar/server/qualitygate/RegisterQualityGates.java
sns-seb: well, ratio between debt and dev cost is multiplied by 100 before storage, so I'm pretty sure we should compare to 5

PR #535, opened 2015-09-25
SONAR-6665 Fix StackOverflow error when analyzing project with severa…

henryju:
…l modules having same key


dbmeneses:
LGTM if ITs pass

PR #536, opened 2015-09-25
SONAR-6547 Clear rules overloaded debt

julienlancelot:
SONAR-6547 Clear rules overloaded debt


julienlancelot:
FYI, build fails with Java 8, should be done with Java 6/7

julienlancelot:
FYI, build fails with Java 8, should be done with Java 6/7

server/sonar-server/src/main/java/org/sonar/server/platform/ServerComponents.java
```diff
@@ -20,6 +20,9 @@
 package org.sonar.server.platform;
 
 import com.google.common.collect.Lists;
+import java.util.Arrays;
```
simonbrandhof: lots of changes on imports. Is that expected ?

simonbrandhof:
LGTM

server/sonar-server/src/main/java/org/sonar/server/platform/ServerComponents.java
julienlancelot: It's because of changes in configuration on import

PR #538, opened 2015-09-25
SONAR-6143 update default QG to be fully based on water leak period

sns-seb:
SONAR-6143 was incompletely implemented, my bad


julienlancelot:
LGTM

PR #539, opened 2015-09-28
SONAR-6332 Prevent provisioning projects with same key on MySQL

julienlancelot:
SONAR-6332 Prevent provisioning projects with same key on MySQL


server/sonar-server/src/main/java/org/sonar/server/component/ComponentService.java
```diff
@@ -174,19 +179,37 @@ public ComponentDto create(NewComponent newComponent) {
...
+  /**
+   * On MySQL, as PROJECTS.KEE is not unique, if the same project is provisioned multiple times, then it will be duplicated in the database.
+   * So, after creating a project, we commit, and we search in the db if their are some duplications and we remove them.
+   *
+   * SONAR-6332
+   */
+  private void removeDuplicatedProjects(String projectKey) {
+    DbSession session = dbClient.openSession(false);
+    try {
+      List<ComponentDto> duplicated = dbClient.componentDao().selectComponentsHavingSameKey(session, projectKey);
```
simonbrandhof: `selectComponentsHavingSameKey` could be renamed with something like  `selectComponentsHavingSameKeyOrderedById` so that we understand why only the first one is kept.

simonbrandhof:
one comment then LGTM

server/sonar-server/src/main/java/org/sonar/server/component/ComponentService.java
julienlancelot: ok

PR #541, opened 2015-09-28
SONAR-6143 'SonarQube way' quality gate should be set as default

julienlancelot:
SONAR-6143 'SonarQube way' quality gate should be set as default


sns-seb:
images

PR #550, opened 2015-09-30
SONAR-6857 WS components/search search for components

teryk:


teryk:
I need to add tests, obviously, but I checked manually and it's working.

teryk:
@stas-vilchik you can redesign the bulk-deletion page on this branch. 3 things to do for you:
- replace calls of api/components/search by api/components/search_view_components
- change the parameter componentUuid to componentId of WS api/components/search_view_components
- rely on api/components/search to search for components

sonar-db/src/main/java/org/sonar/db/component/ComponentQuery.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public ComponentQuery(Database database, @Nullable String nameOrKeyQuery, String... qualifiers) {
+    checkArgument(qualifiers.length > 0, "At least one qualifier must be provided");
+
+    this.database = database;
+    this.nameOrKeyQuery = nameOrKeyQueryToSql(nameOrKeyQuery);
+    this.qualifiers = qualifiers;
+  }
+
+  public String getNameOrKeyQuery() {
```
simonbrandhof: Missing @CheckForNull

sonar-db/src/main/java/org/sonar/db/component/ComponentQuery.java
```diff
@@ -0,0 +1,79 @@
...
+public class ComponentQuery {
+  private final Database database;
+  private final String nameOrKeyQuery;
+  private final String[] qualifiers;
+
+  public ComponentQuery(Database database, @Nullable String nameOrKeyQuery, String... qualifiers) {
+    checkArgument(qualifiers.length > 0, "At least one qualifier must be provided");
+
+    this.database = database;
+    this.nameOrKeyQuery = nameOrKeyQueryToSql(nameOrKeyQuery);
```
simonbrandhof: The SQL representation should not be stored in the field as it hides the initial value. Why not converting to SQL at runtime in a public method `getNameOrKeyQueryAsSql` ?

sonar-db/src/main/java/org/sonar/db/component/ComponentQuery.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public String getNameOrKeyQuery() {
+    return nameOrKeyQuery;
+  }
+
+  public String[] getQualifiers() {
+    return qualifiers;
+  }
+
+  private String nameOrKeyQueryToSql(@Nullable String nameOrKeyQuery) {
```
simonbrandhof: missing @CheckForNull

sonar-db/src/main/java/org/sonar/db/component/ComponentQuery.java
```diff
@@ -0,0 +1,79 @@
...
+    if (nameOrKeyQuery == null) {
+      return null;
+    }
+
+    return appendEscapeForSomeDb(
```
simonbrandhof: this escaping should be handled by a single dedicated method on `Dialect`. What about `Dialect#escapeLikeValue(String)` ?

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
```diff
@@ -250,6 +250,40 @@
...
+      <foreach collection="qualifiers" item="qualifier" open="(" close=")" separator=",">
+        #{qualifier}
+      </foreach>
+      <if test="nameOrKeyQuery!=null">
+        AND ri.qualifier in
+        <foreach collection="qualifiers" item="qualifier" open="(" close=")" separator=",">
+          #{qualifier}
+        </foreach>
+        AND (ri.kee like #{nameOrKeyQuery}
+        OR LOWER(p.kee) like #{nameOrKeyQuery})
```
simonbrandhof: LOWER(p) must absolutely be removed and nameOrKeyQuery must be lower-case.

sonar-db/src/test/java/org/sonar/db/component/ComponentDaoTest.java
```diff
@@ -51,16 +51,17 @@
 
   @Rule
   public DbTester db = DbTester.create(System2.INSTANCE);
+  ComponentDbTester componentDb = new ComponentDbTester(db);
 
-  DbSession dbSession = db.getSession();
+  DbSession session = db.getSession();
```
simonbrandhof: Personally I prefer using the term `dbSession` as it removes ambiguity with `UserSession`

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
```diff
@@ -250,6 +250,40 @@
     ORDER BY UPPER(p.name), p.name
   </select>
 
+  <select id="selectByQuery" resultType="Component">
+    select
+    <include refid="componentColumns"/>
+    <include refid="sqlSelectByQuery" />
+    ORDER BY LOWER(p.name), p.name
```
simonbrandhof: What this order by `LOWER(p.name)` already there in previous versions ?

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
```diff
@@ -250,6 +250,40 @@
...
+    <include refid="sqlSelectByQuery" />
+  </select>
+
+  <sql id="sqlSelectByQuery">
+    from projects p
+    <if test="nameOrKeyQuery!=null">
+      inner join resource_index ri on p.id = ri.resource_id
+    </if>
+    <where>
+      AND p.enabled=${_true}
```
simonbrandhof: Minor - `<where> AND p.enabled` can be replaced by `where p.enabled` to decrease coupling with mybatis.

sonar-db/src/main/java/org/sonar/db/component/ComponentQuery.java
teryk: yep, or I could add a nameOrKeyQueryToSql field, so it's static and if you call the method several times, there's only one computation

simonbrandhof: Premature optimization IMO

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
teryk: nameOrKeyQuerySql is lower case. The problem is that the p.kee isn't… 
The solution I see:
- #{nameOrKeyQueryToSql} is not necessarily in lower case
- `ri.kee like LOWER(#(nameOrKeyQueryToSql)) OR p.kee like #{nameOrKeyQueryToSql}`
WDYT @simonbrandhof ?

sonar-db/src/test/java/org/sonar/db/component/ComponentDaoTest.java
teryk: Ok, I did already that outside dao testing. Why not.

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
simonbrandhof: ok, I mixed with RESOURCE_INDEX. For project key, we said that the search was case-sensitive in this version.

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
teryk: we want ordered components, right ?

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
teryk: I could, I find easier to separate the different parts when looking at the xml file

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
simonbrandhof: my point is not about the need for ordering (of course it's required as the results are paged), but about the LOWER conversion.

sonar-db/src/main/java/org/sonar/db/component/ComponentQuery.java
teryk: there are 2 things:
- append escape for oracle and mssql
- escape percent and underscore and add a wildcard
The added wildcard must not be escaped and we cannot append escape for oracle and mssql.
I suggest 2 methods:
- `Dialect#escapeLikeValueWithPostWildcard(String)`
- `Dialect#escapeLikeValueWithPreAndPostWildard(String)`
WDYT Simon ?

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
teryk: it's done elsewhere, and it permits to have a non case sensitive order consistent on all db

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
teryk: I know, that's what my solution adressed. Case insensitive for project name, case sensitive for project key

sonar-db/src/main/resources/org/sonar/db/component/ComponentMapper.xml
simonbrandhof: ok, this part will be improved when searching for components through a dedicated ES index.

simonbrandhof:
LGTM

PR #559, opened 2015-10-02
SONAR-6749 add ignoreZeroValue flag to VariationSumFormula

sns-seb:
hardcoded behavior to ignore zero value in variations to aggregate is suitable for new_coverage metrics but not for other new_* metrics aggregated by views


julienlancelot:
I know that saving a measure with 0 will be tested in views, but it would have been good to have unit test here. But I understand it's hard to test.
Nevertheless, LGTM.

sns-seb:
done another way in #566

PR #561, opened 2015-10-02
Update requires for purge in views

julienlancelot:


sonar-db/src/main/java/org/sonar/db/component/ComponentDao.java
```diff
@@ -147,6 +148,7 @@ public boolean existsById(Long id, DbSession session) {
     return mapper(session).selectComponentsHavingSameKeyOrderedById(key);
   }
 
+
```
sns-seb: useless line return

sonar-db/src/main/java/org/sonar/db/component/ComponentDao.java
```diff
@@ -251,6 +253,16 @@ public long countGhostProjects(DbSession session, @Nullable String query) {
...
   }
 
+  /**
+   * Retrieve enabled components keys with given qualifiers
+   *
+   * Used by Views plugin
+   */
+  public List<ComponentDto> selectComponentsByQualifiers(DbSession dbSession, Collection<String> qualifiers) {
+    checkArgument(!qualifiers.isEmpty(), "Qualifiers cannot be empty");
+    return mapper(dbSession).selectComponentsByQualifiers(qualifiers);
```
sns-seb: either create a `Set` from `Collection` argument or require a `Set` as argument

sonar-db/src/main/java/org/sonar/db/component/ComponentDao.java
```diff
@@ -251,6 +253,16 @@ public long countGhostProjects(DbSession session, @Nullable String query) {
     return mapper(dbSession).selectByProjectUuid(projectUuid);
   }
 
+  /**
+   * Retrieve enabled components keys with given qualifiers
+   *
+   * Used by Views plugin
```
sns-seb: is there any specific order in the returned list?

sonar-db/src/main/java/org/sonar/db/component/ComponentDao.java
julienlancelot: Indeed, I can require a set as argument, and return a Set

sonar-db/src/main/java/org/sonar/db/component/ComponentDao.java
julienlancelot: Nope

sns-seb:
LGTM

PR #567, opened 2015-10-05
SONAR-6397 Never load scm from db on first analysis

julienlancelot:


sns-seb:
some picky comments then LGTM

PR #580, opened 2015-10-14
SONAR-6880 Fix SSF-37

bellingard:
This fixes SSF-37 (no more details provided here).

This should be tested again, and we should make sure that there's no side effect. If everything's OK, this should be backported to LTS as mentioned in the related JIRA ticket.


simonbrandhof:
Bad branch !

bellingard:
Arghhh, I will close it and reopen a new one on 4.5 branch.

PR #582, opened 2015-10-14
SONAR-6652 explicit error when DB is disconnected in WS response

sns-seb:
fix for 5.2 of WS response of DB migration related WebService when the DB can not be reached
not error message is related to the problem and explicit

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/platform/ws/DbMigrationJsonWriter.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/platform/ws/DbMigrationJsonWriter.java
@@ -37,6 +37,7 @@ public class DbMigrationJsonWriter {
   static final String STATUS_MIGRATION_SUCCEEDED = "MIGRATION_SUCCEEDED";
   static final String STATUS_MIGRATION_REQUIRED = "MIGRATION_REQUIRED";
 
+  static final String NO_CONNECTION_TO_DB = "Cannot connect to Database.";
   static final String UNSUPPORTED_DATABASE_MIGRATION_STATUS = "Unsupported DatabaseMigration status";
   static final String MESSAGE_NO_MIGRATION_ON_EMBEDDED_DATABASE = "Upgrade is not supported on embedded database.";
   static final String MESSAGE_MIGRATION_REQUIRED = "Database migration is required. DB migration can be started using WS /api/system/migrate_db.";
--- a/server/sonar-server/src/test/java/org/sonar/server/platform/ws/DbMigrationStatusActionTest.java
+++ b/server/sonar-server/src/test/java/org/sonar/server/platform/ws/DbMigrationStatusActionTest.java
@@ -25,7 +25,9 @@ import java.util.Arrays;
 import java.util.Date;
 import javax.annotation.Nullable;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 import org.sonar.api.server.ws.Request;
 import org.sonar.api.utils.DateUtils;
 import org.sonar.db.Database;
@@ -58,6 +60,8 @@ import static org.sonar.server.platform.ws.DbMigrationJsonWriter.STATUS_NO_MIGRA
 import static org.sonar.test.JsonAssert.assertJson;
 
 public class DbMigrationStatusActionTest {
+  @Rule
+  public ExpectedException expectedException = ExpectedException.none();
 
   private static final Date SOME_DATE = new Date();
   private static final String SOME_THROWABLE_MSG = "blablabla pop !";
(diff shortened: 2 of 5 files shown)


server/sonar-server/src/main/java/org/sonar/server/platform/ws/DbMigrationJsonWriter.java
```diff
@@ -37,6 +37,7 @@
   static final String STATUS_MIGRATION_SUCCEEDED = "MIGRATION_SUCCEEDED";
   static final String STATUS_MIGRATION_REQUIRED = "MIGRATION_REQUIRED";
 
+  static final String NO_CONNECTION_TO_DB = "Can not connect to Database.";
```
teryk: Can not –> Cannot even if can not is acceptable

server/sonar-server/src/test/java/org/sonar/server/platform/ws/DbMigrationStatusActionTest.java
```diff
@@ -92,6 +96,16 @@ public void verify_example() throws Exception {
   }
 
   @Test
+  public void throws_ISE_when_databaseVersion_can_not_be_determined() throws Exception {
+    when(databaseVersion.getVersion()).thenReturn(null);
```
teryk: as databaseVersion is a mock, it will return null anyway, right?

server/sonar-server/src/test/java/org/sonar/server/platform/ws/DbMigrationStatusActionTest.java
```diff
@@ -92,6 +96,16 @@ public void verify_example() throws Exception {
...
+
+    expectedException.expect(IllegalStateException.class);
+    expectedException.expectMessage("Can not connect to Database.");
```
teryk: Can not –> Cannot

server/sonar-server/src/test/java/org/sonar/server/platform/ws/MigrateDbActionTest.java
```diff
@@ -80,10 +84,13 @@ public void wireMocksTogether() {
     when(database.getDialect()).thenReturn(dialect);
   }
 
-  @Test(expected = IllegalStateException.class)
+  @Test
   public void ISE_is_thrown_when_version_can_not_be_retrieved_from_database() throws Exception {
     when(databaseVersion.getVersion()).thenReturn(null);
 
+    expectedException.expect(IllegalStateException.class);
+    expectedException.expectMessage("Can not connect to Database.");
```
teryk: Can not –> Cannot

server/sonar-server/src/test/java/org/sonar/server/platform/ws/MigrateDbActionTest.java
```diff
@@ -80,10 +84,13 @@ public void wireMocksTogether() {
     when(database.getDialect()).thenReturn(dialect);
   }
 
-  @Test(expected = IllegalStateException.class)
+  @Test
   public void ISE_is_thrown_when_version_can_not_be_retrieved_from_database() throws Exception {
     when(databaseVersion.getVersion()).thenReturn(null);
```
teryk: same here, if my understanding is correct you don't need to specify that it returns null

teryk:
Apply feedback and LGTM

server/sonar-server/src/main/java/org/sonar/server/platform/ws/DbMigrationJsonWriter.java
sns-seb: @ganncamp what do you think? "Can not" or "Cannot"? I believe the second is less correct English.

server/sonar-server/src/test/java/org/sonar/server/platform/ws/DbMigrationStatusActionTest.java
sns-seb: yes, but I'd rather make it explicit as it is the base of this test

server/sonar-server/src/main/java/org/sonar/server/platform/ws/DbMigrationJsonWriter.java
ganncamp: @sns-seb many would argue that "cannot" is more correct, and indeed the trend seems to be in that direction. I'm going to weigh-in on @teryk 's side here.

server/sonar-server/src/test/java/org/sonar/server/platform/ws/DbMigrationStatusActionTest.java
teryk: ok

server/sonar-server/src/main/java/org/sonar/server/platform/ws/DbMigrationJsonWriter.java
teryk: I checked http://www.oxforddictionaries.com/words/cannot-or-can-not to be sure

sns-seb: ok

sns-seb: FYI, there is roughly 105 usages of "can not" in SQ production code at the moment

teryk: I know there are a lot, I fix them when I see one

PR #585, opened 2015-10-19
SONAR-6582 issues on single-module projects

simonbrandhof:
The WS api/issues/* must not return the module key, but only the project key,
on single-module projects. Indeed UI shows duplicate items in component 
breadcrumb when displaying issues.


julienlancelot:
LGTM

PR #586, opened 2015-10-19
Use "doc values" on some fields of ES issues index

simonbrandhof:
Using doc_values decreases memory pressure in some contexts.


julienlancelot:
LGTM

PR #587, opened 2015-10-19
Feature/rule repository in single db call

sns-seb:
RuleRepository now implements a stonger contract (which is documented)
method hasKey(RuleKey) is replaced by method findByKey(RuleKey) which returns an Optional (saves using two methods in a row when Rule exists)
methods getById(int) and findById(int) added to be able to replace Views' specific RuleRepository with the one from the Compute Engine


server/sonar-server/src/main/java/org/sonar/server/computation/container/ReportComputeEngineContainerPopulator.java
```diff
@@ -140,7 +139,6 @@ private static List componentClasses() {
       ScmInfoRepositoryImpl.class,
 
       // issues
-      RuleCacheLoader.class,
```
julienlancelot: RuleCacheLoader is no more used, you can remove it

server/sonar-server/src/main/java/org/sonar/server/computation/issue/RuleRepository.java
```diff
@@ -19,11 +19,29 @@
  */
 package org.sonar.server.computation.issue;
 
+import com.google.common.base.Optional;
 import org.sonar.api.rule.RuleKey;
 
+/**
+ * Repository of every rule in DB (including manual rules).
```
julienlancelot: You should maybe mention that it's also returning disabled rules

server/sonar-server/src/test/java/org/sonar/server/computation/issue/RuleRepositoryImplTest.java
```diff
@@ -19,37 +19,223 @@
...
+  private static final RuleDto AB_RULE = createABRuleDto();
+  private static final RuleKey AC_RULE_KEY = RuleKey.of("a", "c");
+  private static final int AC_RULE_ID = 684;
+
+  @org.junit.Rule
+  public ExpectedException expectedException = ExpectedException.none();
+
+  private DbClient dbClient = mock(DbClient.class);
+  private DbSession dbSession = mock(DbSession.class);
+  private RuleDao ruleDao = mock(RuleDao.class);
```
julienlancelot: Why did you not used DbTester ?

julienlancelot: I imagine it's because you wanted to test that the call to the db is only do once.
To do that with dbTester, you can for instance to a first call to load all rules from db, then remove them from the db and check that you can still get some rules.

julienlancelot:
Somme comments and LGTM

julienlancelot:
Do not forget to update Views in order to use the RuleRepository provided by SQ and to remove the one from Views

server/sonar-server/src/test/java/org/sonar/server/computation/issue/RuleRepositoryImplTest.java
sns-seb: ...which is time and CPU consuming.

DBTester is appropriate when testing DAOs, complex interactions with the DB and/or DB vendor sensitive code. Otherwise, it creates slower tests for low benefits.

julienlancelot: I understand, but here you're very coupled with the db...

sns-seb: the class is coupled to a DAO, not to the DB. It's the point of a DAO, to abstract away the DB. And interaction with the DAO is simple and read-only.

julienlancelot: But it could be interesting for instance to validate in this test that disabled rules are well returned.

sns-seb:
merged into branch "branch-5.2"

PR #588, opened 2015-10-20
SONAR-6822 langName is not returned

julienlancelot:
/api/rules/search?f=langName is now returning langName
/api/rules/search?f=lang doest not return anymore langName


server/sonar-server/src/main/java/org/sonar/server/rule/ws/RuleMapping.java
```diff
@@ -317,6 +318,11 @@ private static void setInternalKey(Rules.Rule.Builder ruleResponse, Rule ruleDoc
   private void setLanguage(Rules.Rule.Builder ruleResponse, Rule ruleDoc, Set<String> fieldsToReturn) {
```
simonbrandhof: Can be static.

server/sonar-server/src/main/java/org/sonar/server/rule/ws/RuleMapping.java
```diff
@@ -317,6 +318,11 @@ private static void setInternalKey(Rules.Rule.Builder ruleResponse, Rule ruleDoc
...
     if (shouldReturnField(fieldsToReturn, RuleNormalizer.RuleField.LANGUAGE) && ruleDoc.language() != null) {
       ruleResponse.setLang(ruleDoc.language());
+    }
+  }
+
+  private void setLanguageName(Rules.Rule.Builder ruleResponse, Rule ruleDoc, Set<String> fieldsToReturn) {
+    if (shouldReturnField(fieldsToReturn, "langName") && ruleDoc.language() != null) {
       Language language = languages.get(ruleDoc.language());
       if (language != null) {
         ruleResponse.setLangName(language.getName());
```
simonbrandhof: I suggest to return the lang key when the name is not available.

simonbrandhof:
Two feedbacks then LGTM

server/sonar-server/src/main/java/org/sonar/server/rule/ws/RuleMapping.java
julienlancelot: Right !

server/sonar-server/src/main/java/org/sonar/server/rule/ws/RuleMapping.java
julienlancelot: It will change existing behaviour but I think it's a good idea.

simonbrandhof: of course it's a good idea :-)

PR #589, opened 2015-10-20
Hack for IE - use mime type text/plain for api/qualityprofiles/create

simonbrandhof:
See comment in CreationAction#guessMediaType()


julienlancelot:
LGTM

PR #590, opened 2015-10-20
SONAR-6318 Not fail on bad encoding in ComputeEngine

julienlancelot:
When a source is badly encoded, the Compute Engine will no more fail but will ignore remaining highlighting / symbols and will display a warning log.


julienlancelot:
Wrong branch...

PR #592, opened 2015-10-20
Improve api/tests WS to follow conventions

simonbrandhof:


julienlancelot:
Be careful as jobs are failing otherwise LGTM

PR #595, opened 2015-10-21
SONAR-6467 When updating a user, the default group should not be set

julienlancelot:


sns-seb:
no sebastienade on this one, way to small :)

LGTM

PR #599, opened 2015-10-26
SONAR-5794 Send notification step should not use batch repo to get project name

julienlancelot:
SONAR-5794 Send notification step should not use batch repo to get project name


simonbrandhof:
LGTM

PR #600, opened 2015-10-27
SONAR-6414 Indexing tests should update updatedAt field

julienlancelot:
SONAR-6414 Indexing tests should update updatedAt field


simonbrandhof:
LGTM

PR #603, opened 2015-10-28
SONAR-6729 CE not fail anymore when no quality profile found

julienlancelot:
SONAR-6729 CE not fail anymore when no quality profile found


simonbrandhof:
LGTM

PR #604, opened 2015-10-28
SONAR-6356 Use first analysis when no previous version exist

julienlancelot:


sonar-db/src/main/java/org/sonar/core/timemachine/Periods.java
```diff
@@ -40,88 +47,107 @@ public Periods(Settings settings, I18n i18n) {
...
+    }
+    return label("since_previous_version_detailed", shortLabel, param, date);
+  }
+
+  private String label(String key, boolean shortLabel, Object... parameters) {
+    String msgKey = key;
+    if (shortLabel) {
+      msgKey += ".short";
+    }
+    return i18n.message(getLocale(), msgKey, null, parameters);
```
simonbrandhof: What about inlining `Local.ENGLISH`  instead of using a private method for a single call ?

julienlancelot: indeed, good point

sonar-db/src/main/java/org/sonar/db/component/SnapshotDao.java
```diff
@@ -69,6 +69,12 @@ public boolean hasLastSnapshotByComponentUuid(DbSession session, String componen
     return mapper(session).selectPreviousVersionSnapshots(componentId, lastVersion);
   }
 
+  @CheckForNull
+  public SnapshotDto selectFirstSnapshot(DbSession session, long componentId) {
+    List<SnapshotDto> snapshotDtos = mapper(session).selectSnapshotsOrderedByAscCreatedAt(componentId);
```
simonbrandhof: Filtering the first row must be done in DB to avoid loading of huge history of snapshots. See RowBounds of MyBatis.

julienlancelot: Great ! I thought the other way was to do specific queries... I'll do this.

sonar-db/src/main/resources/org/sonar/db/component/SnapshotMapper.xml
```diff
@@ -118,6 +118,14 @@
     ORDER BY e.event_date DESC
   </select>
 
+  <select id="selectSnapshotsOrderedByAscCreatedAt" parameterType="map" resultType="Snapshot">
+    SELECT
+    <include refid="snapshotColumns"/>
+    FROM snapshots s
+    INNER JOIN projects p ON p.id=s.project_id AND p.id=#{componentId}
```
simonbrandhof: why not directly `s.project_id=#{componentId}` ? To be sure that the row exists in PROJECTS ?

sonar-db/src/main/java/org/sonar/db/component/SnapshotDao.java
```diff
@@ -69,6 +69,12 @@ public boolean hasLastSnapshotByComponentUuid(DbSession session, String componen
     return mapper(session).selectPreviousVersionSnapshots(componentId, lastVersion);
   }
 
+  @CheckForNull
+  public SnapshotDto selectFirstSnapshot(DbSession session, long componentId) {
```
simonbrandhof: To be renamed `selectOldestSnapshot` ?

sonar-db/src/main/resources/org/sonar/db/component/SnapshotMapper.xml
julienlancelot: Hum... No reason... A copy/paste from selectPreviousVersionSnapshots, which should then also be updated.

sonar-db/src/main/java/org/sonar/db/component/SnapshotDao.java
julienlancelot: +1

julienlancelot:
Feedbacks applied @simonbrandhof

simonbrandhof:
LGTM

PR #608, opened 2015-10-29
SONAR-5851 Add back the /profiles/export Web Page

teryk:


simonbrandhof:
WS is missing not documentation. Moreover it should be described as deprecated.

teryk:
It was not a WS, it was a web page which returned data, that's why it's not listed in the list of WS. I was quite surprised.

simonbrandhof:
LGTM

PR #629, opened 2015-11-13
SONAR-5676 Improve validation of highlighting API

dbmeneses:


it/it-tests/src/test/java/it/test/HighlightingTest.java
```diff
@@ -0,0 +1,25 @@
...
+import static util.ItUtils.projectDir;
+import com.sonar.orchestrator.Orchestrator;
+import it.Category2Suite;
+import org.junit.ClassRule;
+
+public class HighlightingTest {
+  @ClassRule
+  public static Orchestrator orchestrator = Category2Suite.ORCHESTRATOR;
+  
+  @Test
```
henryju: Do we really need an IT for that? Was medium test not enough?

henryju:
LGTM except that I would remove the IT

PR #633, opened 2015-11-13
SONAR-5894 Support symbol references with different length

dbmeneses:


sonar-plugin-api/src/main/java/org/sonar/api/source/Symbolizable.java
```diff
@@ -36,8 +36,11 @@
     Symbol newSymbol(int fromOffset, int toOffset);
 
     void newReference(Symbol symbol, int fromOffset);
```
henryju: Maybe take the opportunity to improve Javadoc. Here you can explain that reference is supposed to have same lenght than symbol declaration.

henryju:
LGTM

PR #643, opened 2015-11-20
add AbstractComputationSteps

sns-seb:
reduce duplication of code logic in implementations of ComputationSteps accross SQ and CE tasks in plugins


julienlancelot:
LGTM

PR #654, opened 2015-11-26
WS user_tokens/revoke and WS user_tokens/search

teryk:


server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/RevokeAction.java
```diff
@@ -0,0 +1,93 @@
...
+  private final UserSession userSession;
+
+  public RevokeAction(DbClient dbClient, UserSession userSession) {
+    this.dbClient = dbClient;
+    this.userSession = userSession;
+  }
+
+  @Override
+  public void define(WebService.NewController context) {
+    WebService.NewAction action = context.createAction(ACTION_REVOKE)
```
simonbrandhof: + "Requires permission to administrate system" ?

server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/RevokeAction.java
```diff
@@ -0,0 +1,93 @@
...
+  private void doHandle(RevokeWsRequest request) {
+    userSession.checkLoggedIn().checkGlobalPermission(SYSTEM_ADMIN);
+
+    DbSession dbSession = dbClient.openSession(false);
+    try {
+      UserTokenDto userTokenDto = checkFoundWithOptional(
+        dbClient.userTokenDao().selectByLoginAndName(dbSession, request.getLogin(), request.getName()),
+        "User token with login '%s' and name '%s' not found", request.getLogin(), request.getName());
+
+      dbClient.userTokenDao().deleteByTokenHash(dbSession, userTokenDto.getTokenHash());
```
simonbrandhof: It's better to directly delete by login and name, whatever the number of results. No need to load token.

server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/SearchAction.java
```diff
@@ -0,0 +1,101 @@
...
+import org.sonar.server.user.UserSession;
+import org.sonarqube.ws.WsUserTokens.SearchWsResponse;
+import org.sonarqube.ws.client.usertoken.SearchWsRequest;
+
+import static org.sonar.api.utils.DateUtils.formatDateTime;
+import static org.sonar.server.ws.WsUtils.writeProtobuf;
+import static org.sonarqube.ws.client.usertoken.UserTokensWsParameters.ACTION_SEARCH;
+import static org.sonarqube.ws.client.usertoken.UserTokensWsParameters.PARAM_LOGIN;
+
+public class SearchAction implements UserTokensWsAction {
```
simonbrandhof: Here it's not search, but "list" ?

server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/SearchAction.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public SearchAction(DbClient dbClient, UserSession userSession) {
+    this.dbClient = dbClient;
+    this.userSession = userSession;
+  }
+
+  @Override
+  public void define(WebService.NewController context) {
+    WebService.NewAction action = context.createAction(ACTION_SEARCH)
+      .setDescription("Search a user access tokens")
```
simonbrandhof: "List the access tokens of a user" ?

sns-seb: "List**s**", right?

server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/SearchAction.java
```diff
@@ -0,0 +1,101 @@
...
+    SearchWsResponse searchWsResponse = doHandle(toSearchWsRequest(request));
+    writeProtobuf(searchWsResponse, request, response);
+  }
+
+  private SearchWsResponse doHandle(SearchWsRequest request) {
+    userSession.checkLoggedIn().checkGlobalPermission(GlobalPermissions.SYSTEM_ADMIN);
+
+    DbSession dbSession = dbClient.openSession(false);
+    try {
+      List<UserTokenDto> userTokens = dbClient.userTokenDao().selectByLogin(dbSession, request.getLogin());
```
simonbrandhof: Don't you think that the WS should throw a bad request error if the login does not exist ?

server/sonar-server/src/main/resources/org/sonar/server/usertoken/ws/search-example.json
```diff
@@ -0,0 +1,19 @@
+{
+  "userTokens": [
+    {
+      "login": "grace.hopper",
```
simonbrandhof: it's not/it won't be possible to return tokens on different users, so repeating the login is misleading. I would move the field "login" to the root node.

server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/RevokeAction.java
teryk: yes, true for all the new WS

server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/RevokeAction.java
teryk: I don't agree. It should match one result exactly. If the token is not found, there's a problem. The couple login+name is unique

server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/SearchAction.java
teryk: well, we use search for most of the WS, even when list could have been appropriate. The only case I can think of is api/rules/list which lists ALL the rules, without any parameter. Apart from that we decided (or so I thought) to use search even if list could make sense.

server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/SearchAction.java
teryk: @sns-seb nope list

server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/SearchAction.java
teryk: ok

server/sonar-server/src/main/resources/org/sonar/server/usertoken/ws/search-example.json
teryk: ok

server/sonar-server/src/main/java/org/sonar/server/usertoken/ws/RevokeAction.java
simonbrandhof: The point is not about the case if more than 1 result, but what happens if the requested token does not exist. I think it should be success, not 404.

teryk: ok

simonbrandhof:
LGTM

PR #657, opened 2015-11-27
SONAR-6892 WS permissions/search_project_permissions filter by qualifier

teryk:


server/sonar-server/src/main/java/org/sonar/server/permission/ws/PermissionsWsParametersBuilder.java
```diff
@@ -100,33 +120,90 @@ public static void createTemplateParameters(NewAction action) {
...
-    action.createParam(org.sonarqube.ws.client.permission.PermissionsWsParameters.PARAM_ID)
+    action.createParam(PARAM_ID)
       .setRequired(true)
       .setDescription("Id")
       .setExampleValue("af8cb8cc-1e78-4c4e-8c00-ee8e814009a5");
   }
+
+  public static NewParam createQualifierParameter(NewAction action, QualifierParameterContext context) {
+    return action.createParam(PARAM_QUALIFIER)
+      .setDescription("Project qualifier. Possible values are:" + buildRootQualifiersDescription(context))
```
simonbrandhof: Missing since 5.3

simonbrandhof: Description could be improved with something like "Filters the results with the specified qualifier" or "Indicates which type of projects to return".
Maybe a reference to Views and Cockpit plugin would be useful too.

server/sonar-server/src/main/java/org/sonar/server/permission/ws/PermissionsWsParametersBuilder.java
```diff
@@ -100,33 +120,90 @@ public static void createTemplateParameters(NewAction action) {
...
+    String qualifierPattern = "<li>%s - %s</li>";
+    for (String qualifier : getRootQualifiers(context.getResourceTypes())) {
+      description.append(format(qualifierPattern, qualifier, i18n(context, qualifier)));
+    }
+    description.append("</ul>");
+
+    return description.toString();
+  }
+
+  private static String i18n(QualifierParameterContext context, String qualifier) {
```
simonbrandhof: Method name `i18n` is not intuitive out of context. What about `qualifierLabel` ?

simonbrandhof:
Added some comments then LGTM

PR #664, opened 2015-12-01
fix CE processing starting before Views or Devcockpit at startup

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/computation/queue/CeQueueInitializer.java
```diff
@@ -47,7 +49,14 @@ public CeQueueInitializer(DbClient dbClient, CEQueueStatus queueStatus, CeQueueC
   }
 
   @Override
-  public void start() {
+  public void onServerStart(Server server) {
+    if (!done) {
```
julienlancelot: For my information, why do you need to store this done state ? Can the onServerStart() be called many times on the same ServerStartHandler ?

julienlancelot: Ok, I've got my answer below, with the UT you've added.

julienlancelot:
LGTM

server/sonar-server/src/main/java/org/sonar/server/computation/queue/CeQueueInitializer.java
sns-seb: this is code for the sole purpose of testing (in medium tests). I think it's crapy but I couldn't find an alternative

julienlancelot: Indeed, we'll maybe need to update MT later.

PR #665, opened 2015-12-02
SONAR-7030 Drop most default hardcoded built-in dashboards

stas-vilchik:


julienlancelot:
Well done Stas ! You could join us to do some Java code !
LGTM

PR #668, opened 2015-12-02
SONAR-7100 Disable WS cache by default in issues mode

henryju:


dbmeneses:
LGTM
I'm quite surprised that no IT got broken!

PR #672, opened 2015-12-03
SONAR-6995 CEQueueStatus get method should not modify state

sns-seb:
and they should not required the CEQueueStatusImpl object to be initialized either
this way, monitoring can query the metrics of CEQueueStatus through JMX even before the object is initialized


julienlancelot:
LGTM !

PR #677, opened 2015-12-04
SONAR-7098 Always feed issues authors

julienlancelot:
SONAR-7098 Always feed issues authors


server/sonar-server/src/main/java/org/sonar/server/computation/issue/IssueAssigner.java
```diff
@@ -59,24 +59,20 @@ public IssueAssigner(ScmInfoRepository scmInfoRepository, ScmAccountToUser scmAc
 
   @Override
   public void onIssue(Component component, DefaultIssue issue) {
-    if (issue.isNew()) {
```
simonbrandhof: This is not correct. The algorithm must be:
1. if author is null
   1.1. load scm
   1.2. guess author
2. if assignee is null
    2.1. if author just set during step 1.2, guess assignee. Else that means that issue was manually unassigned by end-users and "unassigned" status must not be changed

julienlancelot:
@simonbrandhof Thanks for your feedback, I've applied it, can you have a look please ?

sonar-core/src/main/java/org/sonar/core/issue/IssueUpdater.java
```diff
@@ -22,7 +22,9 @@
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
 import com.google.common.base.Objects;
+import com.google.common.base.Preconditions;
```
SonarTech: MINOR Remove this unused import 'com.google.common.base.Preconditions'. [rule](https://dory.sonarsource.com/coding_rules#rule_key=squid%3AUselessImportCheck)

sonar-core/src/main/java/org/sonar/core/issue/IssueUpdater.java
```diff
@@ -22,7 +22,9 @@
...
 import com.google.common.base.Predicate;
+import com.google.common.base.Strings;
```
SonarTech: MINOR Remove this unused import 'com.google.common.base.Strings'. [rule](https://dory.sonarsource.com/coding_rules#rule_key=squid%3AUselessImportCheck)

SonarTech:
SonarQube analysis reported 2 issues:
* MINOR 2 minor

SonarTech:
SonarQube analysis reported 2 issues:
* MINOR 2 minor

server/sonar-server/src/main/java/org/sonar/server/computation/issue/IssueAssigner.java
```diff
@@ -46,37 +53,38 @@
...
-      issue.setAuthorLogin(scmAuthor);
-      if (scmAuthor != null) {
-        String assigneeLogin = scmAccountToUser.getNullable(scmAuthor);
-        if (assigneeLogin == null) {
-          issue.setAssignee(defaultAssigne.getLogin());
-        } else {
-          issue.setAssignee(assigneeLogin);
-        }
-      }
+      issueUpdater.setNewAuthor(issue, scmAuthor, changeContext);
```
simonbrandhof: I think `scmAuthor` can be null if there's no default assignee. In this case the issue does not need to be updated and `authorHasBeenSet` should be false.

simonbrandhof: why "hasBeen" but not "was" ?

julienlancelot: you're right scmAuthor can be null, and in this case issueUpdater. setNewAuthor won't do anything

julienlancelot: was is also good :)

PR #682, opened 2015-12-04
SONAR-7088 fix RejectedExecutionException at SQ shutdown

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/computation/taskprocessor/CeProcessingSchedulerImpl.java
```diff
@@ -48,39 +54,74 @@ public CeProcessingSchedulerImpl(CeProcessingSchedulerExecutorService processing
...
+    this.chainingCallback.stop();
+  }
+
   private class ChainingCallback implements FutureCallback<Boolean> {
+    private final AtomicBoolean keepRunning = new AtomicBoolean(true);
+    @CheckForNull
+    private ListenableFuture<Boolean> workerFuture;
+
     @Override
     public void onSuccess(@Nullable Boolean result) {
```
simonbrandhof: According to `CeWorkerRunnableImpl#call()`, the boolean result can never be null ?

simonbrandhof:
LGTM

server/sonar-server/src/main/java/org/sonar/server/computation/taskprocessor/CeProcessingSchedulerImpl.java
sns-seb: a non nullable `Boolean` would be a `boolean` but it can not be because of generics, so I stick with the signatore defined by `FutureCallBack` and keep the `@Nullable`

PR #684, opened 2015-12-07
SONAR-6820 WS has page size limit of 500

teryk:


sns-seb:
@teryk PR back to you

PR #698, opened 2015-12-17
SONAR-7129 WS api/components/tree

teryk:


SonarTech:
* MAJOR 1 major

* MAJOR [Protected method 'configureModule(...)' is never used.](https://github.com/SonarSource/sonarqube/blob/d1e69e71db559b439cb453b687039deca2ef76ab/server/sonar-server/src/main/java/org/sonar/server/component/ws/ComponentsWsModule.java#L27) [rule](https://dory.sonarsource.com/coding_rules#rule_key=squid%3AUnusedProtectedMethod)

simonbrandhof:
The parameter `qualifiers` for filtering results is missing.

simonbrandhof:
When no properties are set, then the error message is `The component key or the component id must be provided, not both.`. The expression `not both` is misleading. I suggest `One of the component id or key must be set`.

simonbrandhof:
The field `createdAt` can be removed from response IMO.

simonbrandhof:
The field `projectId` has the same value for all components, correct ? In that case I suggest to move it outside the field `components`.

simonbrandhof:
Sorting by qualifier should be supported. It can make sense if the strategy is `all`. Moreover the parameter `s` should support a list of values (it's maybe already the case, but it's not documented), for instance `qualifier,path`

simonbrandhof:
I know that I suggested the term `strategy`... but in fact I think `nodes` or `type` would be more appropriate (preference for `nodes` as we're in the concept of tree).

simonbrandhof:
Documentation: since 5.4, but not 5.3.

simonbrandhof:
Why does it need "'Browse' AND 'See Source Code' on the specified project" ? Why not "Browse" only ?

simonbrandhof:
Open question : should the field baseComponent be returned ?

simonbrandhof:
The field `parentId` should be returned in order to allow to rebuild tree hierarchy (for instance from a UI gadget).

server/sonar-server/src/main/java/org/sonar/server/component/ws/TreeAction.java
```diff
@@ -0,0 +1,264 @@
...
+      List<ComponentDto> components;
+      int total;
+      switch (treeWsRequest.getStrategy()) {
+        case CHILDREN_STRATEGY:
+          components = dbClient.componentDao().selectDirectChildren(dbSession, query);
+          total = dbClient.componentDao().countDirectChildren(dbSession, query);
+          break;
+        case LEAVES_STRATEGY:
+        case ALL_STRATEGY:
+        default:
```
simonbrandhof: `default` should throw an IllegalStateException to avoid bad results the day a new constant value is added.

server/sonar-server/src/main/java/org/sonar/server/component/ws/TreeAction.java
```diff
@@ -0,0 +1,264 @@
...
+      .setQualifier(dto.qualifier())
+      .setCreatedAt(DateUtils.formatDateTime(dto.getCreatedAt()));
+    if (dto.path() != null) {
+      wsComponent.setPath(dto.path());
+    }
+    if (dto.projectUuid() != null) {
+      wsComponent.setProjectId(dto.projectUuid());
+    }
+    if (dto.description() != null) {
+      wsComponent.setDescription(dto.description());
```
simonbrandhof: I didn't see this field in example response.

sns-seb:
> I suggest `One of the component id or key must be set`.

What about `One, and only one, of the component id or key must be set`?

simonbrandhof:
Views are not supported, because of the SQL condition `p.copy_resource_id is null`

SonarTech:
* MAJOR 1 major

* MAJOR [Protected method 'configureModule(...)' is never used.](https://github.com/SonarSource/sonarqube/blob/fd2555f800c276c41254984dba7d5b8c876b170c/server/sonar-server/src/main/java/org/sonar/server/component/ws/ComponentsWsModule.java#L27) [rule](https://dory.sonarsource.com/coding_rules#rule_key=squid%3AUnusedProtectedMethod)

server/sonar-server/src/main/java/org/sonar/server/component/ws/TreeAction.java
teryk: done

server/sonar-server/src/main/java/org/sonar/server/component/ws/TreeAction.java
teryk: added

SonarTech:
* MAJOR 1 major

* MAJOR [Protected method 'configureModule(...)' is never used.](https://github.com/SonarSource/sonarqube/blob/459cfa2b3a11abdcd825a95272cef6373a22ca31/server/sonar-server/src/main/java/org/sonar/server/component/ws/ComponentsWsModule.java#L27) [rule](https://dory.sonarsource.com/coding_rules#rule_key=squid%3AUnusedProtectedMethod)

PR #701, opened 2016-01-02
Fix Map.containsKey call that would never return true

Shredder121:


sns-seb:
Hello @Shredder121 

Thank you for your contribution and well spotted.

I've merged your commit manually into master.

Shredder121:
No problem!
Keep up the good work!

PR #713, opened 2016-01-13
SONAR-7197 Drop Solaris scripts from distribution

simonbrandhof:
SONAR-7197 Drop Solaris scripts from distribution


sns-seb:
LGTM

PR #719, opened 2016-01-16
Replace hardcoded string in AddedRemovedDebt component

felipebz:
HI guys.

I found these two non-localizable strings in the SonarQube UI. It's a trivial change.


stas-vilchik:
Thanks for the contribution!

PR #745, opened 2016-02-01
SONAR-7020 forbid login with a space when creating a user

teryk:


julienlancelot:
There's no db migration, but it would be a pain to add one a a lot of tables are using the login as a reference to the user.
Moreover, I see no impact to have some logins with space and some without (as there's no more explanation in the JIRA ticket ;) )

julienlancelot:
@teryk, as there will be no migration, please add some comments in the code that before SonarQube 5.4, spaces were allowed in login (you can add comments for instance in UserDtop.getLogin()).
Then LGTM !

PR #746, opened 2016-02-01
SONAR-7257 make SymbolsLineReader instantaneous

sns-seb:
fix PersistFilesSources taking forever

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/source/SymbolsLineReader.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/source/SymbolsLineReader.java
@@ -19,17 +19,16 @@
  */
 package org.sonar.server.computation.source;
 
+import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Lists;
-import java.io.Serializable;
+import com.google.common.collect.SetMultimap;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import org.sonar.api.utils.log.Logger;
 import org.sonar.api.utils.log.Loggers;
 import org.sonar.batch.protocol.output.BatchReport;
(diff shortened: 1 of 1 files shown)


server/sonar-server/src/main/java/org/sonar/server/computation/source/SymbolsLineReader.java
```diff
@@ -77,19 +77,23 @@ public void read(DbFileSources.Line.Builder lineBuilder) {
 
   private void processSymbols(DbFileSources.Line.Builder lineBuilder) {
     int line = lineBuilder.getLine();
-    List<BatchReport.Symbol> lineSymbols = findSymbolsMatchingLine(line);
+
+    List<BatchReport.Symbol> lineSymbols = new ArrayList<>(this.symbolsPerLine.get(line));
+    // Sort symbols to have deterministic results and avoid false variation that would lead to an unnecessary update of the source files
+    // data
+    Collections.sort(lineSymbols, SymbolsComparator.INSTANCE);
```
julienlancelot: If you replace the SetMultimap by a SortedSetMultimap, I think you could avoid this sort, or am I wrong ?

server/sonar-server/src/main/java/org/sonar/server/computation/source/SymbolsLineReader.java
```diff
@@ -140,7 +125,26 @@ private static boolean matchLine(BatchReport.TextRange range, int line) {
...
+    return res;
+  }
+
+  private static void putForTextRange(SetMultimap<Integer, BatchReport.Symbol> res, BatchReport.Symbol symbol, BatchReport.TextRange declaration) {
+    for (int i = declaration.getStartLine(); i <= declaration.getEndLine(); i++) {
+      res.put(i, symbol);
+    }
+  }
+
   private static class SymbolsComparator implements Comparator<BatchReport.Symbol>, Serializable {
```
julienlancelot: you can replace this class by an enum

server/sonar-server/src/main/java/org/sonar/server/computation/source/SymbolsLineReader.java
julienlancelot: Or more probably a ArrayListMultimap

sns-seb: I need a HashSetMultimap to remove duplicate symbols per line based on their hash
a SortedSetMultimap will use the comparator to both sort and remove duplicates, and it will use it for each insert. The current comparator implementation is not strong enough to be used for comparison.
in addition, hash method of protobuf objects is memoized for better performance
I'd rather sort once and at the _only_ time the list of symbols of a line is used

server/sonar-server/src/main/java/org/sonar/server/computation/source/SymbolsLineReader.java
sns-seb: true

server/sonar-server/src/main/java/org/sonar/server/computation/source/SymbolsLineReader.java
julienlancelot: ok, thanks for the explanation

julienlancelot:
LGTM

PR #749, opened 2016-02-01
add AuthorDao.selectAll

sns-seb:
added for Developer Cockpit to implement a much faster `DeveloperRepository`

### DIFF
--- a/sonar-db/src/main/java/org/sonar/db/user/AuthorDao.java
+++ b/sonar-db/src/main/java/org/sonar/db/user/AuthorDao.java
@@ -37,6 +37,10 @@ public class AuthorDao implements Dao {
     return getMapper(session).selectByLogin(login);
   }
 
+  public List<AuthorDto> selectAll(DbSession session) {
+    return getMapper(session).selectAll();
+  }
+
   public int countDeveloperLogins(DbSession session, long developerId) {
     return getMapper(session).countDeveloperLogins(developerId);
   }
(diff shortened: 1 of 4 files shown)


julienlancelot:
Isn't it dangerous to all all developer in memory ? Wouldn't it be possible to do some streaming ?

sns-seb:
we are loading rows of table AUTHORS. I don't think that will represent a lot of rows

julienlancelot:
The number of authors is the number of commiters, it can be up to something to 100.000 maybe

sns-seb:
looking at how the result of `selectAll` is used, I don't see any benefits from streaming data from table AUTHORS. We want all the perso ids to that we can then read table PROJECTS with the least possible queries.

julienlancelot:
So why not stream and only keep in memory what you need ?

sns-seb:
I need 2 fields out or 5 from `AuthorDto`.
So, I'll need to create another type to hold these 2 fields and add code complexity to use a `ResultHandler`
I'm really not sure of the maintenance and memory benefit

SonarTech:
* MAJOR 1 major

* MAJOR Add a 'package-info.java' file to document the 'protobuf' package (org.sonarsource.sonarqube:sonar-db:target/generated-sources/protobuf/org/sonar/db/protobuf) [rule](https://dory.sonarsource.com/coding_rules#rule_key=squid%3AS1228)

julienlancelot:
@sns-seb Let's talk about it with @simonbrandhof

sns-seb:
changes introduced by this PR dropped, closing

PR #753, opened 2016-02-02
SONAR-6584 WS api/resources/index limit number of measures and components

teryk:


server/sonar-server/src/main/java/org/sonar/server/component/ws/ResourcesWs.java
```diff
@@ -44,11 +44,14 @@ private void defineIndexAction(NewController controller) {
...
-        "<li>if you need components only: api/components/tree</li>" +
-        "<li>if you need components with measures: api/measures/component_tree</li>" +
+        "<li>if you need one component without measures: api/components/show</li>" +
+        "<li>if you need one component with measures: api/measures/component</li>" +
+        "<li>if you need several components without measures: api/components/tree</li>" +
+        "<li>if you need several components with measures: api/measures/component_tree</li>" +
         "</ul>" +
-        "When you provide one metric, the number of results is limited to 500. When several metrics are provided, the number of results is not limited. " +
-        "This is a known limitation and it won't be fixed.")
+        "When you provide one metric, the number of results is limited to 500. When several metrics are provided, the number of measures is limited to 10000. " +
```
simonbrandhof: As components are limited to 500, limiting measures is not required, especially because 10'000 means 200 measures per component, which is very high.

teryk: the problem is that measures are loaded BEFORE the components. I took an arbitrary high number because I want to be have all the measures of the 500 components in most of the cases. Not ideal, but good enough IMO as replacements exist and avoid the OOM problem

PR #755, opened 2016-02-02
SONAR-6987 WS api/tests/list search tests by source file key

teryk:


server/sonar-server/src/main/java/org/sonar/server/test/ws/ListAction.java
```diff
@@ -105,12 +108,18 @@ public void define(WebService.NewController controller) {
 
     action
       .createParam(SOURCE_FILE_ID)
-      .setDescription("IF of source file. Must be provided with the source file line number.")
+      .setDescription("ID of source file. Must be provided with the source file line number.")
```
julienlancelot: :)

server/sonar-server/src/main/java/org/sonar/server/test/ws/ListAction.java
```diff
@@ -185,15 +194,8 @@ private static int coveredLines(List<CoveredFileDoc> coveredFiles) {
...
   }
 
-  private static class TestToFileUuidFunction implements Function<TestDoc, String> {
-    @Override
-    public String apply(@Nonnull TestDoc testDoc) {
-      return testDoc.fileUuid();
-    }
-  }
-
   private SearchResult<TestDoc> searchTests(DbSession dbSession, @Nullable String testUuid, @Nullable String testFileUuid, @Nullable String testFileKey,
```
julienlancelot: Lot of parameters...

julienlancelot: It would be interesting to creating a class like a TestsFields that would be created from the request, and that would return all you need : testFileUuid, testFileKey, sourceFileUuid, etc.
This class would also have the responsability to check arguments.

server/sonar-server/src/test/java/org/sonar/server/test/ws/ListActionTest.java
```diff
@@ -192,6 +192,59 @@ public void list_based_on_main_file_and_line_number() throws Exception {
     request.execute().assertJson(getClass(), "list-main-file.json");
   }
 
+  @Test
+  public void list_based_on_source_file_key_and_line_number() throws Exception {
+    String sourceFileUuid = "MAIN-FILE-UUID";
+    String sourceFileKey = "MAIN-FILE-KEY";
+    userSessionRule.addProjectUuidPermissions(UserRole.CODEVIEWER, TestFile1.PROJECT_UUID);
+    dbClient.componentDao().insert(db.getSession(),
+      new ComponentDto()
```
julienlancelot: All these components creation could be factorised -> FILE_1, FILE_2, or something like that.

julienlancelot:
Possible refactoring to do, and LGTM.

PR #756, opened 2016-02-03
SONAR-7266 Project admin can manage project permissions

teryk:


server/sonar-server/src/main/java/org/sonar/server/permission/ws/template/ApplyTemplateAction.java
```diff
@@ -64,7 +64,7 @@ public void define(WebService.NewController context) {
       .setHandler(this);
 
     createTemplateParameters(action);
-    createProjectParameter(action);
+    createProjectParameters(action);
```
julienlancelot: It would have been nice to have this renaming done in another commit, as it's a "side" modification.
Moving it in another commit makes commits clearer

server/sonar-server/src/test/java/org/sonar/server/permission/ws/template/SearchTemplatesActionTest.java
```diff
@@ -175,14 +172,6 @@ public void fail_if_not_logged_in() {
   }
 
   @Test
-  public void fail_if_not_global_admin() {
```
julienlancelot: Why have you removed this test ? It should just be replaced by a test that check that the user needs to be logged.

julienlancelot:
Description of the WS api/permissions/search_templates must be updated to explain that the user needs to be logged to access the WS

julienlancelot:
Required : Please update the WS description and add the missing UT.
Nice to have : split your modification into 2 commits.
Then LGTM !

server/sonar-server/src/test/java/org/sonar/server/permission/ws/template/SearchTemplatesActionTest.java
teryk: because there is another test that tests just that

julienlancelot: Oups, indeed sorry

PR #757, opened 2016-02-03
SONAR-6226 Apply various feedback

julienlancelot:


julienlancelot:
Associated PR on GitHub plugin : https://github.com/SonarSource/sonar-auth-github/pull/3

server/sonar-web/src/main/webapp/WEB-INF/app/views/sessions/_form.html.erb
```diff
@@ -7,9 +7,10 @@
       <% auth_providers.each do |provider| %>
         <li>
           <a href="<%= ApplicationController.root_context -%>/sessions/init/<%= provider.getKey().to_s %>"
```
simonbrandhof: No need for <%= ApplicationController.root_context -%> anymore (same for line 13)

sonar-plugin-api/src/main/java/org/sonar/api/server/authentication/Display.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static class Builder {
+
+    private String iconPath;
+    private String backgroundColor;
```
simonbrandhof: It's not always appropriate to define a color. I suggest to use a default color that can be overridden.

sonar-plugin-api/src/main/java/org/sonar/api/server/authentication/Display.java
```diff
@@ -0,0 +1,103 @@
...
+  private final String backgroundColor;
+
+  private Display(Builder builder) {
+    this.iconPath = builder.iconPath;
+    this.backgroundColor = builder.backgroundColor;
+  }
+
+  /**
+   * URL path to the provider icon, as deployed at runtime, for example "/static/authgithub/github.svg" (in this
+   * case "authgithub" is the plugin key. Source file is "src/main/resources/static/github.svg"). Must not be blank.
```
simonbrandhof: "Must not be blank and must start with slash" ?

server/sonar-web/src/main/webapp/WEB-INF/app/views/sessions/_form.html.erb
julienlancelot: Indeed !

sonar-plugin-api/src/main/java/org/sonar/api/server/authentication/Display.java
julienlancelot: I was asking myself the same question. So default value should be black ?

sonar-plugin-api/src/main/java/org/sonar/api/server/authentication/Display.java
julienlancelot: Ok

PR #759, opened 2016-02-04
SONAR-7169 Remove fallback from AverageFormula

julienlancelot:
SONAR-7169 Remove fallback from AverageFormula


sns-seb:
LGTM but PR build on travis is not yet green

PR #760, opened 2016-02-04
SONAR-6858 Do not write into webapp directory at runtime

simonbrandhof:


julienlancelot:
When settings the whole direction of SonarQube in read only, and settings the properties sonar.path.logs, sonar.path.data, sonar.path.temp on external directories, here's what happen : 
- Error when writing the PID : 
```
wrapper  | ERROR: Could not write pid file /Users/julienlancelot/Dev/Sandbox/temp/sonarqube/sonarqube-5.4-SNAPSHOT/bin/macosx-universal-64/./SonarQube.pid: Permission denied
```
- Error when creating /extensions/deprecated
```
java.lang.IllegalStateException: The following directory can not be created: /Users/julienlancelot/Dev/Sandbox/temp/sonarqube/sonarqube-5.4-SNAPSHOT/extensions/deprecated
```

julienlancelot:
Web folder is now configurable to target of the ticket is validated.
If the goal is to have a full ready only directory installation, then please check if having the directory where is written the PID write access is an issue.

simonbrandhof:
feedback applied

PR #761, opened 2016-02-04
SONAR-7114 Add line when changeset is invalid

julienlancelot:
SONAR-7114 Add line when changeset is invalid


sns-seb:
small comment then LGTM

PR #762, opened 2016-02-04
SONAR-7112 Order views and subviews first in the search box

teryk:


server/sonar-web/src/main/webapp/WEB-INF/app/controllers/api/components_controller.rb
```diff
@@ -59,7 +59,7 @@ def suggestions
 
     json = {'total' => results.size}
     json_results = []
-    java_facade.getResourceTypes().each do |resource_type|
+    ordered_resource_types(java_facade.getResourceTypes()).each do |resource_type|
```
simonbrandhof: why not doing that in Java code (getResourceTypes()) ?

simonbrandhof:
LGTM for code. Will test on new deployed milestone....

PR #763, opened 2016-02-04
less access to file sources

sns-seb:
related to SONAR-6940 but since it's only partial implementation, the ticket won't be closed by added to the backlog


julienlancelot:
Would be nice to have some unit tests, then LGTM

PR #764, opened 2016-02-04
SONAR-7166 add explicit error when CeTask.componentKey is null

sns-seb:
+ use `MessageException` instead of `IllegalStateException`

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/computation/step/LoadReportAnalysisMetadataHolderStep.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/computation/step/LoadReportAnalysisMetadataHolderStep.java
@@ -19,12 +19,13 @@
  */
 package org.sonar.server.computation.step;
 
+import org.sonar.api.utils.MessageException;
 import org.sonar.batch.protocol.output.BatchReport;
 import org.sonar.server.computation.analysis.MutableAnalysisMetadataHolder;
 import org.sonar.server.computation.batch.BatchReportReader;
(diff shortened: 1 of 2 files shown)


server/sonar-server/src/main/java/org/sonar/server/computation/step/LoadReportAnalysisMetadataHolderStep.java
```diff
@@ -55,11 +56,19 @@ public void execute() {
 
   private void checkProjectKeyConsistency(BatchReport.Metadata reportMetadata) {
     String reportProjectKey = projectKeyFromReport(reportMetadata);
-    checkState(
-      ceTask.getComponentKey().equals(reportProjectKey),
-      "ProjectKey in report (%s) is not consistent with projectKey under which the report as been submitted (%s)",
-      reportProjectKey,
-      ceTask.getComponentKey());
+    if (ceTask.getComponentKey() == null) {
+      throw MessageException.of(format(
```
julienlancelot: Why not put this check directly in CeTask.Builder.setComponentKey() ?

julienlancelot:
One suggestion, else LGTM

server/sonar-server/src/main/java/org/sonar/server/computation/step/LoadReportAnalysisMetadataHolderStep.java
sns-seb: It's a totally legit for a CeTask to not have a component key, either it a task unrelated to a component or to a non existing or not yet existing one

PR #765, opened 2016-02-04
SONAR-7123 Improve size of projects name columns

julienlancelot:
SONAR-7123 Improve size of projects name columns


simonbrandhof:
LGTM

PR #769, opened 2016-02-04
remove dependency on sonar-channel and switch to sslr

lmihalkovic:
I noticed sonar-markdown makes use of sonar-channel whereas most other pieces of the project have switched to sslr. It also seems that a difference in guava version stands in the way. This is a possible way to move to sslr.

### DIFF
--- a/sonar-markdown/pom.xml
+++ b/sonar-markdown/pom.xml
@@ -14,10 +14,6 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.codehaus.sonar</groupId>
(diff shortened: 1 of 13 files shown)


simonbrandhof:
You're right about the usage of the deprecated sonar-channel, but the guava conflict prevents from moving to sslr-core. 
In any case I think that this sonar-markdown lib should be dropped and replaced by a standard lib like txtmark.

simonbrandhof:
Closing this PR as build fails.

PR #779, opened 2016-02-16
SONAR-7352 Deterministic order of groups in My Account page

teryk:


simonbrandhof:
Note: this ordering is case-sensitive ("aaa", "bbb" then "AAA"), but it's not an issue in this context.
LGTM

PR #788, opened 2016-02-24
SONAR-7316 use String.intern in ChangeSet + fix DBScmInfo

sns-seb:


sns-seb:
PR based on master when it should be based on branch-5.4

PR #792, opened 2016-02-25
SONAR-6810 Remove unused internal WS api/issues/show

simonbrandhof:
Stas confirmed that it is not used by webapp. Moreover it is not supported by sonar-ws-client 4.5 LTS.


julienlancelot:
LGTM

PR #796, opened 2016-02-29
Replace SQL "count(*)" by "count(1)"

simonbrandhof:
Replace SQL "count(*)" by "count(1)"

It can have slight performance improvements on some databases.


julienlancelot:
LGTM

PR #804, opened 2016-03-01
SONAR-7267 Do not accept multiple rules with the same key in a repository

simonbrandhof:
SONAR-7267 Do not accept multiple rules with the same key in a repository

### DIFF
--- a/sonar-plugin-api/src/main/java/org/sonar/api/server/rule/RulesDefinition.java
+++ b/sonar-plugin-api/src/main/java/org/sonar/api/server/rule/RulesDefinition.java
@@ -20,6 +20,7 @@
 package org.sonar.api.server.rule;
 
 import com.google.common.base.Objects;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
@@ -46,7 +47,6 @@ import org.sonar.api.server.ServerSide;
 import org.sonar.api.server.debt.DebtRemediationFunction;
 import org.sonar.api.utils.log.Loggers;
 
-import static com.google.common.base.Objects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 import static java.lang.String.format;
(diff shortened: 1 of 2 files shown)


sonar-plugin-api/src/main/java/org/sonar/api/server/rule/RulesDefinition.java
```diff
@@ -488,14 +488,13 @@ public NewRepositoryImpl setName(@Nullable String s) {
...
+     */
     @Override
     public NewRule createRule(String ruleKey) {
-      if (newRules.containsKey(ruleKey)) {
-        // Should fail in a perfect world, but at the time being the Findbugs plugin
-        // defines several times the rule EC_INCOMPATIBLE_ARRAY_COMPARE
-        // See http://jira.sonarsource.com/browse/SONARJAVA-428
-        Loggers.get(getClass()).warn(format("The rule '%s' of repository '%s' is declared several times", ruleKey, key));
-      }
+      Preconditions.checkArgument(!newRules.containsKey(ruleKey), "The rule '%s' of repository '%s' is declared several times", ruleKey, key);
```
sns-seb: you could do a static import here

sonar-plugin-api/src/test/java/org/sonar/api/server/rule/RulesDefinitionTest.java
```diff
@@ -257,13 +260,12 @@ public void cant_set_blank_repository_name() {
   }
 
   @Test
-  public void warning_if_duplicated_rule_keys() {
+  public void fail_if_duplicated_rule_keys_in_the_same_repository() {
+    expectedException.expect(IllegalArgumentException.class);
```
sns-seb: @julienlancelot and I find it a better practice to declare the expectedException right before the statement supposed to throw it
It avoids catching an exception raised by another statement and not have a test failure when we should and it's also better for readability of the UT (it shows where the exception is expected to be raised)

sonar-plugin-api/src/main/java/org/sonar/api/server/rule/RulesDefinition.java
```diff
@@ -488,14 +488,13 @@ public NewRepositoryImpl setName(@Nullable String s) {
       return this;
     }
 
+    /**
+     * The specified rule key must be unique among the repository. Since version 5.5 it raises
+     * a {@link IllegalArgumentException}, whereas it logs a warning in previous versions.
+     */
```
sns-seb: why not use standard javadoc `@throws`
in addition, shouldn't this be documented on interface method `NewExtendedRepository.createRule(String ruleKey)`?

sns-seb:
some documentation and UT practices comments, otherwise LGTM

simonbrandhof:
@sns-seb feedback applied, thanks

PR #809, opened 2016-03-03
Add missing Javadoc

Godin:
Method was added in SQ 5.3 as part of SONAR-5894.


dbmeneses:
thanks, i've rebased & merged

PR #816, opened 2016-03-06
SONAR-7430 Wrong remote IP in access.log when server is behind a reverse proxy

simonbrandhof:
SONAR-7430 Wrong remote IP in access.log when server is behind a reverse proxy


sns-seb:
LGTM

PR #828, opened 2016-03-10
SONAR-7325 Display values of parameters in SQL TRACE logs

simonbrandhof:


julienlancelot:
I like this feature !!! (But not tested, just see the notification)

sonar-db/src/main/java/org/sonar/db/profiling/SqlLogFormatter.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.apache.commons.lang.StringUtils.replace;
+
+public enum SqlLogFormatter {
+  FORMATTER;
+
+  public static final String PARAM_NULL = "[null]";
+  public static final int PARAM_MAX_WIDTH = 500;
+
+  public String formatSql(String sql) {
+    return StringUtils.replaceChars(sql, '\n', ' ');
```
sns-seb: don't use StringUtils, instantiate a Matcher and store it in a static variable, you'll avoid wasting time and heap on creating the same object again and again

sns-seb: this actually was a general comment on the whole class
in case of chars replacement, might not be worse it.
Anyway, looking at the implementation is very important because this method method will intensively be called

sonar-db/src/main/java/org/sonar/db/profiling/ProfilingPreparedStatementHandler.java
```diff
@@ -19,43 +19,44 @@
...
-    }
+    int argCount = StringUtils.countMatches(sql, "?");
+    sqlParams = new Object[argCount];
   }
 
   @Override
   public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
     if (method.getName().startsWith("execute")) {
-      Profiler profiler = Profiler.create(ProfiledDataSource.SQL_LOGGER).start();
+      Profiler profiler = Profiler.createIfTrace(ProfiledDataSource.SQL_LOGGER).start();
```
sns-seb: we should be smarter than this
if logger is not at TRACE level, we shouldn't ever have the profiler and populate it's context at all (see finally block)
we are wasting memory and CPU on building strings which will never be used

simonbrandhof: No need. Package `org.sonar.db.profiling` is enabled only when TRACE level is enabled.

simonbrandhof: Here `createIfTrace` is a defensive approach in case of bug.

sonar-db/src/main/java/org/sonar/db/profiling/SqlLogFormatter.java
simonbrandhof: I don't get your point about Matcher. `StringUtils.replaceChars` uses `String.replace()` which simply iterates over the characters.

sonar-db/src/main/java/org/sonar/db/profiling/ProfilingPreparedStatementHandler.java
```diff
@@ -19,43 +19,44 @@
...
+  private final Object[] sqlParams;
 
   ProfilingPreparedStatementHandler(PreparedStatement statement, String sql) {
     this.statement = statement;
     this.sql = sql;
-    this.arguments = Lists.newArrayList();
-    for (int argCount = 0; argCount < StringUtils.countMatches(sql, "?"); argCount++) {
-      arguments.add("!");
-    }
+    int argCount = StringUtils.countMatches(sql, "?");
```
sns-seb: we should have our own implementation of countMatches which take a char as an argument
this method from StringUtils is based on `String.indexOf(String, int)` and I bet it could be lightning faster to just loop over the char array of the string

sonar-db/src/main/java/org/sonar/db/profiling/ProfilingStatementHandler.java
```diff
@@ -36,12 +37,13 @@
...
   public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
     if (method.getName().startsWith("execute")) {
-      Profiler profiler = Profiler.create(ProfiledDataSource.SQL_LOGGER).start();
+      Profiler profiler = Profiler.createIfTrace(ProfiledDataSource.SQL_LOGGER).start();
       Object result = null;
       try {
         result = InvocationUtils.invokeQuietly(statement, method, args);
       } finally {
-        profiler.addContext("sql", StringUtils.remove((String) args[0], '\n'));
+        String sql = (String) args[0];
```
sns-seb: we should not build the context at all when logger is not at TRACE level

simonbrandhof: See my previous comment. This is not possible.

sonar-db/src/main/java/org/sonar/db/profiling/ProfilingPreparedStatementHandler.java
simonbrandhof: I agree for this one.

sonar-db/src/main/java/org/sonar/db/profiling/SqlLogFormatter.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public String formatParam(@Nullable Object param) {
+    if (param == null) {
+      return PARAM_NULL;
+    }
+    return replace(abbreviate(param.toString(), PARAM_MAX_WIDTH), "\n", "\\n");
```
sns-seb: what's the reason of the replace?

it's a shame we can't factor replace with abbreviate

anyway, abbreviate is hardly efficient, it creates many substrings
in case of replace, we should use our own Matcher as a static variable. In addition, StringUtils is based on a slow `StringBuilder`

simonbrandhof: request should be logged on a single line, but still the newlines should not be removed for debugging purpose. That's why newlines are escaped.

abbreviate does not suffer from performance issues when length is less than limit, which is the most case.

I agree for `replace` though.

sonar-db/src/main/java/org/sonar/db/profiling/ProfilingPreparedStatementHandler.java
sns-seb: why not handle the bug completely then?

you made the change and it made me think the code could run when logger is not at TRACE level

either revert this change or handle the bug correctly, in current form, it's confusing

sonar-db/src/main/java/org/sonar/db/profiling/SqlLogFormatter.java
sns-seb: see my second comment, I noticed it was characters after my first comment
we actually agree

simonbrandhof:
@sns-seb feedback applied

sonar-db/src/main/java/org/sonar/db/profiling/ProfilingStatementHandler.java
```diff
@@ -36,12 +37,13 @@
   @Override
   public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
     if (method.getName().startsWith("execute")) {
-      Profiler profiler = Profiler.create(ProfiledDataSource.SQL_LOGGER).start();
+      Profiler profiler = Profiler.createIfTrace(ProfiledDataSource.SQL_LOGGER).start();
```
sns-seb: you forgot to remove that `createIfTrace`

sonar-db/src/main/java/org/sonar/db/profiling/SqlLogFormatter.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+import org.apache.commons.lang.StringUtils;
+
+import static org.apache.commons.lang.StringUtils.abbreviate;
+
+public enum SqlLogFormatter {
+  FORMATTER;
+
+  public static final String PARAM_NULL = "[null]";
```
sns-seb: could be private

sonar-db/src/main/java/org/sonar/db/profiling/SqlLogFormatter.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+package org.sonar.db.profiling;
+
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+import org.apache.commons.lang.StringUtils;
+
+import static org.apache.commons.lang.StringUtils.abbreviate;
+
+public enum SqlLogFormatter {
```
sns-seb: all methods in this class could be static

sns-seb:
LGTM

sonar-db/src/main/java/org/sonar/db/profiling/SqlLogFormatter.java
sns-seb: @simonbrandhof I'm not sure you saw this last comment before the LGTM

sonar-db/src/main/java/org/sonar/db/profiling/SqlLogFormatter.java
sns-seb: @simonbrandhof I'm not sure you saw this last comment before the LGTM

simonbrandhof: @sns-seb I did. It is private : https://github.com/SonarSource/sonarqube/blob/master/sonar-db/src/main/java/org/sonar/db/profiling/SqlLogFormatter.java#L31

PR #831, opened 2016-03-10
Improve configuration of HTTP proxies

simonbrandhof:
SONAR-7429 support HTTPS proxy
SONAR-7465 support encryption of proxy properties

### DIFF
--- a/sonar-application/src/main/assembly/conf/sonar.properties
+++ b/sonar-application/src/main/assembly/conf/sonar.properties
@@ -248,6 +248,10 @@
 #http.proxyHost=
 #http.proxyPort=
 
+# HTTPS proxy (default none)
+#https.proxyHost=
+#https.proxyPort=
+
 # NT domain name if NTLM proxy is used
 #http.auth.ntlm.domain=
 
@@ -255,7 +259,7 @@
 #socksProxyHost=
 #socksProxyPort=
 
-# proxy authentication. The 2 following properties are used for HTTP and SOCKS proxies.
+# Proxy authentication (used for HTTP, HTTPS and SOCKS proxies)
 #http.proxyUser=
 #http.proxyPassword=
 
--- a/sonar-core/src/main/java/org/sonar/core/util/DefaultHttpDownloader.java
+++ b/sonar-core/src/main/java/org/sonar/core/util/DefaultHttpDownloader.java
@@ -175,6 +175,7 @@ public class DefaultHttpDownloader extends HttpDownloader {
 
     private static final List<String> PROXY_SETTINGS = ImmutableList.of(
       "http.proxyHost", "http.proxyPort", "http.nonProxyHosts",
+      "https.proxyHost", "https.proxyPort",
       "http.auth.ntlm.domain", "socksProxyHost", "socksProxyPort");
 
     private String userAgent;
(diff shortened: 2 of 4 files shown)


SonarTech:
* BLOCKER 1 blocker

* BLOCKER [NullPointerException might be thrown as 'value' is nullable here](https://github.com/SonarSource/sonarqube/blob/8369ccc2aaa65cb3213bdee93a260d95d7a90e3c/sonar-plugin-api/src/main/java/org/sonar/api/config/Settings.java#L269) [rule](https://dory.sonarsource.com/coding_rules#rule_key=squid%3AS2259)

sonar-application/src/main/assembly/conf/sonar.properties
```diff
@@ -255,7 +259,7 @@
 #socksProxyHost=
 #socksProxyPort=
 
-# proxy authentication. The 2 following properties are used for HTTP and SOCKS proxies.
+# Proxy authentication (used for HTTP, HTTPS and SOCKS proxies)
 #http.proxyUser=
 #http.proxyPassword=
```
sns-seb: By design, you can't have a user/password specific to https or http or sock.

Is it because you can only configure one type of proxy at a time?

sonar-core/src/main/java/org/sonar/core/util/DefaultHttpDownloader.java
```diff
@@ -75,7 +75,7 @@ public DefaultHttpDownloader(Server server, Settings settings, @Nullable Integer
   public DefaultHttpDownloader(Server server, Settings settings, @Nullable Integer connectTimeout, @Nullable Integer readTimeout) {
     this.readTimeout = readTimeout;
     this.connectTimeout = connectTimeout;
-    downloader = new BaseHttpDownloader(settings.getProperties(), server.getVersion());
+    downloader = new BaseHttpDownloader(new ProxySystem(), settings, server.getVersion());
```
sns-seb: this pattern of `new ProxySystem()` and settings as arguments of the `BaseHttpDownloader`'s constructor caught my eye

it looks like it's only justified by the UT code testing how `ProxySystem` is populated from the `Settings`. This object is only built from `Settings`.

My advice, move `ProxySystem` out of `DefaultHttpDownloader` as a package protected class, make the constructor private and create a static method to create a `ProxySystem` instance from a `Settings` instance. Test the built object in the UT of `ProxySystem`. You'll end up with a `BaseHttpDownloader(ProxySystem system, @Nullable String userAgent)` constructor and better SOC.

sns-seb:
`ProxySystem` should be moved out of `DefaultHttpDownloader`, then LGTM

sonar-application/src/main/assembly/conf/sonar.properties
simonbrandhof: Exact. It's currently an accepted limitation.

simonbrandhof:
`ProxySystem` is a facade to allow unit tests to verify calls to `java.lang.System`. This class could be replaced by `org.sonar.api.utils.System2`,  but it sounds overkill to define the method `setDefaultAuthenticator(Authenticator)` in it.
Explanation added to Javadoc.

sonar-core/src/main/java/org/sonar/core/util/DefaultHttpDownloader.java
sns-seb: I misunderstood the point of `ProxySystem`, forget my comment

PR #832, opened 2016-03-11
Remove commons-collections

simonbrandhof:


sns-seb:
some optimisations and maybe a forgotten unit test, then LGTM

PR #838, opened 2016-03-14
SONAR-7479 Remove system tags from DB when no tags in plugin

julienlancelot:
SONAR-7479 Remove system tags from DB when no tags in plugin

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
@@ -24,7 +24,6 @@ import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -60,6 +59,7 @@ import org.sonar.server.rule.index.RuleIndexer;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.Lists.newArrayList;
+import static com.google.common.collect.Sets.intersection;
 
 /**
  * Register rules at server startup
@@ -77,7 +77,7 @@ public class RegisterRules implements Startable {
   private final System2 system2;
 
   public RegisterRules(RuleDefinitionsLoader defLoader, RuleActivator ruleActivator, DbClient dbClient, RuleIndexer ruleIndexer,
-                       ActiveRuleIndexer activeRuleIndexer, Languages languages, System2 system2) {
+    ActiveRuleIndexer activeRuleIndexer, Languages languages, System2 system2) {
     this.defLoader = defLoader;
     this.ruleActivator = ruleActivator;
     this.dbClient = dbClient;
@@ -200,10 +200,6 @@ public class RegisterRules implements Startable {
       changed = true;
     }
     if (mergeDescription(def, dto)) {
-      changed = true;
-    }
-    if (!dto.getSystemTags().containsAll(def.tags())) {
-      dto.setSystemTags(def.tags());
       changed = true;
     }
     if (!StringUtils.equals(dto.getConfigKey(), def.internalKey())) {
@@ -348,8 +344,9 @@ public class RegisterRules implements Startable {
     if (RuleStatus.REMOVED == ruleDef.status()) {
       dto.setSystemTags(Collections.<String>emptySet());
       changed = true;
-    } else if (!dto.getSystemTags().containsAll(ruleDef.tags())
-      || !Sets.intersection(dto.getTags(), ruleDef.tags()).isEmpty()) {
+    } else if (dto.getSystemTags().size() != ruleDef.tags().size() ||
+      !dto.getSystemTags().containsAll(ruleDef.tags())
+      || !intersection(dto.getTags(), ruleDef.tags()).isEmpty()) {
       dto.setSystemTags(ruleDef.tags());
       // remove end-user tags that are now declared as system
       RuleTagHelper.applyTags(dto, ImmutableSet.copyOf(dto.getTags()));
--- a/server/sonar-server/src/test/java/org/sonar/server/rule/RegisterRulesTest.java
+++ b/server/sonar-server/src/test/java/org/sonar/server/rule/RegisterRulesTest.java
@@ -19,7 +19,6 @@
  */
 package org.sonar.server.rule;
 
-import com.google.common.collect.Sets;
 import java.util.Date;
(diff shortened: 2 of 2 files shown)


SonarTech:
* MAJOR 1 major

* MAJOR [Make "mergeTags" a "static" method.](https://github.com/SonarSource/sonarqube/blob/9f74f7465c6fb365627f720820b152f97805a054/server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java#L346) [rule](https://dory.sonarsource.com/coding_rules#rule_key=squid%3AS2325)

server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
```diff
@@ -60,6 +59,8 @@
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.Lists.newArrayList;
+import static com.google.common.collect.Sets.difference;
```
teryk: Class should be called RulesRegister

julienlancelot: In fact the pattern is RegisterXXX, have a look at RegisterQualityProfiles

julienlancelot: But there'll be some refactoring of these 2 classes as some optimization needs to be done, so it will be a good opportunity to improve the name !

server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
```diff
@@ -202,7 +203,7 @@ private boolean mergeRule(RulesDefinition.Rule def, RuleDto dto) {
     if (mergeDescription(def, dto)) {
       changed = true;
     }
-    if (!dto.getSystemTags().containsAll(def.tags())) {
+    if (!difference(dto.getSystemTags(), def.tags()).isEmpty()) {
```
teryk: if my understanding is correct, is `def.tags()` has tags not in db, it won't update the db! The `difference(Set,Set)` method works only in one way, not the other

julienlancelot: I disagree, otherwise every tests would fail.

server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
```diff
@@ -342,14 +343,14 @@ private boolean mergeParam(RuleParamDto paramDto, RulesDefinition.Param paramDef
...
+  private static boolean mergeTags(RulesDefinition.Rule ruleDef, RuleDto dto) {
     boolean changed = false;
 
     if (RuleStatus.REMOVED == ruleDef.status()) {
       dto.setSystemTags(Collections.<String>emptySet());
       changed = true;
     } else if (!dto.getSystemTags().containsAll(ruleDef.tags())
-      || !Sets.intersection(dto.getTags(), ruleDef.tags()).isEmpty()) {
+      || !intersection(dto.getTags(), ruleDef.tags()).isEmpty()) {
       dto.setSystemTags(ruleDef.tags());
```
teryk: how are system tags different from 'normal' tags? Difference between `dto.tags()` & `dto.getSystemTags()`?

server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
```diff
@@ -342,14 +343,14 @@ private boolean mergeParam(RuleParamDto paramDto, RulesDefinition.Param paramDef
...
       // remove end-user tags that are now declared as system
       RuleTagHelper.applyTags(dto, ImmutableSet.copyOf(dto.getTags()));
```
teryk: do you know why end-user tags are now declared as system?

server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
julienlancelot: Hum, in fact you're probably right, I'm adding a UT.

teryk: I confirm. It works because afterwards there's the `mergeTags` method that makes different tests. I don't understand why there are 2 places to test system tags.

teryk:
Not your fault but I think the code lacks clarity concerning tags merging.

teryk:
It works correctly but the code is misleading

server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
julienlancelot: Indeed, it should have failed, but another merge of tags is done line 346 !
I'm fixing this

server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
julienlancelot: system tags = tags provided by plugin => cannot be changed by user
other tag = user tags

server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
julienlancelot: The goal is to remove user tags that are now system tags

julienlancelot: See RegisterRulesMediumTest#remove_user_tags_that_are_newly_declared_as_system

server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
```diff
@@ -349,7 +344,8 @@ private static boolean mergeTags(RulesDefinition.Rule ruleDef, RuleDto dto) {
     if (RuleStatus.REMOVED == ruleDef.status()) {
       dto.setSystemTags(Collections.<String>emptySet());
       changed = true;
-    } else if (!dto.getSystemTags().containsAll(ruleDef.tags())
+    } else if (dto.getSystemTags().size() != ruleDef.tags().size() ||
```
teryk: I would replace by `!intersection(dto.getSystemTags(). ruleDef.tags()).size() != ruleDef.tags().size`? WDYT?

teryk: do we know that `dto.getSystemTags()` and `ruleDef.tags()` have the same equivalence order? Both are HashSet?

teryk: Otherwise your alternative is better

teryk:
LGTM

PR #841, opened 2016-03-15
SONAR-6719 Wrong casing when indexing individual files lead to issue …

dbmeneses:
…during scm blame


dbmeneses:
Tested this on Windows and confirmed that it is creating InputFiles with the actual case but without following links.

henryju:
Could you add a unit or medium test specific to Windows that create a project with folder "src" and try to analyze with sonar.sources=SRC
Appart from that LGTM

dbmeneses:
Ok, even though it's not running in AppVeyor

PR #849, opened 2016-03-17
SONAR-7273 API returns different paths in case of single-module and m…

dbmeneses:
…ulti-module configurations


dbmeneses:
This is the last remaining getCanonical in the batch.

henryju:
LGTM

PR #852, opened 2016-03-18
SONAR-6941 Partition files of scanner report in different folders

henryju:


sonar-scanner-protocol/src/main/java/org/sonar/scanner/protocol/output/FileStructure.java
```diff
@@ -71,8 +75,30 @@ public File activeRules() {
...
+      result = result.resolve(String.valueOf(c));
+    }
+    return result;
+  }
+
   public File fileFor(Domain domain, int componentRef) {
-    return new File(dir, domain.filePrefix + componentRef + domain.fileSuffix);
+    Path parent = dir.toPath().resolve(getSubDirFor(componentRef));
+    try {
+      FileUtils.forceMkdir(parent.toFile());
```
dbmeneses: Files.createDirectories(parent)

dbmeneses:
1 comment, then LGTM

henryju:
@dbmeneses Good catch, fixed.

sns-seb:
@Duarte I think we should rather stick with Java NIO and use https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createDirectories(java.nio.file.Path,%20java.nio.file.attribute.FileAttribute...)

Commons IO is packed with performance issues and after Commons Collection, I think we should aim at removing this Commons library next

henryju:
@sns-seb That was exactly the point of @dbmeneses feedback :)

sns-seb:
true, I read it the otherway around, sorry

duarte:
Hey, I think you @ mentioned the wrong duarte

PR #856, opened 2016-03-21
SONAR-7478 External identity information in users WS

teryk:


sonar-db/src/test/java/org/sonar/db/user/GroupTesting.java
```diff
@@ -30,8 +29,8 @@ public static GroupDto newGroupDto() {
     GroupDto group = new GroupDto()
       .setName(randomAlphanumeric(255))
       .setDescription(randomAlphanumeric(200));
-    group.setCreatedAt(new Date(nextLong()));
-    group.setUpdatedAt(new Date(nextLong()));
+    group.setCreatedAt(new Date());
```
julienlancelot: XXXTesting should always generate some random data : if we need some fixed data, we should manually set them and not expect the dto generated by the XXXTesting to have some known value in fields.
So I would prefer to revert these 2 changes, and to set the new Date() in your test.

server/sonar-server/src/test/java/org/sonar/server/user/ws/SearchActionTest.java
```diff
@@ -192,16 +227,12 @@ public void search_with_groups() throws Exception {
...
 
-      users[index] = new UserDoc()
-        .setActive(true)
-        .setCreatedAt(createdAt)
-        .setEmail(email)
-        .setLogin(login)
-        .setName(name)
-        .setScmAccounts(scmAccounts)
-        .setUpdatedAt(createdAt);
+      users[index] = toUserDoc(userDto);
```
julienlancelot: To prevent having to create UserDoc, you know that you can use the UserIndexer (This feedback is not to be applied now).

julienlancelot:
Works well for the /users/search WS, but you've forgot to work on the /users/current WS.

server/sonar-server/src/test/java/org/sonar/server/user/ws/SearchActionTest.java
teryk: ok, I didn't know I could use it in tests

sonar-db/src/test/java/org/sonar/db/user/GroupTesting.java
teryk: ok

teryk:
users/current added, thanks for the feedback @julienlancelot

julienlancelot:
LGTM

PR #869, opened 2016-03-24
SONAR-7458 Expose SQ Version in SensorContext

henryju:


sonar-core/src/main/java/org/sonar/core/platform/SonarQubeVersionProvider.java
```diff
@@ -36,14 +36,18 @@
...
-        version = new SonarQubeVersion(Version.parse(versionInFile));
-      } catch (IOException e) {
-        throw new IllegalStateException("Can not load " + FILE_PATH + " from classpath", e);
-      }
+      version = getVersion(system);
     }
     return version;
   }
+
+  public static SonarQubeVersion getVersion(System2 system) {
```
simonbrandhof: This method should be private. Moreover I suggest to name it `loadVersion()`.

sonar-plugin-api/src/main/java/org/sonar/api/batch/sensor/internal/SensorContextTester.java
```diff
@@ -117,6 +133,15 @@ public void setActiveRules(ActiveRules activeRules) {
   }
 
   @Override
+  public SonarQubeVersion getSonarQubeVersion() {
+    return sqVersion;
+  }
+
+  public void setSonarQubeVersion(Version version) {
```
simonbrandhof: Method could be fluent and return `this`.

simonbrandhof: `@since` is missing.

sonar-plugin-api/src/main/java/org/sonar/api/batch/sensor/internal/SensorContextTester.java
```diff
@@ -117,6 +133,15 @@ public void setActiveRules(ActiveRules activeRules) {
   }
 
   @Override
+  public SonarQubeVersion getSonarQubeVersion() {
```
simonbrandhof: `@since` is missing.

simonbrandhof: javadoc should mention that the default value is the version of API dependency and it can be overridden with `setSonarQubeVersion()`

sonar-plugin-api/src/main/java/org/sonar/api/batch/sensor/internal/SensorContextTester.java
```diff
@@ -76,16 +82,26 @@
...
+  private SonarQubeVersion sqVersion;
 
-  private SensorContextTester(File moduleBaseDir) {
+  private SensorContextTester(Path moduleBaseDir) {
     this.settings = new Settings();
     this.fs = new DefaultFileSystem(moduleBaseDir);
     this.activeRules = new ActiveRulesBuilder().build();
     this.sensorStorage = new InMemorySensorStorage();
     this.module = new DefaultInputModule("projectKey");
+    try {
```
simonbrandhof: This block should be replaced with `SonarQubeVersionProvider#loadVersion()`

sonar-plugin-api/src/main/java/org/sonar/api/batch/sensor/SensorContext.java
```diff
@@ -62,6 +63,11 @@
    */
   InputModule module();
 
+  /**
+   * @since 5.5
+   */
+  SonarQubeVersion getSonarQubeVersion();
```
simonbrandhof: Do you think that this shortcut method deserves to be documented in the example of `SonarQubeVersion` ?

sonar-core/src/main/java/org/sonar/core/platform/SonarQubeVersionProvider.java
henryju: I intended to use it in SensorContextTester then realized that sonar-core is not visible from sonar-plugin-api...

henryju: I'll move this code to sonar-plugin-api

henryju:
@simonbrandhof Feedback applied

sonar-plugin-api/src/main/java/org/sonar/api/batch/sensor/SensorContext.java
henryju: I have added special example for Sensors

sonar-plugin-api/src/main/java/org/sonar/api/batch/sensor/internal/SensorContextTester.java
```diff
@@ -76,16 +80,22 @@
...
+  private SonarQubeVersion sqVersion;
 
-  private SensorContextTester(File moduleBaseDir) {
+  private SensorContextTester(Path moduleBaseDir) {
     this.settings = new Settings();
     this.fs = new DefaultFileSystem(moduleBaseDir);
     this.activeRules = new ActiveRulesBuilder().build();
     this.sensorStorage = new InMemorySensorStorage();
     this.module = new DefaultInputModule("projectKey");
+    this.sqVersion = SonarQubeVersion.getVersion(new System2());
```
simonbrandhof: System2.INSTANCE must be used.

PR #873, opened 2016-03-25
System Info must not fail if the monitored process is not up

simonbrandhof:
System Info must not fail if the monitored process is not up

### DIFF
--- a/server/sonar-process/src/main/java/org/sonar/process/jmx/JmxConnectionFactory.java
+++ b/server/sonar-process/src/main/java/org/sonar/process/jmx/JmxConnectionFactory.java
@@ -20,6 +20,7 @@
 package org.sonar.process.jmx;
 
 import java.io.File;
+import javax.annotation.CheckForNull;
 import javax.annotation.concurrent.Immutable;
 import javax.management.remote.JMXConnector;
 import javax.management.remote.JMXConnectorFactory;
@@ -44,14 +45,23 @@ public class JmxConnectionFactory {
     this.ipcSharedDir = props.nonNullValueAsFile(ProcessEntryPoint.PROPERTY_SHARED_PATH);
   }
 
+  @CheckForNull
   public JmxConnection create(ProcessId processId) {
     try (DefaultProcessCommands commands = DefaultProcessCommands.secondary(ipcSharedDir, processId.getIpcIndex())) {
-      String url = commands.getJmxUrl();
-      JMXConnector jmxConnector = JMXConnectorFactory.newJMXConnector(new JMXServiceURL(url), null);
-      jmxConnector.connect();
-      return new JmxConnection(jmxConnector);
+      if (commands.isUp()) {
+        String url = commands.getJmxUrl();
+        JMXConnector jmxConnector = JMXConnectorFactory.newJMXConnector(new JMXServiceURL(url), null);
+        jmxConnector.connect();
+        return new JmxConnection(jmxConnector);
+      }
+      return null;
     } catch (Exception e) {
       throw new IllegalStateException("Can not connect to process " + processId, e);
     }
+  }
+
+  // visible for testing
+  File getIpcSharedDir() {
+    return ipcSharedDir;
   }
 }
--- a/server/sonar-process/src/test/java/org/sonar/process/jmx/JmxConnectionTest.java
+++ b/server/sonar-process/src/test/java/org/sonar/process/jmx/JmxConnectionTest.java
@@ -19,11 +19,46 @@
  */
 package org.sonar.process.jmx;
 
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
+import java.util.Map;
+import javax.management.ObjectName;
+import javax.management.remote.JMXConnector;
+import javax.management.remote.JMXConnectorFactory;
+import javax.management.remote.JMXConnectorServer;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class JmxConnectionTest {
+
+  public static final String CUSTOM_OBJECT_NAME = "Test:name=test";
+
+  @Rule
+  public ExpectedException expectedException = ExpectedException.none();
+
+  JMXConnectorServer jmxServer;
+  JmxConnection underTest;
+
+  @Before
+  public void setUp() throws Exception {
+    jmxServer = JmxTestUtils.startJmxServer();
+    jmxServer.getMBeanServer().registerMBean(new Fake(), new ObjectName(CUSTOM_OBJECT_NAME));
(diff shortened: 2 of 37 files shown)


server/sonar-server/src/main/java/org/sonar/server/platform/monitoring/Monitor.java
```diff
@@ -38,5 +38,5 @@
    * Type of attribute values must be supported by {@link org.sonar.api.utils.text.JsonWriter#valueObject(Object)}
    * because of JSON export by {@link InfoAction}
    */
-  LinkedHashMap<String, Object> attributes();
+  Optional<Map<String, Object>> attributes();
```
sns-seb: why/when `Optional.absent()` can be returned should be documented

server/sonar-process/src/main/java/org/sonar/process/jmx/JmxConnectionFactory.java
```diff
@@ -44,14 +45,23 @@ public JmxConnectionFactory(Props props) {
...
+        return new JmxConnection(jmxConnector);
+      }
+      return null;
     } catch (Exception e) {
       throw new IllegalStateException("Can not connect to process " + processId, e);
     }
   }
+
+  // visible for testing
+  File getIpcSharedDir() {
```
sns-seb: you could remove this method and the constructor with a `File` argument.
code would be as well covered and would not be cluttered with code for tests only

server/sonar-process/src/test/java/org/sonar/process/jmx/JmxConnectionTest.java
```diff
@@ -19,12 +19,47 @@
...
+import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class JmxConnectionTest {
 
+  public static final String CUSTOM_OBJECT_NAME = "Test:name=test";
```
sns-seb: I known it's only a UT, but this could be private

server/sonar-process/src/test/java/org/sonar/process/jmx/JmxConnectionTest.java
```diff
@@ -19,12 +19,47 @@
...
+
+  @Rule
+  public ExpectedException expectedException = ExpectedException.none();
+
+  JMXConnectorServer jmxServer;
+  JmxConnection underTest;
+
+  @Before
+  public void setUp() throws Exception {
+    jmxServer = JmxTestUtils.startJmxServer();
```
sns-seb: you're so close to create a Rule to create a Test JMX server :)

server/sonar-server/src/main/java/org/sonar/server/platform/monitoring/CeDatabaseMonitor.java
```diff
@@ -35,13 +37,16 @@ public CeDatabaseMonitor(JmxConnectionFactory jmxConnectionFactory) {
...
 
   @Override
-  public LinkedHashMap<String, Object> attributes() {
+  public Optional<Map<String, Object>> attributes() {
     try (JmxConnection connection = jmxConnectionFactory.create(ProcessId.COMPUTE_ENGINE)) {
-      LinkedHashMap<String, Object> result = new LinkedHashMap<>();
+      if (connection == null) {
+        return Optional.absent();
+      }
+      Map<String, Object> result = new LinkedHashMap<>();
```
sns-seb: minor memory usage optimization, but since we know the number of elements, we could create a map of 9 rather than 16 (default capacity)

server/sonar-server/src/main/java/org/sonar/server/platform/monitoring/CeTasksMonitor.java
```diff
@@ -39,9 +41,12 @@ public String name() {
...
 
   @Override
-  public LinkedHashMap<String, Object> attributes() {
+  public Optional<Map<String, Object>> attributes() {
     try (JmxConnection connection = jmxConnectionFactory.create(ProcessId.COMPUTE_ENGINE)) {
-      LinkedHashMap<String, Object> result = new LinkedHashMap<>();
+      if (connection == null) {
+        return Optional.absent();
+      }
+      Map<String, Object> result = new LinkedHashMap<>();
```
sns-seb: could create a map of 6 rather than 16 (default capacity)

sns-seb: implementation details checked, the actual backing array size is the nearest superior power of two, so here, 8. We gain 50% usage.

server/sonar-server/src/main/java/org/sonar/server/platform/monitoring/Monitor.java
sns-seb: and what difference from returning an empty map

server/sonar-server/src/test/java/org/sonar/server/platform/monitoring/CeStateMonitorTest.java
```diff
@@ -45,9 +46,16 @@ public void testName() {
...
+    assertThat(attributes.get()).containsExactly(
       MapEntry.entry("bar", "bar_val"),
       MapEntry.entry("foo", "foo_val"));
   }
+
+  @Test
+  public void absent_attributes_if_CE_is_down() {
+    when(jmxConnectionFactory.create(ProcessId.COMPUTE_ENGINE)).thenReturn(null);
+    Optional<Map<String, Object>> attributes = underTest.attributes();
+    assertThat(attributes.isPresent()).isFalse();
```
sns-seb: tip: you can statically import `org.assertj.guava.api.Assertions.assertThat` and then write `assertThat(attributes).isPresent()`

server/sonar-server/src/test/java/org/sonar/server/platform/monitoring/FakeMonitor.java
```diff
@@ -34,7 +36,8 @@ public String name() {
   }
 
   @Override
-  public LinkedHashMap<String, Object> attributes() {
-    return new LinkedHashMap<>();
+  public Optional<Map<String, Object>> attributes() {
+    Map<String, Object> map = new LinkedHashMap<>();
+    return Optional.of(map);
```
sns-seb: you can now use `Optional.of(Collections.emptyMap())`

sns-seb:
some cosmetic comments
actually tested on my box, LGTM

server/sonar-process/src/main/java/org/sonar/process/jmx/JmxConnectionFactory.java
simonbrandhof: No, the constructor with `File` is used by `JmxConnectionFactoryProvider` which does not have `Props`.

server/sonar-process/src/test/java/org/sonar/process/jmx/JmxConnectionTest.java
simonbrandhof: done !

PR #874, opened 2016-03-28
Add System2.SYSTEM for better static imports

simonbrandhof:
Add System2.SYSTEM for better static imports

### DIFF
--- a/server/sonar-server/src/test/java/org/sonar/server/user/ws/ChangePasswordActionTest.java
+++ b/server/sonar-server/src/test/java/org/sonar/server/user/ws/ChangePasswordActionTest.java
@@ -26,7 +26,6 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.sonar.api.config.Settings;
 import org.sonar.api.server.ws.WebService;
-import org.sonar.api.utils.System2;
 import org.sonar.core.permission.GlobalPermissions;
 import org.sonar.db.DbClient;
 import org.sonar.db.DbSession;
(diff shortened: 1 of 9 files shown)


server/sonar-server/src/test/java/org/sonar/server/user/ws/ChangePasswordActionTest.java
```diff
@@ -54,13 +53,14 @@
...
 import static org.mockito.Mockito.when;
+import static org.sonar.api.utils.System2.SYSTEM;
 
 public class ChangePasswordActionTest {
 
   static final Settings settings = new Settings().setProperty("sonar.defaultGroup", "sonar-users");
 
   @Rule
-  public DbTester dbTester = DbTester.create(System2.INSTANCE);
+  public DbTester dbTester = DbTester.create(SYSTEM);
```
sns-seb: considering static import (only), I would have gone further and named the constant `SYSTEM2`

sonar-plugin-api/src/main/java/org/sonar/api/utils/System2.java
```diff
@@ -69,6 +69,16 @@
   public static final System2 INSTANCE = new System2();
 
   /**
+   * Same as {@link #INSTANCE} but naming is more appropriate for static imports.
+   * @since 5.5
+   */
+  public static final System2 SYSTEM = INSTANCE;
+
+  protected System2() {
```
sns-seb: no issue in SQ for missing comment here?

why not use `private` (and put a comment saying "use INSTANCE" or "prevents direct instanciation") ?

sns-seb:
all modified classes should have static import of the new constant for this change to be beneficial
why changing some classes and not other? How were they chosen?
why not changing code in all unit tests?

I would put this change to the team's vote

PR #880, opened 2016-03-30
SONAR-7498 Keep Java Keystore and certificates when sending HTTPS requests on Java 8

jycr:


sns-seb:
Thanks you for your contribution.
Your pull request won't be applied as is but the actual one is greatly inspired from it.

PR #882, opened 2016-03-31
Drop Issue Action API and WS

teryk:


sonar-scanner-engine/src/test/java/org/sonar/batch/mediumtest/tasks/TasksMediumTest.java
```diff
@@ -28,7 +28,6 @@
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.sonar.api.SonarPlugin;
-import org.sonar.api.issue.action.Actions;
```
julienlancelot: Very strange that the scanner was testing this API...

julienlancelot:
LGTM

PR #896, opened 2016-04-14
Complete sonar-ws

simonbrandhof:
Adds support of :
- POST api/projects/create
- POST api/projects/delete
- GET api/ce/task

### DIFF
--- a/sonar-ws/src/main/java/org/sonarqube/ws/client/ce/CeService.java
+++ b/sonar-ws/src/main/java/org/sonarqube/ws/client/ce/CeService.java
@@ -20,6 +20,7 @@
 
 package org.sonarqube.ws.client.ce;
 
+import org.sonarqube.ws.WsCe;
 import org.sonarqube.ws.WsCe.ActivityResponse;
 import org.sonarqube.ws.WsCe.TaskTypesWsResponse;
 import org.sonarqube.ws.client.BaseService;
@@ -33,6 +34,9 @@ import static org.sonarqube.ws.client.ce.CeWsParameters.PARAM_ONLY_CURRENTS;
 import static org.sonarqube.ws.client.ce.CeWsParameters.PARAM_STATUS;
 import static org.sonarqube.ws.client.ce.CeWsParameters.PARAM_TYPE;
 
+/**
+ * Maps web service {@code api/ce} (Compute Engine).
+ */
 public class CeService extends BaseService {
 
   public CeService(WsConnector wsConnector) {
@@ -58,4 +62,13 @@ public class CeService extends BaseService {
     return call(new GetRequest(path("task_types")), TaskTypesWsResponse.parser());
   }
 
+  /**
+   * Gets details of a Compute Engine task.
+   *
+   * @throws org.sonarqube.ws.client.HttpException if HTTP status code is not 2xx.
+   * @since 5.5
+   */
+  public WsCe.TaskResponse task(String id) {
+    return call(new GetRequest(path("task")).setParam("id", id), WsCe.TaskResponse.parser());
+  }
 }
--- a/sonar-ws/src/main/java/org/sonarqube/ws/client/project/ProjectsService.java
+++ b/sonar-ws/src/main/java/org/sonarqube/ws/client/project/ProjectsService.java
@@ -1,0 +1,67 @@
+/*
+ * SonarQube
+ * Copyright (C) 2009-2016 SonarSource SA
+ * mailto:contact AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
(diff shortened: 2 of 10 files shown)


sonar-ws/src/main/java/org/sonarqube/ws/client/ce/CeService.java
```diff
@@ -58,4 +62,13 @@ public TaskTypesWsResponse taskTypes() {
     return call(new GetRequest(path("task_types")), TaskTypesWsResponse.parser());
   }
 
+  /**
+   * Gets details of a Compute Engine task.
+   *
+   * @throws org.sonarqube.ws.client.HttpException if HTTP status code is not 2xx.
+   * @since 5.5
+   */
+  public WsCe.TaskResponse task(String id) {
```
teryk: the pattern used is to have a TaskWsRequest POJO. It's debatable when you have only one field, but I would prefer to have the same pattern everywhere

sonar-ws/src/main/java/org/sonarqube/ws/client/project/ProjectsService.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Provisions a new project. This method does not allow to specify the parameter {@code branch}.
+   *
+   * @param projectKey the non-null key of project
+   * @param projectName the non-null name of project
+   * @throws org.sonarqube.ws.client.HttpException if HTTP status code is not 2xx.
+   */
+  public void create(String projectKey, String projectName) {
```
teryk: please create a CreateWsRequest POJO with 2 fields

sonar-ws/src/main/java/org/sonarqube/ws/client/project/ProjectsService.java
```diff
@@ -0,0 +1,67 @@
...
+    PostRequest request = new PostRequest(path("create")).setParam("key", projectKey).setParam("name", projectName);
+    call(request);
+  }
+
+  /**
+   * Deletes a project with the specified id.
+   *
+   * @throws org.sonarqube.ws.client.HttpException if HTTP status code is not 2xx.
+   */
+  public void deleteById(String projectId) {
```
teryk: please follow the Web Api convention, and provide only one method delete(DeleteWsRequest)

simonbrandhof: I think this is overkill and not mandatory as a baby step. Nothing prevents to add such generic method in a future version if more parameters are created.

sonar-ws/src/main/java/org/sonarqube/ws/client/ce/CeService.java
simonbrandhof: This is overkill and not mandatory as a baby step. Nothing prevents to add such generic method in a future version if more parameters are created.

sonar-ws/src/main/java/org/sonarqube/ws/client/project/ProjectsService.java
simonbrandhof: Will do.

simonbrandhof: Are you sure about the name `CreateWsRequest` ? As protobuf is not used yet on this service, a simple inner class `NewProject` could be added.

sonar-ws/src/main/java/org/sonarqube/ws/client/project/ProjectsService.java
simonbrandhof: Another argument is that DeleteWsRequest would offer the same ambiguity than WS : are both fields mandatory ? Complementary ?

teryk: at least the method must be `delete` like the name of the WS

sonar-ws/src/main/java/org/sonarqube/ws/client/project/ProjectsService.java
teryk: this is the pattern used for all other WS. See [ActivityWsRequest](https://github.com/SonarSource/sonarqube/blob/bf27baa7ef13bef1a1296c1c09b845580e65be44/sonar-ws/src/main/java/org/sonarqube/ws/client/ce/ActivityWsRequest.java#L27-L27), one [SearchWsRequest](https://github.com/SonarSource/sonarqube/blob/f5d3461ffd3c5936a3bdc0999f6a9f71b9f10cd2/sonar-ws/src/main/java/org/sonarqube/ws/client/component/SearchWsRequest.java#L28-L28) for examples. You'll fine dozen of other examples in the ws client code. My point is I prefer to have the same pattern everywhere and we know where to find the information when needed.
Having the *WsRequest used in the Action gives a good way to validate the class, as it's well testeed

sonar-ws/src/main/java/org/sonarqube/ws/client/project/ProjectsService.java
simonbrandhof: impossible as both id and key are String.

teryk: we said we didn't want to have a documentation on the ws client side. We don't want business logic, and it's a way to introduce business logic. The documentation is the Web Api.
If we had to introduce business logic, I would much prefer a Builder pattern.
I'm ok to have the String projectId instead of the DeleteWsRequest for the time being. I'd like a few braking changes in forward versions as possible. In this case, it will be if we add in a later version a DeleteWsRequest

teryk: `delete(DeleteWsRequest)`follows the pattern used on the whole ws-client

simonbrandhof: Having `deleteById()` does not forbid forward-compatibility. Let's discuss offline.

teryk:
LGTM

PR #897, opened 2016-04-14
SONAR-7187 Fix perf issue

julienlancelot:


sonar-db/src/test/java/org/sonar/db/ce/CeActivityDaoTest.java
```diff
@@ -123,6 +123,23 @@ public void test_selectByQuery() {
   }
 
   @Test
+  public void test_selectByQuery_pagination() {
+    insert("TASK_1", REPORT, "PROJECT_1", CeActivityDto.Status.SUCCESS);
+    insert("TASK_2", REPORT, "PROJECT_1", CeActivityDto.Status.FAILED);
+    insert("TASK_3", REPORT, "PROJECT_2", CeActivityDto.Status.SUCCESS);
+    insert("TASK_4", "views", null, CeActivityDto.Status.SUCCESS);
+
+    CeTaskQuery query = new CeTaskQuery().setStatuses(Collections.<String>emptyList());
```
simonbrandhof: is it needed to have `setStatuses(Collections.<String>emptyList())` ?

sonar-db/src/test/java/org/sonar/db/ce/CeActivityDaoTest.java
```diff
@@ -123,6 +123,23 @@ public void test_selectByQuery() {
...
+    assertThat(underTest.selectByQuery(db.getSession(), query, 0, 1)).extracting("uuid").containsOnly("TASK_4");
```
simonbrandhof: You should document the fact that results are ordered by id desc.

sonar-db/src/test/java/org/sonar/db/ce/CeActivityDaoTest.java
julienlancelot: nope

sonar-db/src/test/java/org/sonar/db/ce/CeActivityDaoTest.java
julienlancelot: indeed

simonbrandhof: for readability of test, I suggest to extract a method like :
```
List<String> selectPageOfUuids(int offset, int pageSize) {}
```
so that assertions look like:
```
assertThat(selectPageOfUuids(1, 1)).containsExactly("TASK_3");
```

simonbrandhof: also I suggest to use `containsExactly()` everywhere instead of `hasSize` and `containsOnly`.

simonbrandhof:
I can't validate MSSQL and Oracle requests, so obviously LGTM if database tests pass !

sonar-db/src/test/java/org/sonar/db/ce/CeActivityDaoTest.java
julienlancelot: Good point

PR #898, opened 2016-04-14
Add context when failing to map a db measure to a ws measure

teryk:


server/sonar-server/src/main/java/org/sonar/server/measure/ws/MeasureDtoToWsMeasure.java
```diff
@@ -33,24 +33,28 @@ private MeasureDtoToWsMeasure() {
...
+            .setIndex(i)
+            .setValue(formatNumericalValue(measureDto.getVariation(i), metricDto)));
+        }
       }
-    }
 
-    return measure.build();
+      return measure.build();
+    } catch (Exception e) {
+      throw new IllegalStateException(String.format("Error while mapping a measure of metric key '%s' and characteristics %s", metricDto.getKey(), measureDto.toString()), e);
```
julienlancelot: characteristics => measure ?

julienlancelot: Rephrase the sentence "Error while mapping measure '%s' of metric ''%s" ?

teryk: I had difficulties. I'm talking about measure parameters

julienlancelot:
Apply feedback then LGTM

PR #913, opened 2016-04-27
Update components copy resource id when persisting components in CE

julienlancelot:
Update components copy resource id when persisting components in CE

### DIFF
--- a/server/sonar-server/src/test/java/org/sonar/server/computation/step/ViewsPersistComponentsStepTest.java
+++ b/server/sonar-server/src/test/java/org/sonar/server/computation/step/ViewsPersistComponentsStepTest.java
@@ -29,6 +29,7 @@ import org.sonar.api.resources.Scopes;
 import org.sonar.api.utils.System2;
 import org.sonar.db.DbClient;
 import org.sonar.db.DbTester;
+import org.sonar.db.component.ComponentDbTester;
 import org.sonar.db.component.ComponentDto;
 import org.sonar.db.component.ComponentTesting;
 import org.sonar.server.computation.batch.TreeRootHolderRule;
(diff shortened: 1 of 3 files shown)


server/sonar-server/src/test/java/org/sonar/server/computation/step/ViewsPersistComponentsStepTest.java
```diff
@@ -239,16 +294,16 @@ public void update_view_name_and_longName() {
...
       .setName(PROJECT_VIEW_1_NAME)
       .setDescription("project view description is not persisted")
-      .setProjectViewAttributes(new ProjectViewAttributes(PROJECT_1_ID));
+      .setProjectViewAttributes(new ProjectViewAttributes(project.getId()));
   }
 
-  private void persistComponentDto(ComponentDto componentDto) {
-    dbClient.componentDao().insert(dbTester.getSession(), componentDto);
+  private void persistComponentDto(ComponentDto... componentDtos) {
+    dbClient.componentDao().insert(dbTester.getSession(), Arrays.asList(componentDtos));
```
teryk: I would use ComponentDbTester instead of persistComponentDto

server/sonar-server/src/test/java/org/sonar/server/computation/step/ViewsPersistComponentsStepTest.java
```diff
@@ -264,10 +319,14 @@ private void assertDtoNotUpdated(String componentKey) {
     assertThat(getComponentFromDb(componentKey).getCreatedAt()).isNotEqualTo(now);
   }
 
-  private ComponentDto createViewDto() {
+  private static ComponentDto createViewDto() {
     return ComponentTesting.newView(VIEW_UUID).setKey(VIEW_KEY).setName(VIEW_NAME);
   }
 
+  private static ComponentDto createProjectDto() {
+    return ComponentTesting.newProjectDto();
```
teryk: why not use the method directly with static import?

teryk:
LGTM with minor feedbacks

server/sonar-server/src/test/java/org/sonar/server/computation/step/ViewsPersistComponentsStepTest.java
julienlancelot: Good question, it's only to be compliant with createViewDto()

server/sonar-server/src/test/java/org/sonar/server/computation/step/ViewsPersistComponentsStepTest.java
julienlancelot: indeed

server/sonar-server/src/test/java/org/sonar/server/computation/step/ViewsPersistComponentsStepTest.java
teryk: I would rename createViewDto to newViewDto in that case

julienlancelot: ok

PR #924, opened 2016-05-04
encode user login name, fix SONAR-7165

jinkerry:
if username is Chinese, will raise exception


simonbrandhof:
Thanks for the contribution but we don't release patch on non-LTS and non-latest release versions. Bug is already fixed in 5.2 and won't be applied to 5.1.1. Thanks for your understanding.

jinkerry:
Got it.
Thanks for you reply.

PR #933, opened 2016-05-09
SONAR-7275 fail with specific message when loading rating settings

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/computation/qualitymodel/RatingSettings.java
```diff
@@ -50,7 +51,11 @@ public RatingSettings(Settings settings) {
     String[] languageConfigIndexes = settings.getStringArray(LANGUAGE_SPECIFIC_PARAMETERS);
     for (String languageConfigIndex : languageConfigIndexes) {
       String languagePropertyKey = LANGUAGE_SPECIFIC_PARAMETERS + "." + languageConfigIndex + "." + LANGUAGE_SPECIFIC_PARAMETERS_LANGUAGE_KEY;
-      builder.put(settings.getString(languagePropertyKey), LanguageSpecificConfiguration.create(settings, languageConfigIndex));
+      String languageKey = settings.getString(languagePropertyKey);
+      if (languageKey == null) {
+        throw MessageException.of("Technical debt configuration is corrupted. At least one Language specific parameter has no language key");
```
simonbrandhof: why `Language` starts with an upper-case letter ?

simonbrandhof: I think it deserves to give hints about where this is configured (General Settings -> Technical Debt -> property "Language specific parameters")

sns-seb: applied

PR #937, opened 2016-05-09
SONAR-7532 display submitter login in logs when available

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/computation/taskprocessor/CeWorkerCallableImpl.java
```diff
@@ -96,14 +96,25 @@ private void executeTask(CeTask task) {
...
+          task.getComponentKey(), task.getType(), task.getUuid(), getSubmitterLoginForLog(task));
     } else {
-      profiler.stopInfo("Executed task | project={} | type={} | id={}", task.getComponentKey(), task.getType(), task.getUuid());
+      profiler.stopInfo("Executed task | project={} | type={} | id={} | submitter='{}'",
+          task.getComponentKey(), task.getType(), task.getUuid(), getSubmitterLoginForLog(task));
     }
   }
+
+  private static String getSubmitterLoginForLog(CeTask task) {
+    if (task.getSubmitterLogin() == null) {
```
simonbrandhof: it can be replaced by `StringUtils.defaultString(task.getSubmitterLogin())`.

simonbrandhof:
Simple feedback then LGTM

server/sonar-server/src/main/java/org/sonar/server/computation/taskprocessor/CeWorkerCallableImpl.java
sns-seb: I'll still need a specific private method in order to factor code, which make useless to use a library method for such simple code

PR #943, opened 2016-05-10
SONAR-7555 Add measure on number of won'f fix issues

julienlancelot:


it/it-tests/src/test/java/it/Category2Suite.java
```diff
@@ -67,7 +69,9 @@
   IssueActionTest.class,
   IssueBulkChangeTest.class,
   IssueChangelogTest.class,
+  IssueCreationTest.class,
```
teryk: good catch! How did you spot it?

sonar-core/src/main/resources/org/sonar/l10n/core.properties
```diff
@@ -2728,6 +2728,9 @@ metric.profile_version.description=Selected quality profile version
 metric.false_positive_issues.name=False Positive Issues
 metric.false_positive_issues.description=False positive issues
 
+metric.wont_fix_issues.name=Won't Fix  Issues
```
teryk: all metrics should be defined in l10n bundle. Out of the scope of this PR

teryk:
LGTM

it/it-tests/src/test/java/it/Category2Suite.java
julienlancelot: As I wanted to add a new IT, I've added in IssueMeasureTest, but it was failing => As jobs were green I suspect that it was not executed.
Yes, since we've re-organized the ITs it's now easy to forget test, but there's no easy solution for the moment.

sonar-core/src/main/resources/org/sonar/l10n/core.properties
julienlancelot: indeed !

PR #949, opened 2016-05-10
SONAR-7532 do not log submitter when unset

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/computation/taskprocessor/CeWorkerCallableImpl.java
```diff
@@ -97,24 +97,26 @@ private void executeTask(CeTask task) {
 
   private static Profiler startProfiler(CeTask task) {
     Profiler profiler = Profiler.create(LOG);
-    return profiler.startInfo("Execute task | project={} | type={} | id={} | submitter='{}'",
-      task.getComponentKey(), task.getType(), task.getUuid(), getSubmitterLoginForLog(task));
+    return profiler.startInfo("Execute task | project={} | type={} | id={}{}",
+      task.getComponentKey(), task.getType(), task.getUuid(), submitterSuffix(task));
```
simonbrandhof: You can simply use `profiler.addContext(...)` !

sns-seb: indeed, it looks better

sns-seb: hum... but time is now the first "parameter", not the last one...

simonbrandhof: is it so important ? I don't think.

sns-seb:
@simonbrandhof improved `Profiler` to log time after context and so, used it in `CeWorkerCallableImpl`

simonbrandhof:
LGTM

server/sonar-server/src/main/java/org/sonar/server/computation/taskprocessor/CeWorkerCallableImpl.java
sns-seb: yes, for parsing, time is the most relevant information, having it at the end make parsing easier

PR #960, opened 2016-05-12
SONAR-7175 DbCleaner keeps the oldest analysis instead of the newest one

julienlancelot:


travis.sh
```diff
@@ -28,7 +28,7 @@ CI)
         -Pdeploy-sonarsource \
         -B -e -V
 
-  elif [[ "${TRAVIS_BRANCH}" == "branch-"* ]] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
```
teryk: do not push in master : )

julienlancelot: Indeed, that's why it's in the name of the commit !

teryk:
LGTM

teryk:
thank you for the scenario to reproduce the bug

julienlancelot:
Thanks ! And for the scenario to reproduce, it was hard to find, but I'm proud to find it !

PR #961, opened 2016-05-12
SONAR-7545 functional error when updating a project key

teryk:
- when updating a project key and a sub-component has a key longer than 400 characters
- when a component has a name longer than 2000 characters
- when a component has a key longer than 400 characters

### DIFF
--- a/sonar-db/src/main/java/org/sonar/db/component/ComponentValidator.java
+++ b/sonar-db/src/main/java/org/sonar/db/component/ComponentValidator.java
@@ -21,35 +21,31 @@
 package org.sonar.db.component;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static org.sonar.core.component.ComponentKeys.MAX_COMPONENT_KEY_LENGTH;
 
 public class ComponentValidator {
-  private static final int MAX_NAME_LENGTH = 2000;
-  private static final int MAX_KEY_LENGTH = 400;
-  private static final int MAX_QUALIFIER_LENGTH = 10;
+  private static final int MAX_COMPONENT_NAME_LENGTH = 2000;
+  private static final int MAX_COMPONENT_QUALIFIER_LENGTH = 10;
 
   private ComponentValidator() {
     // prevent instantiation
   }
 
   public static String checkComponentName(String name) {
-    checkArgument(name.length() <= MAX_NAME_LENGTH, "Component name length (%s) is longer than the maximum authorized (%s). '%s' was provided.",
-      name.length(), MAX_NAME_LENGTH, name);
+    checkArgument(name.length() <= MAX_COMPONENT_NAME_LENGTH, "Component name length (%s) is longer than the maximum authorized (%s). '%s' was provided.",
+      name.length(), MAX_COMPONENT_NAME_LENGTH, name);
     return name;
   }
 
   public static String checkComponentKey(String key) {
-    checkArgument(isComponentKeyValid(key), "Component key length (%s) is longer than the maximum authorized (%s). '%s' was provided.",
-      key.length(), MAX_KEY_LENGTH, key);
+    checkArgument(key.length() <= MAX_COMPONENT_KEY_LENGTH, "Component key length (%s) is longer than the maximum authorized (%s). '%s' was provided.",
+      key.length(), MAX_COMPONENT_KEY_LENGTH, key);
     return key;
   }
 
-  public static boolean isComponentKeyValid(String key) {
-    return key.length() <= MAX_KEY_LENGTH;
-  }
-
   public static String checkComponentQualifier(String qualifier) {
-    checkArgument(qualifier.length() <= MAX_QUALIFIER_LENGTH, "Component qualifier length (%s) is longer than the maximum authorized (%s). '%s' was provided.",
-      qualifier.length(), MAX_QUALIFIER_LENGTH, qualifier);
+    checkArgument(qualifier.length() <= MAX_COMPONENT_QUALIFIER_LENGTH, "Component qualifier length (%s) is longer than the maximum authorized (%s). '%s' was provided.",
+      qualifier.length(), MAX_COMPONENT_QUALIFIER_LENGTH, qualifier);
     return qualifier;
   }
 }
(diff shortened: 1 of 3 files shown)


server/sonar-server/src/test/java/org/sonar/server/component/NewComponentTest.java
```diff
@@ -23,14 +23,14 @@
...
-import static org.apache.commons.lang.RandomStringUtils.randomAlphabetic;
+import static com.google.common.base.Strings.repeat;
 
 public class NewComponentTest {
   @Rule
   public ExpectedException expectedException = ExpectedException.none();
 
   @Test
-  public void fail_when_key_is_null() {
+  public void fail_with_functional_exception_when_key_is_null() {
```
julienlancelot: name of the test is wrong, it should be something like 'fail_when_key_is_bigger_than_400_characters'

server/sonar-server/src/test/java/org/sonar/server/component/NewComponentTest.java
```diff
@@ -42,11 +42,11 @@ public void fail_when_key_is_longer_than_400_characters() {
     expectedException.expect(IllegalArgumentException.class);
     expectedException.expectMessage("Component key length (401) is longer than the maximum authorized (400)");
 
-    NewComponent.create(randomAlphabetic(401), "name");
+    NewComponent.create(repeat("a", 400 + 1), "name");
   }
 
   @Test
-  public void fail_when_name_is_null() {
+  public void fail_with_functional_exception_when_name_is_null() {
```
julienlancelot: why are you adding "functional_exception" everywhere ?

sonar-db/src/main/java/org/sonar/db/component/ComponentDto.java
```diff
@@ -160,7 +163,7 @@ public String name() {
   }
 
   public ComponentDto setName(String name) {
-    this.name = name;
+    this.name = checkComponentName(name);
```
julienlancelot: Is it really the responsibility of the dto to check this ?

server/sonar-server/src/test/java/org/sonar/server/component/NewComponentTest.java
teryk: not everywhere, just null testing (it's a mistake regarding the 33 line). I use checkArgument and not requireNonNull because I want an IllegalArgumentException and not an NPE. Http code is 4XX and not 5XX when handled by the WS engine

server/sonar-server/src/test/java/org/sonar/server/component/NewComponentTest.java
teryk: this test checks nullability

sonar-db/src/main/java/org/sonar/db/component/ComponentValidator.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public static String checkComponentKey(String key) {
+    checkArgument(isComponentKeyValid(key), "Component key length (%s) is longer than the maximum authorized (%s). '%s' was provided.",
+      key.length(), MAX_KEY_LENGTH, key);
+    return key;
+  }
+
+  public static boolean isComponentKeyValid(String key) {
+    return key.length() <= MAX_KEY_LENGTH;
```
julienlancelot: You should use https://github.com/SonarSource/sonarqube/blob/master/sonar-core/src/main/java/org/sonar/core/component/ComponentKeys.java#L96

sonar-db/src/main/java/org/sonar/db/component/ComponentValidator.java
```diff
@@ -0,0 +1,55 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.db.component;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+public class ComponentValidator {
+  private static final int MAX_NAME_LENGTH = 2000;
+  private static final int MAX_KEY_LENGTH = 400;
```
julienlancelot: Using https://github.com/SonarSource/sonarqube/blob/master/sonar-core/src/main/java/org/sonar/core/component/ComponentKeys.java#L96, you can remove this constant

sonar-db/src/main/java/org/sonar/db/component/ComponentDto.java
teryk: I think it's a new pattern, had a lot of discussions with @simonbrandhof 
Yes, I think it should be checked at dto level in the end to prevent any SQL error

server/sonar-server/src/test/java/org/sonar/server/component/NewComponentTest.java
julienlancelot: Oups sorry, you're right, I've merged the content of this test with next one

server/sonar-server/src/test/java/org/sonar/server/component/NewComponentTest.java
julienlancelot: It's strange, because the exception is the same than before

sonar-db/src/main/java/org/sonar/db/component/ComponentValidator.java
teryk: I prefer to have it in ComponentValidator, but I agree code should be factored. I wouldn't use the pattern matching for each dto creation (too costly), I may be wrong

sonar-db/src/main/java/org/sonar/db/component/ComponentDto.java
julienlancelot: Ok, then it's a new pattern indeed. Good to know

sonar-db/src/main/java/org/sonar/db/component/ComponentValidator.java
teryk: maybe the methods should be called `checkComponentNameLength()` and 'isComponentNameLengthValid'

sonar-db/src/main/java/org/sonar/db/component/ComponentDto.java
simonbrandhof: Yes, it's definitely a good practice !

julienlancelot:
LGTM

PR #963, opened 2016-05-12
SONAR-7539 WS api/properties should return licenses if user is authenticated

simonbrandhof:
SONAR-7539 WS api/properties should return licenses if user is authenticated


julienlancelot:
LGTM

PR #968, opened 2016-05-13
SONAR-7154 main process must create property file in SQ's temp dir

sns-seb:
because the JVM's temp dir might just be read only (eg. temp dir of windows services launched with Local System Account on recent windows versions is actuallt read only)


simonbrandhof:
LGTM

PR #969, opened 2016-05-13
SONAR-6846 don't fail on duplicate row in snapshot_sources

sns-seb:


sonar-db/src/main/java/org/sonar/db/version/v50/FeedFileSources.java
```diff
@@ -137,15 +140,17 @@
     "f.enabled = ? " +
     "AND f.scope = 'FIL' " +
     "AND p.scope = 'PRJ' AND p.qualifier = 'TRK' " +
-    "AND fs.file_uuid IS NULL";
+    "AND fs.file_uuid IS NULL " +
+    "ORDER BY s.id";
```
simonbrandhof: I would order by s.id DESC in order to keep only the most recent one. WDYT ?

sonar-db/src/main/java/org/sonar/db/version/v50/FeedFileSources.java
```diff
@@ -183,6 +188,13 @@ public boolean handle(Row row, SqlStatement update) throws SQLException {
       byte[] longOverallCovCond = row.getNullableBytes(28);
       byte[] shortDuplicationData = row.getNullableBytes(29);
       byte[] longDuplicationData = row.getNullableBytes(30);
+      long snapshotId = row.getLong(31);
+
+      if (snapshotId == previousSnapshotId) {
+        return false;
+      } else {
```
simonbrandhof: {{else}} can be dropped. Better for vertical readability.

simonbrandhof:
Feedback on ordering then LGTM

sonar-db/src/main/java/org/sonar/db/version/v50/FeedFileSources.java
sns-seb: yes it can

sonar-db/src/main/java/org/sonar/db/version/v50/FeedFileSources.java
sns-seb: this ordering applies to snapshot, not to snapshot_sources, to achieve what you suggest we should add a second ordering (desc) `snapshot_sources.id`. That's not a bad idea.

PR #971, opened 2016-05-13
SONAR-7400 Improve performance when searching for activated rules

julienlancelot:


server/sonar-server/src/main/java/org/sonar/server/rule/ws/ActiveRuleCompleter.java
```diff
@@ -141,34 +143,30 @@ void completeSearch(DbSession dbSession, RuleQuery query, List<RuleDto> rules, S
   }
 
   void completeShow(DbSession dbSession, RuleDto rule, ShowResponse.Builder response) {
-    List<ActiveRule> activeRules = loader.findActiveRulesByRule(rule.getKey());
-    List<ActiveRuleDto> activeRuleDtos = dbClient.activeRuleDao().selectByKeys(dbSession, Lists.transform(activeRules, ActiveRuleToKey.INSTANCE));
+    List<ActiveRuleDto> activeRuleDtos = dbClient.activeRuleDao().selectByRule(dbSession, rule);
```
simonbrandhof: This method should be replaced by `selectByRuleId(dbSession, rule.getId())`.

simonbrandhof:
Minor feedback else LGTM

server/sonar-server/src/main/java/org/sonar/server/rule/ws/ActiveRuleCompleter.java
julienlancelot: indeed

PR #974, opened 2016-05-13
SONAR-7626 Deprecate issues report (html/console) feature

dbmeneses:


sonar-scanner-engine/src/main/java/org/sonar/batch/scan/report/ConsoleReport.java
```diff
@@ -98,6 +98,7 @@ public void setNoFile(boolean value) {
   @Override
   public void execute() {
     if (settings.getBoolean(CONSOLE_REPORT_ENABLED_KEY)) {
+      LOG.info("Console report is deprecated. Use SonarLint CLI to have local reports of issues");
```
henryju: I would log with WARN level

sonar-scanner-engine/src/main/java/org/sonar/batch/scan/report/HtmlReport.java
```diff
@@ -77,6 +77,7 @@ public HtmlReport(Settings settings, FileSystem fs, IssuesReportBuilder builder,
   @Override
   public void execute() {
     if (settings.getBoolean(HTML_REPORT_ENABLED_KEY)) {
+      LOG.info("HTML report is deprecated. Use SonarLint CLI to have local reports of issues");
```
henryju: Same comment

henryju:
LGTM but I would log with WARN level

dbmeneses:
alright, changing to warn

PR #978, opened 2016-05-16
SONAR-7640 WS should return unauthenticated when login/password are set but invalid

julienlancelot:


n-b-sonar:
Run a few tests on LDAP/AD and it's looking good. Green light for me from a functional point of view.

it/it-tests/src/test/java/it/user/RailsExternalAuthenticationTest.java
```diff
@@ -340,6 +345,35 @@ public void update_password_of_technical_user() throws Exception {
...
+
+    // When user created in external system
+    users.put(login + ".password", password);
+    updateUsersInExtAuth(users);
+
+//    assertThat(checkAuthenticationWithWebService(login, password).code()).isEqualTo(200);
+    assertThat(checkAuthenticationWithWebService("wrong", password).code()).isEqualTo(401);
+//    assertThat(checkAuthenticationWithWebService(login, "wrong").code()).isEqualTo(401);
+//    assertThat(checkAuthenticationWithWebService(login, null).code()).isEqualTo(401);
+//    assertThat(checkAuthenticationWithWebService(null, null).code()).isEqualTo(200);
```
simonbrandhof: these 4 commented-out lines can be dropped

it/it-tests/src/test/java/it/user/RailsExternalAuthenticationTest.java
```diff
@@ -340,6 +345,35 @@ public void update_password_of_technical_user() throws Exception {
...
+
+    setServerProperty(orchestrator, "sonar.forceAuthentication", "true");
+
+    assertThat(checkAuthenticationWithWebService(login, password).code()).isEqualTo(200);
+    assertThat(checkAuthenticationWithWebService("wrong", password).code()).isEqualTo(401);
+    assertThat(checkAuthenticationWithWebService(login, "wrong").code()).isEqualTo(401);
+    assertThat(checkAuthenticationWithWebService(login, null).code()).isEqualTo(401);
+    assertThat(checkAuthenticationWithWebService(null, null).code()).isEqualTo(401);
```
simonbrandhof: 200 and 401 should be defined in constants, or use `java.net.HttpURLConnection.HTTP_UNAUTHORIZED` and `HTTP_OK`

server/sonar-web/src/main/webapp/WEB-INF/app/controllers/api/authentication_controller.rb
```diff
@@ -46,7 +46,15 @@ def validate
   private
 
   def valid?
-    logged_in? || (!force_authentication? && anonymous?)
+    is_logged_in?
+  end
+
+  def is_logged_in?
```
simonbrandhof: why this new method instead of inlining code in `valid?` ?

it/it-tests/src/test/java/it/user/RailsExternalAuthenticationTest.java
julienlancelot: arg, no I should uncomment them

it/it-tests/src/test/java/it/user/RailsExternalAuthenticationTest.java
julienlancelot: indeed

server/sonar-web/src/main/webapp/WEB-INF/app/controllers/api/authentication_controller.rb
julienlancelot: heu, indeed, I've forgot to merge them

simonbrandhof:
WS `api/authentication/validate` does not create the user session anymore. It only checks if credentials are valid. I don't think it can have side-effects on the services that call this WS (setup forms of SonarLint for instance). To keep it in mind. It deserves to send a small email to whole platform team.
LGTM

PR #980, opened 2016-05-16
Do not store _source in index rules/activeRule

simonbrandhof:
Do not store _source in index rules/activeRule


julienlancelot:
Very good changes, LGTM !

PR #981, opened 2016-05-16
Disable ES norms when possible

simonbrandhof:
Disable ES norms when possible

Norms are used when a string field is involved
in a match query. Basically they are useless
when only filters or aggregations are executed.
In this case they can be disabled. It won't
consume memory for nothing.


server/sonar-server/src/main/java/org/sonar/server/es/NewIndex.java
```diff
@@ -235,6 +236,17 @@ public StringFieldBuilder enableGramSearch() {
     }
 
     /**
+     * Norms consume useless memory if string field is used for filtering or aggregations.
+     *
+     * https://www.elastic.co/guide/en/elasticsearch/reference/2.3/norms.html
+     * https://www.elastic.co/guide/en/elasticsearch/guide/current/scoring-theory.html#field-norm
+     */
+    public StringFieldBuilder disableNorms() {
```
julienlancelot: From what I see, norms are mot of the time disabled on string fields. Shouldn't 'disabled' be the default value ?

PR #983, opened 2016-05-16
SONAR-6912 prevent admin lockout -- partial fix

janosgyerik:
Treats one part of one part of [SONAR-6912](https://jira.sonarsource.com/browse/SONAR-6912):

> The checks that should be done:
> - **when deleting a group: check if it's the last group with global admin rights** and there's no global admin user
> - when removing the last user of a group: check if it's the last group with global admin rights and there's no global admin user

That is, it will refuse to delete a group when it is the only system admin group. This is a bit more strict than needs to be. If there is an individual user with system admin role, it should allow deleting the group, but it doesn't, I marked with a "TODO" the place this condition could be implemented.

If you prefer I can implement that missing step now, or in a next Pull Request.

WS concerned: `api/user_groups/delete`

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/usergroups/ws/DeleteAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/usergroups/ws/DeleteAction.java
@@ -19,7 +19,6 @@
  */
 package org.sonar.server.usergroups.ws;
 
-import java.util.List;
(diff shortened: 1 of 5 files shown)


teryk:
thanks a lot for the contribution, it will be integrated in SQ shortly. Sorry for the delay. You were so impatient you decided to join SonarSource. That's great! : )

PR #985, opened 2016-05-17
SONAR-6438 Sanitize description of WS sections

stas-vilchik:


server/sonar-server/src/main/java/org/sonar/server/component/ws/ComponentsWs.java
```diff
@@ -39,7 +39,7 @@ public ComponentsWs(AppAction appAction, SearchViewComponentsAction searchViewCo
   public void define(Context context) {
     NewController controller = context.createController("api/components")
       .setSince("4.2")
-      .setDescription("Components management");
+      .setDescription("Get information about a component (file, directory, project, ...) and its ancestors or children.");
```
teryk: I don't agree with the provided description if we take into account internal WS ex: api/components/search

teryk: It's very specific whereas the domain goal is to search for components, more broadly @ganncamp WDYT?

server/sonar-server/src/main/java/org/sonar/server/issue/ws/IssuesWs.java
```diff
@@ -46,7 +46,7 @@ public IssuesWs(IssuesWsAction... actions) {
   @Override
   public void define(Context context) {
     NewController controller = context.createController(API_ENDPOINT);
-    controller.setDescription("Coding rule issues");
+    controller.setDescription("Read and update issues.");
```
teryk: why not Manage Issues? we search for issues, which is an important element of the domain @ganncamp WDYT?

server/sonar-server/src/main/java/org/sonar/server/test/ws/TestsWs.java
```diff
@@ -33,7 +33,7 @@ public TestsWs(TestsWsAction... actions) {
   public void define(Context context) {
     NewController controller = context.createController("api/tests")
       .setSince("4.4")
-      .setDescription("Tests management. Deprecated since 5.6");
+      .setDescription("Get details on test files. See also api/sources.");
```
teryk: you should keep the fact it's deprecated since 5.6

server/sonar-server/src/main/java/org/sonar/server/measure/ws/TimeMachineWs.java
```diff
@@ -30,7 +30,7 @@
   @Override
   public void define(Context context) {
     NewController controller = context.createController("api/timemachine")
-      .setDescription("Get a list of past measures")
+      .setDescription("Get project measure data from old analyses.")
```
teryk: Get project measure data from **past** analyses.

teryk:
LGTM eventually ann feedback will add some extra work

server/sonar-server/src/main/java/org/sonar/server/component/ws/ComponentsWs.java
ganncamp: Counter-propose @teryk :-)

server/sonar-server/src/main/java/org/sonar/server/issue/ws/IssuesWs.java
ganncamp: I didn't use 'manage' because it's not full CRUD @teryk

teryk: ok

server/sonar-server/src/main/java/org/sonar/server/component/ws/ComponentsWs.java
teryk: : ) you've got a point @ganncamp. My concern is that by being more specific we increase the likelihood of having a out-of-date documentation. I wasn't bothered by the Manage DOMAIN form personnaly. Here, I would have said « Manage components »

PR #986, opened 2016-05-17
SONAR-7271 Add license and serverID status to System Info 

simonbrandhof:


sonar-plugin-api/src/main/java/org/sonar/api/database/DatabaseProperties.java
```diff
@@ -21,21 +21,13 @@
...
 
-  /**
-   * @deprecated not documented for a while (before version 3.6) and not supported since version 5.2.
-   */
-  @Deprecated
-  String PROP_USER_DEPRECATED = "sonar.jdbc.user";
   String PROP_USER_DEFAULT_VALUE = "";
   String PROP_PASSWORD = "sonar.jdbc.password";
   String PROP_PASSWORD_DEFAULT_VALUE = "";
-  String PROP_DIALECT = "sonar.jdbc.dialect";
```
sns-seb: this one and `MAX_TEXT_SIZE` were actually not deprecated

server/sonar-web/src/main/js/apps/system/main.js
```diff
@@ -61,9 +61,11 @@ export default React.createClass({
   render() {
     let sections = null;
     if (this.state && this.state.sections) {
-      sections = this.state.sections.map(section => {
-        return <Section key={section.name} section={section.name} items={section.items}/>;
-      });
+      sections = this.state.sections
+        .filter(section => SECTIONS_ORDER.indexOf(section.name)>=0)
```
sns-seb: I don't understand the goal of this change

sns-seb: ok, it's explained in the commit description

server/sonar-server/src/main/java/org/sonar/server/platform/ServerIdLoader.java
```diff
@@ -0,0 +1,53 @@
...
+package org.sonar.server.platform;
+
+import com.google.common.base.Optional;
+import org.sonar.api.CoreProperties;
+import org.sonar.api.config.Settings;
+
+import static com.google.common.base.Optional.fromNullable;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public class ServerIdLoader {
```
sns-seb: I'd rather return an Object with the serverId and it's validity status than expose two methods
the current public methods of this class make is look like a move of code rather than a design of a class with a responsibility

server/sonar-server/src/main/java/org/sonar/server/platform/monitoring/SonarQubeMonitor.java
```diff
@@ -52,19 +54,23 @@
...
     this.server = server;
     this.serverLogging = serverLogging;
+    this.serverIdLoader = serverIdLoader;
   }
 
   @Override
   public String getServerId() {
-    return settings.getString(CoreProperties.PERMANENT_SERVER_ID);
+    Optional<String> idOpt = serverIdLoader.get();
+    return idOpt.isPresent() ? idOpt.get() : null;
```
sns-seb: you can use `Optional#orNull()` to single line this

server/sonar-web/src/main/js/apps/system/main.js
simonbrandhof: yep :-)

server/sonar-server/src/main/java/org/sonar/server/platform/ServerIdLoader.java
sns-seb: I see there might be a perf gain not doing the validation of the server id in some cases
so, I'll suggest create method `getAndValidate` returning a dedicated object instead of `isValid\

sonar-plugin-api/src/main/java/org/sonar/api/database/DatabaseProperties.java
simonbrandhof: you're right, but they're not used for a while

sonar-core/src/test/java/org/sonar/core/config/CorePropertyDefinitionsTest.java
```diff
@@ -31,4 +37,25 @@ public void all() {
     List<PropertyDefinition> defs = CorePropertyDefinitions.all();
     assertThat(defs.size()).isGreaterThan(9);
   }
+
+  @Test
+  public void jdbc_password_property_has_password_type() {
+    List<PropertyDefinition> defs = CorePropertyDefinitions.all();
+    Optional<PropertyDefinition> prop = from(defs).filter(new HasKeyPredicate(DatabaseProperties.PROP_PASSWORD)).first();
+    assertThat(prop.isPresent()).isTrue();
```
sns-seb: since `prop.get` will fail anyway at the next line, this assertion is not necessary

server/sonar-server/src/main/java/org/sonar/server/platform/monitoring/SonarQubeMonitor.java
simonbrandhof: excellent, I didn't know this method.

simonbrandhof:
@sns-seb Feedback applied.

server/sonar-server/src/main/java/org/sonar/server/platform/ServerId.java
```diff
@@ -0,0 +1,38 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.platform;
+
+public class ServerId {
```
sns-seb: missing `@Immutable`

simonbrandhof: should the class marked as `final` in order to be immutable ? The underlying question is : is the annotation `@Immutable` inherited by children ?

sns-seb: I didn't find any clear answer. But from what I read, this annotation concerns the design of the annotated class, not the object, so, it's not a problem the class is not final.

https://books.google.fr/books?id=EK43StEVfJIC&pg=PA353&lpg=PA353&dq=java+should+class+annotated+with+@Immutable+be+final&source=bl&ots=uoZAB7qQou&sig=TkjMUg-ntUQr0lLM8IgflISWbnE&hl=fr&sa=X&ved=0ahUKEwjijKK1nOPMAhWKvRoKHaRIC7YQ6AEIVzAH#v=onepage&q=java%20should%20class%20annotated%20with%20%40Immutable%20be%20final&f=false

Imho, this defensive approach of making the class final is not _required_ in this specific case (it would be required in an API) but in the other hand, making the class final won't hurt, so, I'll make it final.

sns-seb:
LGTM

PR #990, opened 2016-05-17
SONAR-7142 Cannot restart server after wiping out Server ID fields

simonbrandhof:
SONAR-7142 Cannot restart server after wiping out Server ID fields


sns-seb:
LGTM

PR #995, opened 2016-05-18
SONAR-7237 Measure variations are not computed when a component is created inside the period time frame

julienlancelot:


server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputeMeasureVariationsStep.java
```diff
@@ -100,14 +102,14 @@ public void execute() {
 
     private final DbSession session;
     private final Set<Integer> metricIds;
-    private final Map<String, Metric> metricByKeys;
+    private final Iterable<Metric> metrics;
 
-    public VariationMeasuresVisitor(DbSession session, Iterable<Metric> metrics) {
+    public VariationMeasuresVisitor(DbSession session, List<Metric> metrics) {
```
teryk: why change `Iterable` to `List` and not store the `metrics` as a `List`. There's a inconsistency here

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputeMeasureVariationsStep.java
```diff
@@ -165,7 +151,22 @@ private double computeVariation(Measure measure, Double pastValue) {
         case BOOLEAN:
           return (measure.getBooleanValue() ? 1d : 0d) - pastValue;
         default:
-          throw new IllegalArgumentException("Unsupported Measure.ValueType " + measure.getValueType());
+          throw new IllegalArgumentException(format("Unsupported Measure.ValueType on measure '%s'", measure));
```
teryk: you confirm the `toString()` is overloaded in Measure?

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputeMeasureVariationsStep.java
julienlancelot: You're right, the metrics should be stored as a list.

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputeMeasureVariationsStep.java
julienlancelot: https://github.com/SonarSource/sonarqube/blob/master/server/sonar-server/src/main/java/org/sonar/server/computation/measure/api/MeasureImpl.java#L85

server/sonar-server/src/test/java/org/sonar/server/computation/step/ReportComputeMeasureVariationsStepTest.java
```diff
@@ -151,6 +153,70 @@ public void set_variation() {
...
+    // Project
+    SnapshotDto period1ProjectSnapshot = newSnapshotForProject(PROJECT_DTO);
+    dbClient.snapshotDao().insert(session, period1ProjectSnapshot);
+    dbClient.measureDao().insert(session, newMeasureDto(ISSUES_METRIC.getId(), PROJECT_DTO.getId(), period1ProjectSnapshot.getId(), 60d));
+
+    // Directory
+    ComponentDto directoryDto = ComponentTesting.newDirectory(PROJECT_DTO, "dir");
+    dbClient.componentDao().insert(session, directoryDto);
+    SnapshotDto period1DirectorySnapshot = createForComponent(directoryDto, period1ProjectSnapshot);
+    dbClient.snapshotDao().insert(session, period1DirectorySnapshot);
```
teryk: no need to insert a snapshot and measure for directory

teryk: or add a test without snapshot

server/sonar-server/src/main/java/org/sonar/server/computation/step/ComputeMeasureVariationsStep.java
```diff
@@ -126,35 +128,19 @@ private MeasuresWithVariationRepository computeMeasuresWithVariations(Component
...
-            measureWithVariations.getVariation(4),
-            measureWithVariations.getVariation(5)))
-          .create();
-        measureRepository.update(component, metric, measure);
-      }
-    }
-
     private void setVariationMeasures(Component component, List<PastMeasureDto> pastMeasures, int period, MeasuresWithVariationRepository measuresWithVariationRepository) {
       Map<MeasureKey, PastMeasureDto> pastMeasuresByMeasureKey = from(pastMeasures).uniqueIndex(pastMeasureToMeasureKey);
-      for (Map.Entry<String, Measure> entry : from(measureRepository.getRawMeasures(component).entries()).filter(NotDeveloperMeasure.INSTANCE)) {
```
teryk: no need to check developer measure?

julienlancelot: No more indeed ! It was previously required when using measureRepository.getRawMeasures(component) because it's also returning dev measures, but measureRepository.getRawMeasure(component, metric) does not return dev measure, so I've been able to reduce the number of checks.
Moreover, using measureRepository.getRawMeasures(component) was returning far more measures that we want (it was returning dev measures and also measures on metric that we don't care, as we're only needing numeric metrics).

teryk:
feedback to make the test clearer and…

lgtm-bush

PR #1001, opened 2016-05-19
SONAR-6964 Do not display administration sensitive data when opening browser history

teryk:


it/it-tests/src/test/java/it/http/HttpHeadersTest.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.IOException;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import util.QaOnly;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Category(QaOnly.class)
+public class HttpHeadersTest {
```
julienlancelot: You should check that image and css are cached

teryk: done

julienlancelot: good !

julienlancelot:
LGTM

PR #1002, opened 2016-05-20
SONAR-7661 add interface Startable

sns-seb:


sonar-plugin-api/src/main/java/org/sonar/api/Startable.java
```diff
@@ -0,0 +1,54 @@
...
+ * {@link org.sonar.api.server.ServerSide} and/or {@link org.sonar.api.ce.ComputeEngineSide} (referred to below as
+ * "component") that can be started and stopped.
+ * <p>
+ * The method {@link #start()} is called at the begin of the component lifecycle.
+ * It can be called again only after a call to {@link #stop()}. The {@link #stop()} method is called at the end of the
+ * component lifecycle, and can further be called after every {@link Startable#start()}.
+ * </p>
+ * <p>
+ * In the WebServer, a component is started once: either right when the WebServer is started if there is no migration, otherwise
+ * only Database has been successfully migrated. And it is stopped once when the WebServer/ComputeEngine is shutdown.
```
simonbrandhof: "/ComputeEngine" can be dropped here as it's explained in next paragraph.

sonar-plugin-api/src/main/java/org/sonar/api/Startable.java
```diff
@@ -0,0 +1,54 @@
...
+ * </p>
+ * <p>
+ * In the Compute Engine, a component is started once when the Compute Engine is started and stopped once when the
+ * ComputeEngine is shut down.
+ * </p>
+ * <p>
+ * On Batch side, the lifecycle of a component depends on the value of the {@link org.sonar.api.batch.InstantiationStrategy}
```
simonbrandhof: "on batch side" should be renamed "on scanner side" (the annotation `@BatchSide` should be renamed too, but that's ouf of scope here).

sonar-plugin-api/src/main/java/org/sonar/api/Startable.java
```diff
@@ -0,0 +1,54 @@
...
+ * annotation.
+ * </p>
+ */
+public interface Startable {
+  /**
+   * Start this component. Called initially at the begin of the lifecycle. It can be called again after a stop.
```
simonbrandhof: Should we mention that throwing a `RuntimeException" makes startup fail or is it too obvious ?

simonbrandhof:
Remarks about comments, else LGTM

sonar-plugin-api/src/main/java/org/sonar/api/Startable.java
sns-seb: good catch

sonar-plugin-api/src/main/java/org/sonar/api/Startable.java
sns-seb: I'll remove use of "batch"

Regarding `@BatchSide` we should deprecate it in 5.6 to avoid having it around until 7.X LTS

sonar-plugin-api/src/main/java/org/sonar/api/Startable.java
sns-seb: adding it for good measure

sonar-plugin-api/src/main/java/org/sonar/api/Startable.java
sns-seb: can't be done in 5.6 (devs are frozen), https://jira.sonarsource.com/browse/SONAR-7665 opened to keep track of this in 6.X

PR #1005, opened 2016-05-23
Update property_utils.sh

priitliivak:
Now works on OSX. More reliable OS independent version of this script. Fixed incorrect backslash count in REGEX variable and applied recommendations for "if grep" http://mywiki.wooledge.org/BashPitfalls#if_.5Bgrep_foo_myfile.5D

### DIFF
--- a/scripts/property_utils.sh
+++ b/scripts/property_utils.sh
@@ -36,8 +36,11 @@ function set_property() {
      # delete line of specified property
     LINE_COUNT=$(cnt_lines $FILE)
 
-    # Create new file and replace afterwards. Avoids corruption if update fails
-    sed -e /${REGEXP}/d "$FILE" > "${FILE}.new" && mv "${FILE}.new" "${FILE}"
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+      sed -i '' /${REGEXP}/d "$FILE"
+    else
+      sed -i /${REGEXP}/d "$FILE"
+    fi
 
     # add property if at least one line deleted
     NEW_LINE_COUNT=$(cnt_lines $FILE)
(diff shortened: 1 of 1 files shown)


scripts/property_utils.sh
```diff
@@ -36,19 +30,24 @@ function set_property() {
...
-    # delete line of specified property
+  REGEXP="${PROPERTY//\./\\.}\\s*="
+
+  if grep -q "$REGEXP" "$FILE"; then
+     # delete line of specified property
     LINE_COUNT=$(cnt_lines $FILE)
-    sed -i $SED_DISABLE_BACKUP "/${REGEXP}/d" $FILE
+
+    # Create new file and replace afterwards. Avoids corruption if update fails
+    sed -e /${REGEXP}/d "$FILE" > "${FILE}.new" && mv "${FILE}.new" "${FILE}"
```
sns-seb: While I agree with other changes (thanks for the link in PR description btw, I learned a few stuff) I think this one is overkill.

set_property is intended to be used once or many times in a row in a script. This change implies as many copy and delete of files as there is properties to replace/overwrite. Seemless on SSD, maybe not so much on other systems.
Anyway, the work is half-done here since in case of failure, file "*.new" will be left on disk.

Unless this change is actually justified to make code more portable to Mac using `-e` option, is it ?

priitliivak: Yes -e flag is for Mac. 
Having the SED_DISABLE_BACKUP variable caused to create a backup file with quotes in the end (filename.properties"). Another option is to inline OS check. That would actually look nicer since it doesn't create a global variable as well.

sns-seb:
@priitliivak thanks for your contribution, merged into master after sqash and rebase

PR #1012, opened 2016-05-31
SONAR-7669 WS api/issues/search createdBefore cannot be equal to createdAfter

teryk:


it/it-tests/src/test/java/it/issue/IssueSearchTest.java
```diff
@@ -191,7 +190,7 @@ public void search_issues_by_dates() {
       search(IssueQuery.create().createdBefore(past).createdAfter(today)).list();
       Fail.fail("Expecting 400 from issues search WS");
     } catch (HttpException exception) {
-      assertThat(exception.getMessage()).contains("Start bound cannot be larger than end bound");
+      assertThat(exception.getMessage()).contains("Start bound cannot be larger or equal to end bound");
```
julienlancelot: It would have been best to use an ExpectedException rule :)

server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
```diff
@@ -380,8 +380,8 @@ private void addDatesFilter(Map<String, FilterBuilder> filters, IssueQuery query
   private void validateCreationDateBounds(Date createdBefore, Date createdAfter) {
     Preconditions.checkArgument(createdAfter == null || createdAfter.before(new Date(system.now())),
       "Start bound cannot be in the future");
-    Preconditions.checkArgument(createdAfter == null || createdAfter.equals(createdBefore) || createdBefore == null || createdAfter.before(createdBefore),
-      "Start bound cannot be larger than end bound");
+    Preconditions.checkArgument(createdAfter == null || createdBefore == null || createdAfter.before(createdBefore),
```
julienlancelot: This change would deserve a new case in the search_issues_by_dates UT

julienlancelot: Oups, forget my last comment, it's done

julienlancelot:
LGTM

it/it-tests/src/test/java/it/issue/IssueSearchTest.java
teryk: that's what I did on the tests I added

julienlancelot: Yes, but you could also have added it on existing ones (I've added a smiley because you would have gave me this exact same feedback !)

teryk: hahaha, you're right!

julienlancelot: : )

PR #1020, opened 2016-06-07
Functional interface annotation added when needed

rpau:
Hi,

I am sending to you an automatic correction to add the @FunctionalInterface annotation in interfaces that only contains one method. If it is too much to verify, let me know how would you like to proceed. For example: per module. 

According to the nemo project, it should reduce 2 hours approximately of technical debt :)

### DIFF
--- a/server/sonar-plugin-bridge/src/main/java/org/sonar/plugin/ce/ReportAnalysisComponentProvider.java
+++ b/server/sonar-plugin-bridge/src/main/java/org/sonar/plugin/ce/ReportAnalysisComponentProvider.java
@@ -22,6 +22,7 @@ package org.sonar.plugin.ce;
 import java.util.List;
 import org.sonar.api.ce.ComputeEngineSide;
 
+@FunctionalInterface
 @ComputeEngineSide
 public interface ReportAnalysisComponentProvider {
 
--- a/server/sonar-process/src/main/java/org/sonar/process/Stoppable.java
+++ b/server/sonar-process/src/main/java/org/sonar/process/Stoppable.java
@@ -19,6 +19,7 @@
  */
 package org.sonar.process;
 
+@FunctionalInterface
 public interface Stoppable {
 
   void stopAsync();
--- a/server/sonar-process/src/main/java/org/sonar/process/systeminfo/SystemInfoSection.java
+++ b/server/sonar-process/src/main/java/org/sonar/process/systeminfo/SystemInfoSection.java
@@ -21,6 +21,7 @@ package org.sonar.process.systeminfo;
 
 import org.sonar.process.systeminfo.protobuf.ProtobufSystemInfo;
 
+@FunctionalInterface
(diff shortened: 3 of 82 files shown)


sonar-db/src/main/java/org/sonar/db/qualitygate/ProjectQgateAssociationMapper.java
```diff
@@ -22,6 +22,7 @@
 import java.util.List;
 import org.apache.ibatis.annotations.Param;
 
+@FunctionalInterface
```
simonbrandhof: MyBatis mappers do not aim to FunctionalInterfaces.

sonar-plugin-api/src/main/java/org/sonar/api/component/RubyComponentService.java
```diff
@@ -25,6 +25,7 @@
 /**
  * @since 3.6
  */
+@FunctionalInterface
 @ServerSide
 public interface RubyComponentService {
```
simonbrandhof: This class does not aim to be a FunctionalInterface.

sonar-plugin-api/src/main/java/org/sonar/api/issue/batch/IssueFilter.java
```diff
@@ -28,6 +28,7 @@
  * @since 4.0
  * @deprecated since 5.3. Use {@link org.sonar.api.scan.issue.filter.IssueFilter} instead.
  */
+@FunctionalInterface
```
simonbrandhof: Deprecated classes should not be touched.

sonar-plugin-api/src/main/java/org/sonar/api/measures/MeasureBuilder.java
```diff
@@ -27,6 +27,7 @@
  * @since 1.10
  * @deprecated since 5.6. To create a new measure on scanner side use {@link SensorContext#newMeasure()}
  */
+@FunctionalInterface
```
simonbrandhof: Builders are not FunctionalInterface.

simonbrandhof:
Thanks for the contribution but @FunctionalInterface should not be put on all single-method interfaces. It depends on the responsibility of the interface. Moreover deprecated classes are usually not touched.

rpau:
I have removed from the list those classes that you reported me that are not functional interfaces

simonbrandhof:
Thanks, but as there are many touched classes that was only a partial review :-)

rpau:
So, would you prefer I split them by modules?

simonbrandhof:
No. You should add the annotation only where it makes sense, not on all Single Abstract Method (SAM) interfaces. For example ignore data structures and mybatis mappers.

simonbrandhof:
Thanks for the contribution but we will fix add the annotation on missing places during future refactorings.

PR #1035, opened 2016-06-13
SONAR-7683 table SNAPSHOTS should use component UUID

sns-seb:


server/sonar-web/src/main/webapp/WEB-INF/db/migrate/1212_drop_snapshot_project_id_from_snapshots.rb
```diff
@@ -0,0 +1,30 @@
...
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+
+#
+# SonarQube 6.0
+#
+class DropSnapshotProjectIdFromSnapshots < ActiveRecord::Migration
+
+  def self.up
+    remove_index :snapshots, :name => 'snapshot_project_id'
```
simonbrandhof: in order to be reentrant, each `remove_index` should be wrapped into:
```
begin
      remove_index :snapshots, ...
    rescue
      #ignore
    end
```

server/sonar-server/src/main/java/org/sonar/server/computation/dbcleaner/ProjectCleaner.java
```diff
@@ -59,7 +59,7 @@ public ProjectCleaner purge(DbSession session, IdUuidPair idUuidPair, Settings p
 
     PurgeConfiguration configuration = newDefaultPurgeConfiguration(projectSettings, idUuidPair);
 
-    cleanHistoricalData(session, configuration.rootProjectIdUuid().getId(), projectSettings);
+    cleanHistoricalData(session, configuration.rootProjectIdUuid().getUuid(), projectSettings);
```
simonbrandhof: Next refactoring would be to drop the class `IdUuidPair`. Each time UUID should be used.

simonbrandhof:
I didn't spot cause in code changes, but WS default ordering of api/components/tree by name changed from case-insensitive to case-sensitive.

simonbrandhof:
WS `api/measures/component` of project does not return measures anymore on a multi-modules project.
Example for Java plugin: `api/measures/component?componentKey=org.sonarsource.java:java&metricKeys=ncloc`

simonbrandhof:
my bad, I don't reproduce the issue with api/measures/component. The problem related to case-sensitive ordering is out of scope of this PR. I send an email to the team.

server/sonar-web/src/main/webapp/WEB-INF/app/controllers/project_controller.rb
```diff
@@ -320,7 +320,7 @@ def update_version
       end
     end
 
-    redirect_to :action => 'history', :id => snapshot.root_project_id
+    redirect_to :action => 'history', :id => snapshot.root.id
```
simonbrandhof: updating a version fails in webapp. Form is redirected on a wrong id. Are you sure that `snapshot.root` is the project but not the root snapshot ?

sns-seb: I'm not sure at all :)

sns-seb: good catch and you're right, it should be `snapshot.root_project.id`

PR #1047, opened 2016-06-17
SONAR-7691 table CE_ACTIVITY should use analysis UUID

simonbrandhof:


server/sonar-server/src/main/java/org/sonar/server/computation/queue/InternalCeQueueImpl.java
```diff
@@ -101,9 +101,9 @@ public void remove(CeTask task, CeActivityDto.Status status, CeTaskResult taskRe
 
   private static void updateTaskResult(CeActivityDto activityDto, @Nullable CeTaskResult taskResult) {
     if (taskResult != null) {
-      Long snapshotId = taskResult.getSnapshotId();
-      if (snapshotId != null) {
-        activityDto.setSnapshotId(snapshotId);
+      java.util.Optional<String> analysisUuid = taskResult.getAnalysisUuid();
```
sns-seb: the move to `java.util.Optional` should be discusses and organised, I think, it's generating ugliness in code :(

sonar-db/src/main/java/org/sonar/db/component/SnapshotDao.java
```diff
@@ -42,11 +43,19 @@ public static boolean isLast(SnapshotDto snapshotTested, @Nullable SnapshotDto p
     return previousLastSnapshot == null || previousLastSnapshot.getCreatedAt() < snapshotTested.getCreatedAt();
   }
 
+  /**
+   * @deprecated use {@link #selectByUuid(DbSession, String)}
+   */
+  @Deprecated
   @CheckForNull
   public SnapshotDto selectById(DbSession session, long id) {
```
sns-seb: I've checked, it seems we will be able to drop `selectById` and `selectOrFailById` when we remove the use of SNAPSHOTS.PARENT_ID and SNAPSHOTS.ROOT_ID

sonar-db/src/main/java/org/sonar/db/version/v60/PopulateAnalysisUuidColumnOnCeActivity.java
```diff
@@ -0,0 +1,58 @@
...
+    massUpdate.rowPluralName("ce_activity");
+    massUpdate.execute((row, update) -> this.handle(row, update));
+  }
+
+  private boolean handle(Select.Row row, SqlStatement update) throws SQLException {
+    long id = row.getLong(1);
+    String analysisUuid = row.getNullableString(2);
+
+    if (analysisUuid == null) {
+      return false;
```
sns-seb: I've decided to put a TRACE log in other migration which apply the same pattern, just in case support has to investigate rows deletion

sonar-db/src/test/resources/org/sonar/db/version/v60/AddComponentUuidColumnsToSnapshotsTest/old_snapshots.sql
```diff
@@ -30,3 +30,4 @@ CREATE TABLE "SNAPSHOTS" (
   "PERIOD5_PARAM" VARCHAR(100),
   "PERIOD5_DATE" BIGINT
 );
+
```
sns-seb: why this change?

sns-seb:
LGTM

sonar-db/src/main/java/org/sonar/db/component/SnapshotDao.java
simonbrandhof: Exact, all usages of snapshot ID must be dropped.

sonar-db/src/main/java/org/sonar/db/version/v60/PopulateAnalysisUuidColumnOnCeActivity.java
simonbrandhof: in fact it's impossible to have a NULL UUID here. SNAPSHOTS.UUID is not null at that time.

PR #1051, opened 2016-06-20
PropsBuilder respects property "sonar.path.home" if it is set

Argelbargel:
I'm trying to divide my sonarqube-install in such a way that the libraries etc. which do not change and all data and configuration are kept in separate directories (e.g. app/sonarqube and work/sonarqube). 

This is not possible right now because PropsBuilder#detectHome() does not respect the property "sonar.path.home" but uses a path relative to the location of the libraries.


Argelbargel:
Closed because tests might not work as sonar.path.home is not reset in before()

PR #1052, opened 2016-06-20
Respect sonar.path.home in PropsBuilder

Argelbargel:
I'm trying to divide my sonarqube-install in such a way that the libraries etc. which do not change and all data and configuration are kept in separate directories (e.g. app/sonarqube and work/sonarqube).

This is not possible right now because PropsBuilder#detectHome() does not respect the property "sonar.path.home" but uses a path relative to the location of the libraries.

### DIFF
--- a/sonar-application/src/test/java/org/sonar/application/PropsBuilderTest.java
+++ b/sonar-application/src/test/java/org/sonar/application/PropsBuilderTest.java
@@ -45,6 +45,7 @@ public class PropsBuilderTest {
 
   @Before
   public void before() throws IOException {
+    System.setProperty(ProcessProperties.PATH_HOME, "");
     homeDir = temp.newFolder();
   }
 


simonbrandhof:
All the paths to directories containing data are supposed to be configurable:
* `sonar.path.data`
* `sonar.path.logs`
* `sonar.path.temp`
Could you explain why it's not sufficient for your needs ? Thanks.

Argelbargel:
Hi,

as this is was some weeks ago, i am not absolutly sure, but i think without the changes only sonarqube looks for sonar.properties relatively to its sonar-application-<version>.jar (same goes for the extension directory, i think).

We are deploying many apps to our servers an try to separate all "static" data like libraries, executables etc. from data that might change between servers or instances. So we have a folder .../apps/sonarqube where the jar etc. are and a folder .../instances/sonarqube where sonarqube should be run in and store it's data (if not stored in a db), look for configuration and extensions etc.

simonbrandhof:
This is exactly the purpose of the 3 properties I mentioned. In a perfect world they could be merged into a single one.

Argelbargel:
Well, then these settings are not enough ;-) To get sonarqube working in our setup i currently have to copy/link the main sonar-application.jar to my .../instances/sonarqube/lib directory. Otherwise it ignores the sonar.properties in .../instances/sonarqube/conf.  And as `sonar.path.data ` etc. are set in sonar.properties they do not have any effect. 

As sonar.properties contains other configuration (db-connection etc.) it should reside below .../instance/sonarqube/conf in our setup and not in .../apps/sonarqube/conf.

But never mind. For now i've got a working, albeit clunky, solution for my rather peculiar setup so this pr is not really urgent or anything ;-)

btw.: what did i do wrong to break the travis build?

Argelbargel:
As this pr is i bit dated and we're happy with our clunky workaround, i'm closing this...

PR #1062, opened 2016-06-24
SONAR-7798 mssql fix collation now support complex collations

sns-seb:
collations can actually have a combination of case sensitivity and accent sensitivity
also, the might not end with accent sensitivity


eric-duquesnoy:
successfully  tested with CS_AI_KS_WS

PR #1063, opened 2016-06-27
SONAR-7824 Add DB column ACTIVITIES.PROFILE_KEY

teryk:
In the context of SONAR-7789, I add the column ACTIVITIES.PROFILE_KEY to be able to query easily on profile key


server/sonar-server/src/main/java/org/sonar/server/activity/ActivityService.java
```diff
@@ -45,8 +46,15 @@ public void save(Activity activity) {
       .setAction(activity.getAction())
       .setMessage(activity.getMessage())
       .setData(KeyValueFormat.format(activity.getData()))
+      .setProfileKey((String) activity.getData().get("profileKey"))
```
julienlancelot: Why do you get the profileKey from the data of activity ? Why haven't you add a new property in Activity ?

sonar-db/src/main/java/org/sonar/db/activity/ActivityDto.java
```diff
@@ -43,6 +44,15 @@ public String getKey() {
     return key;
   }
 
+  public String getProfileKey() {
+    return profileKey;
+  }
+
+  public ActivityDto setProfileKey(String profileKey) {
```
julienlancelot: You need to add @Nullable

sonar-db/src/main/java/org/sonar/db/activity/ActivityDto.java
```diff
@@ -43,6 +44,15 @@ public String getKey() {
     return key;
   }
 
+  public String getProfileKey() {
```
julienlancelot: You need to add @CheckForNull

sonar-db/src/main/java/org/sonar/db/version/v60/MakeProfileKeyNotNullOnActivities.java
```diff
@@ -0,0 +1,45 @@
...
+
+  public MakeProfileKeyNotNullOnActivities(Database db) {
+    super(db);
+  }
+
+  @Override
+  public void execute(Context context) throws SQLException {
+    context.execute(new AlterColumnsBuilder(getDatabase().getDialect(), TABLE_ACTIVITIES)
+      .updateColumn(newVarcharColumnDefBuilder().setColumnName("profile_key").setLimit(255).setIsNullable(false).build())
+      .build());
```
julienlancelot: Should we really want to have this column not null ? It would means that this table is really only for qprofile, if it's the case the table should be renamed.

sonar-db/src/main/java/org/sonar/db/version/v60/PopulateProfileKeyOfActivities.java
```diff
@@ -0,0 +1,57 @@
...
+public class PopulateProfileKeyOfActivities extends BaseDataChange {
+
+  public PopulateProfileKeyOfActivities(Database db) {
+    super(db);
+  }
+
+  @Override
+  public void execute(Context context) throws SQLException {
+    MassUpdate massUpdate = context.prepareMassUpdate();
+    massUpdate.select("select id, data_field from activities where profile_key is null");
```
julienlancelot: I like it when SQL keyword are in uppercase, so I would write : "SELECT id, data_field FROM activities WHERE profile_key IS NULL".
It's more readable in my opinion.

sonar-db/src/main/java/org/sonar/db/version/v60/PopulateProfileKeyOfActivities.java
```diff
@@ -0,0 +1,57 @@
...
+    massUpdate.rowPluralName("activities");
+    massUpdate.execute(PopulateProfileKeyOfActivities::handle);
+  }
+
+  private static boolean handle(Select.Row row, SqlStatement update) throws SQLException {
+    int id = row.getInt(1);
+    String data = row.getString(2);
+    Map<String, String> fields = KeyValueFormat.parse(data);
+
+    update.setString(1, fields.get("profileKey"));
```
julienlancelot: Is it possible that there's no profileKey ?
I think we should fail fast here, otherwise the next migration that set the profile column to not nullable will fail but it will harder to understand from where.
Failing here will allow to know the id of the row.

server/sonar-server/src/main/java/org/sonar/server/activity/ActivityService.java
julienlancelot: If my understanding is correct, you just need to update the following line : https://github.com/SonarSource/sonarqube/blob/feature/tb/user_updated_at_qp/server/sonar-server/src/main/java/org/sonar/server/qualityprofile/ActiveRuleChange.java#L97.
Replace 
```
activity.setData("profileKey", getKey().qProfile()); 
```
by 
```
activity. setProfileKey(getKey().qProfile());
```

julienlancelot:
Please apply my feedbacks

sonar-db/src/main/java/org/sonar/db/activity/ActivityDto.java
julienlancelot: Forget this comment as the column is not nullable

sonar-db/src/main/java/org/sonar/db/activity/ActivityDto.java
julienlancelot: Forget this comment as the column is not nullable

sonar-db/src/main/java/org/sonar/db/version/v60/MakeProfileKeyNotNullOnActivities.java
teryk: I agree, the table should be renamed. It's already known by the Atlas team

server/sonar-server/src/main/java/org/sonar/server/activity/ActivityService.java
teryk: I will keep the information in the data field for the time being (because of impacts), but I add a new property profileKey in Activity

sonar-db/src/main/java/org/sonar/db/version/v60/PopulateProfileKeyOfActivities.java
teryk: I understand your POV. Do we have a guideline for that?

server/sonar-server/src/main/java/org/sonar/server/activity/ActivityService.java
julienlancelot: I thing this property should really be removed => having the same information twice is very misleading, and if for an unknown reason there are not the same, which one is right ?
There'll be a lot more impact if you leave it like this that if you remove it.

sonar-db/src/main/java/org/sonar/db/version/v60/PopulateProfileKeyOfActivities.java
julienlancelot: I don't know, I'll ask Atlas.

sonar-db/src/main/java/org/sonar/db/version/v60/PopulateProfileKeyOfActivities.java
teryk: good point

server/sonar-server/src/main/java/org/sonar/server/activity/ActivityService.java
teryk: ok

PR #1074, opened 2016-06-30
SONAR-7857 Update to use of Update Center 1.15.1

erichirlemann:
The use of 1.15.1 will bring a couple of bug fixes, it does not require any change on SQ.

The group-id of the UpdateCenter artifact has changed by v1.15.


stas-vilchik:
Already done.

PR #1079, opened 2016-07-04
SONAR-7647 Improve error log when there is a failure in a ProjectBuilder

dbmeneses:


sonar-scanner-engine/src/main/java/org/sonar/batch/scan/ProjectBuildersExecutor.java
```diff
@@ -41,14 +42,18 @@ public ProjectBuildersExecutor() {
...
+        Profiler profiler = Profiler.create(LOG).startInfo("Execute project builders");
+        ProjectBuilderContext context = new ProjectBuilderContext(reactor);
+
+        for (ProjectBuilder projectBuilder : projectBuilders) {
+          projectBuilder.build(context);
+        }
+        profiler.stopInfo();
       }
-      profiler.stopInfo();
+    } catch (Exception e) {
```
henryju: Why don't you make the try..catch surround projectBuilder.build(context); so that you can log exactly what was the ProjectBuilder involved?

henryju: @dbmeneses ping

dbmeneses:
merged

PR #1085, opened 2016-07-06
Activate analysis on SonarQube.com

bellingard:
@simonbrandhof Should be working smoothly :-) Let me know if there are any issues!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,23 @@ language: java
 sudo: false
 install: true
 jdk: oraclejdk8
+addons:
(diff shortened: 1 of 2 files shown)


bellingard:
Also, if it works correctly once merged on SQ repo, then I think that you should generate a token on SQ.com, encrypt it and put it in the Travis YML file (for the moment, the encrypted token is mine).

simonbrandhof:
As discussed this PR won't be applied in the current state as it may conflict with the analysis connected to our internal instance.

PR #1111, opened 2016-07-19
SONAR-7867 persist date of project's snapshot as created_at of ProjectView's snapshot

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/computation/component/ProjectViewAttributes.java
```diff
@@ -19,24 +19,43 @@
...
+  public ProjectViewAttributes(long projectId, @Nullable Long analysisDate) {
     this.projectId = projectId;
+    this.analysisDate = analysisDate;
   }
 
   public long getProjectId() {
     return projectId;
   }
 
+  public Long getAnalysisDate() {
```
simonbrandhof: missing @CheckForNull (or return an Optional)

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistSnapshotsStep.java
```diff
@@ -185,6 +185,16 @@ private SnapshotDto createSnapshot(Component component, Path<SnapshotDtoHolder>
       return snapshotDto;
     }
 
+    private Long computeSnapshotAnalysisDate(Component component, long analysisDate) {
```
simonbrandhof: can return primitive type `long`

simonbrandhof:
LGTM

server/sonar-server/src/main/java/org/sonar/server/computation/step/PersistSnapshotsStep.java
sns-seb: true

server/sonar-server/src/main/java/org/sonar/server/computation/component/ProjectViewAttributes.java
sns-seb: indeed

PR #1115, opened 2016-07-22
Drop project dashboards

stas-vilchik:


simonbrandhof:
Do we want a safeguard when disabling project dashboards (by keeping the list of them) ? If not, then they should be removed from database (tables DASHBOARDS + ACTIVE_DASHBOARDS + WIDGETS + WIDGET_PROPERTIES) and column DASHBOARDS.IS_GLOBAL should be dropped.

stas-vilchik:
@simonbrandhof surely we want, we have a separate ticket for that (cc @teryk)

simonbrandhof:
The extension point org.sonar.api.web.Dashboard allows plugins to define dashboards. The method `setGlobal()` should be kept for backward-compatibility but documentation should mention that project dashboards are ignored (when `setGlobal(true)` is not called).

simonbrandhof:
The startup task `RegisterDashboards` should filter project dashboards.

PR #1120, opened 2016-07-26
Shell scripts improvements

monsieurp:
This PR gathers a couple of improvements for your shell scripts, mainly extraneous calls.


sns-seb:
Hello,
Thanks for your contribution. I didn't know about here strings in bash.
PR merged manually because of amount of conflicts (because it was old).

PR #1134, opened 2016-08-05
Reduce memory consumption when computing line hashes

simonbrandhof:
Reduce memory consumption when computing line hashes

### DIFF
--- a/sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/DefaultInputFile.java
+++ b/sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/DefaultInputFile.java
@@ -24,7 +24,8 @@ import java.io.File;
 import java.io.StringReader;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
-import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
 import org.sonar.api.batch.fs.InputFile;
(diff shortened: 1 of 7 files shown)


sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/DefaultInputFile.java
```diff
@@ -49,7 +50,7 @@
   private int lastValidOffset = -1;
   private String hash;
   private int nonBlankLines;
-  private int[] originalLineOffsets;
+  private List<Integer> originalLineOffsets;
```
sns-seb: to really improve memory consumption, I'd rather track down the reason why we have a ``List`` of ``Integer`` in the first place as a parameter of constructor of class ``Metadata``.

Apparently, this is because of class `org.sonar.api.batch.fs.internal.FileMetadata.LineOffsetCounter`.

This class receives ``char`` and stores them as Integer. This is the initial cause of boxing.

In addition, the internal ``List`` is created with the default size, which might not be pertinent. We might be consuming to much memory here.

But at the end of the day, having an internal array of ``int`` would require a bit of code (dealing with expending the array with a array copy when required) but would not be so hard with a real memory benefit.

henryju: I agree with @sns-seb. Keeping the List<Integer> is not a solution. I have reproduced OOM and it happened before the conversion to int[]. The best solution is to use a custom version of ArrayList that will use an int[] internally. See my P/R: https://github.com/SonarSource/sonarqube/pull/1158

simonbrandhof:
Replaced by https://github.com/SonarSource/sonarqube/pull/1158

PR #1137, opened 2016-08-08
Updated requirements URL

eloy83:
The old URL (redirect) isn't working anymore.


stas-vilchik:
Hi @eloy83,

I don't see any problems, redirect works well. Do I miss something?

eloy83:
It works for me too now. Before I had an error "The page you're looking for does not exist"

PR #1139, opened 2016-08-09
SONAR-7958 udpate uuid path

sns-seb:
shrink column MODULE_UUID_PATH (to 1500) to then be able to add column B_UUID_PATH
populate new column


server/sonar-server/src/test/java/org/sonar/server/component/ws/SearchActionTest.java
```diff
@@ -89,16 +89,16 @@ public void search_json_example() {
         .setName("Project Name")
         .setKey("project-key"));
     ComponentDto module = componentDb.insertComponent(
-      newModuleDto("module-uuid", project)
-        .setName("Module Name")
-        .setKey("module-key"));
-    componentDb.insertComponent(
-      newDirectory(module, "path/to/directoy")
+        newModuleDto("module-uuid", project)
```
teryk: I think you need to update the formatting of IntelliJ

sonar-db/src/test/java/org/sonar/db/component/ComponentTesting.java
```diff
@@ -29,13 +30,13 @@
 
 public class ComponentTesting {
 
-  public static ComponentDto newFileDto(ComponentDto subProjectOrProject) {
-    return newFileDto(subProjectOrProject, Uuids.create());
+  public static ComponentDto newFileDto(ComponentDto subProjectOrProject, @Nullable ComponentDto directory) {
```
teryk: in the vast majority of cases, we don't care about the directory. I suggest to have several constructor method, with and without the directory parameter

server/sonar-server/src/test/java/org/sonar/server/computation/task/projectanalysis/step/ReportPersistComponentsStepTest.java
```diff
@@ -602,6 +624,9 @@ public void persist_components_that_were_previously_removed() {
       .setKey("MODULE_KEY:src/main/java/dir")
       .setEnabled(false);
     ComponentDto removedFile = ComponentTesting.newFileDto(removedModule, "DEFG")
```
teryk: you updated the `ComponentTesting.newFileDto` method now

sonar-db/src/test/java/org/sonar/db/component/ComponentDaoTest.java
```diff
@@ -731,16 +735,16 @@ public void updateBEnabledToFalse() {
...
   }
 
   private Map<String, Object> selectBColumnsForUuid(String uuid) {
     return db.selectFirst(
-        "select b_changed as \"bChanged\", b_copy_component_uuid as \"bCopyComponentUuid\", b_description as \"bDescription\", " +
-          "b_enabled as \"bEnabled\", b_language as \"bLanguage\", b_long_name as \"bLongName\"," +
-          "b_module_uuid as \"bModuleUuid\", b_module_uuid_path as \"bModuleUuidPath\", b_name as \"bName\", " +
-          "b_path as \"bPath\", b_qualifier as \"bQualifier\" " +
-          "from projects where uuid='" + uuid + "'");
+      "select b_changed as \"bChanged\", b_copy_component_uuid as \"bCopyComponentUuid\", b_description as \"bDescription\", " +
```
teryk: suggestion: create a method in ComponentMapper (not in DAO) and call it via `dbSession.getMapper(ComponentMapper.class)`

teryk:
an IT should be added

teryk:
apply feedbacks and LGTM

sonar-db/src/test/java/org/sonar/db/component/ComponentTesting.java
sns-seb: To create a valid File component, we need to know the first parent module (which can be the project) and the first parent (which can be the module/project). Both informations are required and keeping only the methods which emphasis this will avoid creating File component with some incorrect fields.

sonar-db/src/test/java/org/sonar/db/component/ComponentDaoTest.java
sns-seb: one shall not alter production code for test purpose only => I won't change the mapper.

server/sonar-server/src/test/java/org/sonar/server/component/ws/SearchActionTest.java
sns-seb: fixed

sns-seb:
> an IT should be added

it took a while to write it but it's done and allowed to narrow down the scope of the bug

PR #1151, opened 2016-08-17
do not send an overriden language if none is configured locally

t-8ch:
If the browser does not have any language configured
`window.navigator.languages == []` and
`window.navigator.language == ""`.
(at least on FF)
In this case an empty GET parameter is submitted to `/api/l10n/index?locale=`.
This leads to a 400 Bad Request response and no language data.
Therefore the navbar has no UI elements and is unusable, make login impossible.

*Note*: This is untested and probably not the most pretty fix. However it shows the issue.


stas-vilchik:
I've committed that to master with some modifications (53620de6df74d058e73a512da0f2c4143340670d).

Thanks for the contribution!

PR #1157, opened 2016-08-22
Replace CE logs with error message and scanner context in UI

stas-vilchik:


sns-seb:
some usuability feedback made directly, otherwise LGTM

PR #1158, opened 2016-08-22
SONAR-8006 Decrease memory comsumption of FileMetadata

henryju:


sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
```diff
@@ -0,0 +1,140 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.api.batch.fs.internal;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+class IntArrayList {
```
sns-seb: Did you get inspired from somewhere to create this class?
If so, you should not forget to give credit in javadoc (and check licence allows us to do so)

sns-seb: This is class is not unit tested

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * Constructs an empty list with an initial capacity of ten.
+   */
+  public IntArrayList() {
+    this.elementData = DEFAULTCAPACITY_EMPTY_ELEMENTDATA;
+  }
+
+  /**
+   * Trims the capacity of this <tt>IntArrayList</tt> instance to be the
+   * list's current size and return the. An application can use this operation to minimize
```
sns-seb: missing word "return the."

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
```diff
@@ -0,0 +1,140 @@
...
+
+import java.util.Arrays;
+import java.util.Collection;
+
+class IntArrayList {
+
+  /**
+   * Default initial capacity.
+   */
+  private static final int DEFAULT_CAPACITY = 10;
```
sns-seb: I've seen default size be power of two in many collections of Java.
I wonder if this wouldn't be a good practice here as well.

sns-seb: (ArrayList always grows to the nearest power of two, IIRC)

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
henryju: Inspired from JDK's ArrayList.

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+    if (newCapacity - MAX_ARRAY_SIZE > 0) {
+      newCapacity = hugeCapacity(minCapacity);
+    }
+    // minCapacity is usually close to size, so this is a win:
+    elementData = Arrays.copyOf(elementData, newCapacity);
+  }
+
+  private static int hugeCapacity(int minCapacity) {
+    if (minCapacity < 0) {
```
sns-seb: looking at the flow, I don't see how this can occur. Seems to me `minCapacity` is either `size + 1` or `DEFAULT_CAPACITY`
unless argument of `hugeCapacity` at line 115 is incorrect

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
```diff
@@ -0,0 +1,140 @@
...
+   */
+  private static final int[] DEFAULTCAPACITY_EMPTY_ELEMENTDATA = {};
+
+  /**
+   * The maximum size of array to allocate.
+   * Some VMs reserve some header words in an array.
+   * Attempts to allocate larger arrays may result in
+   * OutOfMemoryError: Requested array size exceeds VM limit
+   */
+  private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
```
sns-seb: I think this is useless:
1. initial capacity can't be specified so now one can create a array of MAX_ARRAY_SIZE or above with this class
2. size can only grow one by one, so, when creating a huge array of size X, there will always be an huge array of size X-1 in memory. So, OOM will occur a long time before we reach this limit, ever.

I think you'll be better off letting an OOM occur when it should because we hardly can predict when it will

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * Appends the specified element to the end of this list.
+   *
+   * @param e element to be appended to this list
+   * @return <tt>true</tt> (as specified by {@link Collection#add})
+   */
+  public boolean add(int e) {
+    ensureCapacityInternal(size + 1);
+    elementData[size++] = e;
```
SonarTech: MAJOR Extract this increment or decrement operator into a dedicated statement. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AS881)

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
```diff
@@ -0,0 +1,140 @@
...
+  public int[] trimAndGet() {
+    if (size < elementData.length) {
+      elementData = (size == 0) ? EMPTY_ELEMENTDATA : Arrays.copyOf(elementData, size);
+    }
+    return elementData;
+  }
+
+  private void ensureCapacityInternal(int minCapacity) {
+    if (elementData == DEFAULTCAPACITY_EMPTY_ELEMENTDATA) {
+      minCapacity = Math.max(DEFAULT_CAPACITY, minCapacity);
```
SonarTech: MINOR Introduce a new variable instead of reusing the parameter "minCapacity". [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AS1226)

SonarTech:
SonarQube analysis reported 2 issues
* MAJOR 1 major
* MINOR 1 minor

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
henryju: I agree, BTW it was not possible to unit test.

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    ensureExplicitCapacity(minCapacity);
+  }
+
+  private void ensureExplicitCapacity(int minCapacity) {
+    // overflow-conscious code
```
sns-seb: is this comment still live?

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
```diff
@@ -0,0 +1,120 @@
...
+    if (minCapacity - elementData.length > 0) {
```
sns-seb: why not compare values directly rather than comparing their difference to zero?

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
```diff
@@ -0,0 +1,120 @@
...
+   * Increases the capacity to ensure that it can hold at least the
+   * number of elements specified by the minimum capacity argument.
+   *
+   * @param minCapacity the desired minimum capacity
+   */
+  private void grow(int minCapacity) {
+    // overflow-conscious code
+    int oldCapacity = elementData.length;
+    int newCapacity = oldCapacity + (oldCapacity >> 1);
+    if (newCapacity - minCapacity < 0) {
```
sns-seb: same question here

sonar-plugin-api/src/test/java/org/sonar/api/batch/fs/internal/IntArrayListTest.java
```diff
@@ -0,0 +1,55 @@
...
+    list.add(2);
+    assertThat(list.trimAndGet()).containsExactly(1, 2);
+  }
+
+  @Test
+  public void trimIfNeeded() {
+    IntArrayList list = new IntArrayList();
+    list.add(1);
+    list.add(2);
+    assertThat(list.trimAndGet()).isEqualTo(list.trimAndGet());
```
sns-seb: how does this check that trims occurs only once? (which, I guess from the title, it should do)

henryju: isEqualTo will test that the two objects have the same reference (ie the second call returns the same instance as first call).

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
henryju: This is the value from ArrayList

sonar-plugin-api/src/test/java/org/sonar/api/batch/fs/internal/IntArrayListTest.java
sns-seb: if you mean to test the object reference, I think you mean to use `isSameAs`
this is the code used for `isEqualTo`: https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/util/Objects.java#L37

sonar-plugin-api/src/main/java/org/sonar/api/batch/fs/internal/IntArrayList.java
sns-seb: indeed. I don't remember where I found this power of two optimisation

sonar-ws/src/main/java/org/sonarqube/ws/client/HttpConnector.java
```diff
@@ -101,19 +100,10 @@ private static OkHttpClient buildClient(Builder builder) {
       .supportsTlsExtensions(true)
       .build();
     okHttpClientBuilder.connectionSpecs(asList(tls, ConnectionSpec.CLEARTEXT));
-    okHttpClientBuilder.sslSocketFactory(createSslSocketFactory());
```
sns-seb: I don't see how this change relates to the PR and doesn't look like quality flaw to me

sns-seb:
there is one change which I think should at least be in its own commit (or maybe another PR), otherwise LGTM

PR #1161, opened 2016-08-22
SONAR-7903 add missing primary keys to tables ce_task_input and scanner_context

simonbrandhof:


server/sonar-web/src/main/webapp/WEB-INF/db/migrate/1302_create_table_ce_task_input.rb
```diff
@@ -21,15 +21,15 @@
 #
 # SonarQube 6.1
 #
-class CreateTableCeTaskData < ActiveRecord::Migration
+class CreateTableCeTaskInput < ActiveRecord::Migration
```
sns-seb: now that it's been deployed to Dory, indeed we can merge the two migrations
we couldn't before :)

simonbrandhof: yes I know :)

sns-seb:
this is not fixing an existing DB. So Dory will keep on not having a PK on these tables. (same applies to developer local Databases).
Is is expected?

simonbrandhof:
Yes it is expected. I will send a request to ops team for fixing Dory, and to developers to trash their local dbs.

sns-seb:
then LGTM

PR #1164, opened 2016-08-23
SONAR-7893 end of support of MSSQL 2008

simonbrandhof:
SONAR-7893 end of support of MSSQL 2008


sns-seb:
I haven't check SQ actually fails at startup with MsSQL 2008. I don't have an instance available to me.

I'm assigning the PR to @eric-duquesnoy to test.

PR #1167, opened 2016-08-23
SONAR-7896 set env var TMPDIR to make JRuby use SQ tmp dir for tmp files

sns-seb:
these temp file are especially used when uploading

### DIFF
--- a/sonar-application/src/main/java/org/sonar/application/App.java
+++ b/sonar-application/src/main/java/org/sonar/application/App.java
@@ -92,6 +92,8 @@ public class App implements Stoppable {
       .addJavaOptions(props.nonNullValue(ProcessProperties.WEB_JAVA_ADDITIONAL_OPTS))
       // required for logback tomcat valve
       .setEnvVariable(ProcessProperties.PATH_LOGS, props.nonNullValue(ProcessProperties.PATH_LOGS))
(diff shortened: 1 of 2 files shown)


sonar-application/src/main/java/org/sonar/application/App.java
```diff
@@ -98,6 +98,7 @@ private static JavaCommand createWebServerCommand(Props props, File homeDir) {
       .addJavaOptions(props.nonNullValue(ProcessProperties.WEB_JAVA_ADDITIONAL_OPTS))
       // required for logback tomcat valve
       .setEnvVariable(ProcessProperties.PATH_LOGS, props.nonNullValue(ProcessProperties.PATH_LOGS))
+      .setEnvVariable("TMPDIR", props.nonNullValue(ProcessProperties.PATH_TEMP))
```
simonbrandhof: Don't you think that it deserves to mention JRuby in a comment ? It's quite unusual in a pure Java environment as setting the system property java.io.tmpdir is supposed to be enough.

simonbrandhof:
Strange branch name !

simonbrandhof:
Do not forget to apply this fix to branch-5.6 too.

simonbrandhof:
LGTM

PR #1176, opened 2016-08-25
SONAR-8012 Restore support of IBM JVM in WS client

henryju:


simonbrandhof:
I suggest to upgrade to okkhttp 3.4.1 at the same time. They are frequent fixes about SSL (still I didn't check changes from 3.3.1).

henryju:
According to the [changelog](https://github.com/square/okhttp/blob/master/CHANGELOG.md) there are very few interesting changes for us. Also it seems 3.4.0 came with several regressions. One was fixed in 3.4.1 but the issue tracker has a few issues under investigation.
I would prefer to keep 3.3.1 for the time being. WDYT?

PR #1178, opened 2016-08-25
SONAR-7675 Ability to store server-side internal information

simonbrandhof:


server/sonar-web/src/main/webapp/WEB-INF/db/migrate/1309_create_internal_props_table.rb
```diff
@@ -0,0 +1,36 @@
...
+
+#
+# SonarQube 6.1
+#
+class CreateInternalPropsTable < ActiveRecord::Migration
+
+  def self.up
+    create_table 'internal_props', :id => false do |t|
+      t.column 'kee', :string, :limit => 20, :null => false
+      t.column 'clob_value', :text, :null => true
```
sns-seb: I don't understand why we are taking the same route as table PROPERTIES while we know it's wrong, unless this table is not as generic as its name pretends it is.
Imho, either it is meant for views.xml only and it should be named accordingly, or it should have a varchar column AND a CLOB column for big values

sonar-db/src/main/java/org/sonar/db/version/v61/RemoveViewsDefinitionFromProperties.java
```diff
@@ -0,0 +1,59 @@
...
+public class RemoveViewsDefinitionFromProperties extends BaseDataChange {
+
+  private static final String DEPRECATED_PROPERTY_KEY = "views.def";
+
+  public RemoveViewsDefinitionFromProperties(Database db) {
+    super(db);
+  }
+
+  @Override
+  public void execute(Context context) throws SQLException {
```
sns-seb: this migration is not completely reentrant.

If delete fails, the next time the migration is ran, the insert will be tried again and will fail too.

sonar-db/src/main/java/org/sonar/db/property/InternalPropertyDao.java
```diff
@@ -0,0 +1,84 @@
...
+
+public class InternalPropertyDao implements Dao {
+
+  private final System2 system2;
+
+  public InternalPropertyDao(System2 system2) {
+    this.system2 = system2;
+  }
+
+  public void save(DbSession dbSession, String key, @Nullable String value) {
```
sns-seb: As we discussed recently with @julienlancelot , why not use a mapper here?

Since there is no streaming, there is apparently no specific reason. MyBatis update statement can return the number of updated rows. The mapper method just need to return an `int`.

If we start writing down the full JDBC code, we should discuss where to put the line between using a mapper and not use it.

sns-seb: What is the reason not to simply delete the property when the value is null?
After all, the result of method `selectByKey' would be the same.

In addition, it would save us from issuing two SQL statements when the property doesn't exist. There would be one statement (a DELETE) in all cases.

sns-seb:
closing, will create a new PR

PR #1179, opened 2016-08-25
Handle deprecated key in values and list_definitions WS

julienlancelot:


server/sonar-server/src/test/java/org/sonar/server/settings/ws/ListDefinitionsActionTest.java
```diff
@@ -135,6 +135,25 @@ public void return_settings_definitions_with_minimum_fields() {
...
+  @Test
+  public void return_settings_definitions_with_deprecated_key() {
+    setUserAsSystemAdmin();
+    propertyDefinitions.addComponent(PropertyDefinition
+      .builder("foo")
+      .name("Foo")
+      .deprecatedKey("deprecated")
+      .build());
+
+    ListDefinitionsWsResponse result = newRequest();
```
teryk: suggestion: separate the 3 phases of a test with AAA: Arrange, Act, Assert. Your line `ListDefinitionsWsResponse result = newRequest();` would then be separated from other lines. I find it's an easy to think about tests. It's different for `ExpectedException` though

server/sonar-server/src/main/java/org/sonar/server/settings/ws/ValuesAction.java
```diff
@@ -157,20 +159,27 @@ private void checkAdminPermission(Optional<ComponentDto> component) {
...
+      .map(Setting::createForDefinition)
       .collect(Collectors.toList());
   }
 
+  private Map<String, String> getDeprecatedKeysByValidKeys(Set<String> keys) {
+    return keys.stream()
+      .collect(Collectors.toMap(propertyDefinitions::validKey, Function.identity()));
+  }
+
   private class ValuesResponseBuilder {
```
teryk: can the private class be static?

server/sonar-server/src/main/java/org/sonar/server/settings/ws/ValuesAction.java
```diff
@@ -107,7 +108,8 @@ private ValuesWsResponse doHandle(Request request) {
       Optional<ComponentDto> component = getComponent(dbSession, valuesRequest);
       checkAdminPermission(component);
       Set<String> keys = new HashSet<>(valuesRequest.getKeys());
-      return new ValuesResponseBuilder(loadSettings(dbSession, component, keys), component).build();
+      Map<String, String> deprecatedKeysByValidKeys = getDeprecatedKeysByValidKeys(keys);
```
teryk: do you think it's better to manage the deprecate key mechanism in the SettingsLoader?

teryk: I suggest to rename `deprecatedKeysByValidKeys` by `keysToDisplayMap`. You don't have only deprecated keys, you have mostly valid property key, as key & value of the map.

server/sonar-server/src/main/java/org/sonar/server/settings/ws/ValuesAction.java
```diff
@@ -181,18 +190,18 @@ ValuesWsResponse build() {
...
         valueBuilder.setDefault(setting.isDefault());
         setInherited(setting, valueBuilder);
         setValue(setting, valueBuilder);
       });
     }
 
-    private Settings.Setting.Builder getOrCreateValueBuilder(String propertyKey) {
-      Settings.Setting.Builder valueBuilder = settingsBuilderByKey.get(propertyKey);
+    private Settings.Setting.Builder getOrCreateValueBuilder(String key) {
+      Settings.Setting.Builder valueBuilder = settingsBuilderByKey.get(key);
```
teryk: why do you have to have to check if the setting already exist?

teryk:
check my feedbacks and LGTM

server/sonar-server/src/test/java/org/sonar/server/settings/ws/ListDefinitionsActionTest.java
julienlancelot: ok

server/sonar-server/src/main/java/org/sonar/server/settings/ws/ValuesAction.java
julienlancelot: I completely agree to rename the variable to keysToDisplayMap, as I thought about it but couldn't find a better name.

julienlancelot: To put this logic in SettinsFinder, I don't know... I tend to think that this logic is only for the WS

server/sonar-server/src/main/java/org/sonar/server/settings/ws/ValuesAction.java
julienlancelot: Because the inheritance is handled here : first default settings are created, then global ones, then module ones. 
If a module setting overrides a global setting, then the valueBuilder already exists for that key.

server/sonar-server/src/main/java/org/sonar/server/settings/ws/ValuesAction.java
julienlancelot: Good question, I'll check

SonarTech:

Note: The following issues were found on lines that were not modified in the pull request. Because these issues can't be reported as line comments, they are summarized here:

1. MAJOR [ValuesAction.java#L209](https://github.com/SonarSource/sonarqube/blob/71696ce1a2a5c77fa8b59a7517d7900ee1b2f6f8/server/sonar-server/src/main/java/org/sonar/server/settings/ws/ValuesAction.java#L209): Make "setValue" a "static" method. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AS2325)

PR #1185, opened 2016-08-29
SONAR-7872 support no project version

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/component/ComponentRootBuilder.java
```diff
@@ -73,9 +105,18 @@ private ComponentImpl buildComponent(ScannerReport.Component reportComponent, St
     }
   }
 
+  private ComponentImpl buildProjectComponent(ScannerReport.Component reportComponent, String latestModuleKey) {
+    ComponentImpl.Builder builder = createCommonBuilder(reportComponent, latestModuleKey, latestModuleKey);
+    return builder
+      .setName(nameOfProject(reportComponent, latestModuleKey, projectDtoSupplier))
+      .setReportAttributes(createProjectReportAttributes(reportComponent, builder.getUuid(), analysisSupplier))
```
simonbrandhof: it's quite strange to initialise a builder field from another builder field. Usually builders have only setters. In order to remove the need for adding `ComponentImpl.Builder#getUuid()`, I suggest to extract generation of UUID (`uuidSupplier.apply(componentKey)`) from `createCommonBuilder()`.

sns-seb: It's not that weird, especially considering builder as reusable (eg. protobuf generated builders).
I initially considered moving uuid generation out but it added both complexity (more parameters) and duplication

simonbrandhof:
LGTM

PR #1192, opened 2016-08-30
SONAR-7892 do not support upgrade from before SQ 5.6

sns-seb:


sonar-db/src/main/java/org/sonar/db/version/DatabaseVersion.java
```diff
@@ -37,7 +37,7 @@
    * versions must be previously upgraded to LTS version.
    * Note that the value can't be less than current LTS version.
    */
-  public static final int MIN_UPGRADE_VERSION = 600;
+  public static final int MIN_UPGRADE_VERSION = 1152;
```
simonbrandhof: why not 1150, which is the first one of 5.6 series ?

sns-seb: because 1152 ensures that all migrations of 5.6 has been applied. Seems better to me but I didn't consider 1150 at all. Maybe you have an argument to use it.

It's not 1153 because 1153 has been added with 5.6.1.

simonbrandhof: Installations upgrading from 1150 without having 1152 sounds a very weird corner-case. Probably the upgrade to 6.x would fail. But still, keeping 1152 is ok.

simonbrandhof:
LGTM

PR #1204, opened 2016-09-02
SONAR-7982 Create WS api/settings/encrypt to encrypt a setting value

teryk:


server/sonar-server/src/main/java/org/sonar/server/setting/ws/EncryptAction.java
```diff
@@ -0,0 +1,74 @@
...
+    this.settings = settings;
+  }
+
+  @Override
+  public void define(WebService.NewController context) {
+    WebService.NewAction action = context.createAction("encrypt")
+      .setDescription("Encrypt a setting value.<br>" +
+        "Requires 'Administer System' permission.")
+      .setSince("6.1")
+      .setHandler(this)
```
julienlancelot: As other actions are internal, this one should also be internal

server/sonar-server/src/test/java/org/sonar/server/setting/ws/EncryptActionTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+  WsActionTester ws = new WsActionTester(underTest);
+
+  @Before
+  public void setUp_secret_key() {
+    try {
+      File secretKeyFile = folder.newFile();
+      FileUtils.writeStringToFile(secretKeyFile, "fCVFf/JHRi8Qwu5KLNva7g==");
+
+      encryption = new Encryption(secretKeyFile.getAbsolutePath());
```
julienlancelot: In fact you can use a real Settings object, and then do 
```
settings.getEncryption().setPathToSecretKey(secretKeyFile.getAbsolutePath())
```

julienlancelot:
LGTM

PR #1209, opened 2016-09-05
add flow type checker

stas-vilchik:


vishnuc064:
@ stas-vilchik I have created a project with Es6 and flowtype and angular project. when it run on sonarqube it giving lots of error. is this live now

PR #1210, opened 2016-09-05
SONAR-7968 Return license in api/settings/list_definitions

julienlancelot:
SONAR-7968 Return license in api/settings/list_definitions


teryk:
LGTM

PR #1227, opened 2016-09-09
SONAR-7909 reset sharedMemory at startup

sns-seb:


simonbrandhof:
Tested successfully, the page "System Info" is correctly displayed when disabling Compute Engine and restarting server.
LGTM

PR #1230, opened 2016-09-09
correct wording

stas-vilchik:
adress -> address


stas-vilchik:
@julienlancelot please have a look

PR #1244, opened 2016-09-13
Experimental - test protobuf 3.0.2

simonbrandhof:
Test protobuf 3.0.2


simonbrandhof:
Performance tests of scanner fail. Apparently time to generate report is 10% slower.

PR #1246, opened 2016-09-13
SONAR-8028 Search QP changes with inclusive date and date+time

teryk:


sonar-plugin-api/src/main/java/org/sonar/api/utils/DateUtils.java
```diff
@@ -143,6 +145,51 @@ public static Date parseDateTimeQuietly(@Nullable String s) {
   }
 
   /**
+   *
+   * @param stringDate any string
+   * @throws IllegalArgumentException if stringDate is not a correctly formed date or datetime
+   * @return the datetime, {@code null} if stringDate is null
+   */
+  @CheckForNull
+  public static Date parseDateOrDateTime(@Nullable String stringDate) {
```
julienlancelot: Missing since annotation

sonar-plugin-api/src/test/java/org/sonar/api/server/ws/RequestTest.java
```diff
@@ -214,6 +222,21 @@ public void param_as_date() {
     assertThat(underTest.setParam("a_date", "2014-05-27").paramAsDate("a_date")).isEqualTo(DateUtils.parseDate("2014-05-27"));
   }
 
+  @DataProvider
+  public static Object[][] date_times() {
```
julienlancelot: I'm not really convinced by the usage of DataProvider only for one test...

julienlancelot:
LGTM

sonar-plugin-api/src/test/java/org/sonar/api/server/ws/RequestTest.java
teryk: I understand. It's because I initially tried to work on Request before putting methods in DateUtils. What bothers me is that you need to specify a Runner and you can have only one by test class

teryk: if it was as concise as this, it would be a no brainer (apart from the Runner)
pragmatists_junitparams__parameterised_tests_that_don_t_suck

julienlancelot: Feel free to do a PR on JUnit source code :)

PR #1249, opened 2016-09-14
fix project notifications ui

stas-vilchik:


server/sonar-web/src/main/webapp/WEB-INF/app/controllers/account_controller.rb
```diff
@@ -54,7 +54,7 @@ def update_notifications
       project_notifs.each do |r_id, per_project_notif|
         per_project_notif.each do |dispatch, channels|
           channels.each do |channel, value|
-            Api::Utils.java_facade.saveProperty('notification.' + dispatch + '.' + channel, r_id, current_user.id, 'true')
+            Api::Utils.java_facade.saveProperty('notification.' + dispatch + '.' + channel, Integer(r_id), current_user.id, 'true')
```
julienlancelot: r_id.to_i should have been enough

PR #1253, opened 2016-09-14
Wednesday Hardening!

teryk:
- Http error 400 is sent when trying an incorrect issue transition
- Validate a 400 is sent when WS issues/search cannot parse the date
- revent deletion of the last admin group when no standalone admin user

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/issue/workflow/State.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/issue/workflow/State.java
@@ -21,10 +21,11 @@ package org.sonar.server.issue.workflow;
 
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Sets;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import org.apache.commons.lang.StringUtils;
 import org.sonar.api.issue.Issue;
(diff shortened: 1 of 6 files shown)


server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionMediumTest.java
```diff
@@ -609,14 +615,24 @@ public void display_deprecated_debt_fields() throws Exception {
...
+    WsTester.Result result = wsTester.newGetRequest(API_ENDPOINT, SEARCH_ACTION)
       .setParam("resolved", "false")
       .setParam(WebService.Param.FACETS, "severities")
       .setParam("facetMode", DEPRECATED_FACET_MODE_DEBT)
       .execute();
     result.assertJson(this.getClass(), "display_deprecated_debt_fields.json");
   }
 
+  @Test
+  public void fail_when_invalid_format() throws Exception {
```
julienlancelot: Next time, please do not add tests in class using ServerTester but create another test class that's using EsTester and DbTester

server/sonar-server/src/main/java/org/sonar/server/issue/workflow/State.java
```diff
@@ -43,47 +44,46 @@ public State(String key, Transition[] outTransitions) {
   }
 
   private static void checkDuplications(Transition[] transitions, String stateKey) {
-    Set<String> keys = Sets.newHashSet();
-    for (Transition transition : transitions) {
-      if (keys.contains(transition.key())) {
+    Set<String> keys = new HashSet<>();
+
+    Arrays.stream(transitions)
```
julienlancelot: 👍

server/sonar-server/src/main/java/org/sonar/server/issue/workflow/State.java
```diff
@@ -43,47 +44,46 @@ public State(String key, Transition[] outTransitions) {
...
+      .filter(transition -> !keys.add(transition.key()))
```
julienlancelot: Strange way to collect key in keys...

teryk: I understand. With the lack of accumulator, it's hard to be concise. It's good to use the return value of the `Set.add` method to check for a duplication

PR #1254, opened 2016-09-15
SONAR-7816 Measures on new lines should always be computed but only when SCM is available

julienlancelot:
SONAR-7816 Measures on new lines should always be computed but only when SCM is available

### DIFF
--- a/it/it-tests/src/test/java/it/measureHistory/DifferentialPeriodsTest.java
+++ b/it/it-tests/src/test/java/it/measureHistory/DifferentialPeriodsTest.java
@@ -19,24 +19,35 @@
  */
 package it.measureHistory;
 
+import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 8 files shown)


it/it-tests/src/test/java/it/measureHistory/DifferentialPeriodsTest.java
```diff
@@ -45,24 +56,24 @@
...
+
   @ClassRule
   public static final Orchestrator orchestrator = Category1Suite.ORCHESTRATOR;
 
-  @Before
-  public void cleanUpAnalysisData() {
-    orchestrator.resetData();
+  @BeforeClass
+  public static void createWsClient() throws Exception {
+    CLIENT = newAdminWsClient(orchestrator);
```
julienlancelot: No, as Orchestrator as not started

PR #1255, opened 2016-09-15
SONAR-6407 Missing javadoc on profileImporter and profileExporter

julienlancelot:
SONAR-6407 Missing javadoc on profileImporter and profileExporter

### DIFF
--- a/sonar-plugin-api/src/main/java/org/sonar/api/profiles/ProfileExporter.java
+++ b/sonar-plugin-api/src/main/java/org/sonar/api/profiles/ProfileExporter.java
@@ -28,6 +28,8 @@ import org.sonar.api.batch.ScannerSide;
 import org.sonar.api.server.ServerSide;
 
 /**
+ * Export quality profile rules to a file
+ *
  * @since 2.3
  */
 @ScannerSide
@@ -45,6 +47,11 @@ public abstract class ProfileExporter {
     this.name = name;
   }
 
+  /**
+   * Export activated rule from a quality profile to a writer
+   *
+   * Note that the quality profile can contain some rules from other plugins. It should not fail in this case.
+   */
   public abstract void exportProfile(RulesProfile profile, Writer writer);
 
   public String getKey() {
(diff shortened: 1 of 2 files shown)


sonar-plugin-api/src/main/java/org/sonar/api/profiles/ProfileExporter.java
```diff
@@ -45,6 +47,11 @@ protected ProfileExporter(String key, String name) {
     this.name = name;
   }
 
+  /**
+   * Export activated rule from a quality profile to a writer
+   *
+   * Note that the quality profile can contains some rules from other plugins, and it should not fail in this case.
```
teryk: Note that the quality profile can contain some rules from other plugins. It should not fail in this case.

sonar-plugin-api/src/main/java/org/sonar/api/profiles/ProfileExporter.java
```diff
@@ -28,6 +28,8 @@
 import org.sonar.api.server.ServerSide;
 
 /**
+ * Export activated a quality profile from an external configuration file.
```
teryk: Export quality profile rules to a file

sonar-plugin-api/src/main/java/org/sonar/api/profiles/ProfileExporter.java
```diff
@@ -65,6 +72,10 @@ public final ProfileExporter setName(String s) {
     return this;
   }
 
+  /**
+   * Set the list of language supported by this exporter.
```
teryk: Set the list of languages supported

sonar-plugin-api/src/main/java/org/sonar/api/profiles/ProfileImporter.java
```diff
@@ -27,6 +27,8 @@
 import org.sonar.api.utils.ValidationMessages;
 
 /**
+ * Create a quality profile from an external configuration file.
```
teryk: I don't find a better word, but external configuration file seems to vague. Rules file maybe?

sonar-plugin-api/src/main/java/org/sonar/api/profiles/ProfileImporter.java
```diff
@@ -62,6 +70,10 @@ public final ProfileImporter setName(String s) {
     return this;
   }
 
+  /**
+   * Set the list of language supported by this importer.
```
teryk: Set the list of languages supported

PR #1259, opened 2016-09-15
Harderning Thursday

teryk:


server/sonar-server/src/main/java/org/sonar/server/permission/ws/SearchGlobalPermissionsAction.java
```diff
@@ -78,7 +78,7 @@ private WsSearchGlobalPermissionsResponse buildResponse(DbSession dbSession) {
     Permission.Builder permission = newBuilder();
 
     for (String permissionKey : GlobalPermissions.ALL) {
-      OldPermissionQuery permissionQuery = permissionQuery(permissionKey);
+      PermissionQuery permissionQuery = permissionQuery(permissionKey);
```
julienlancelot: 👍

sonar-plugin-api/src/main/java/org/sonar/api/CoreProperties.java
```diff
@@ -271,7 +271,7 @@
   /**
    * @since 2.11
    */
-  String ORGANISATION = "sonar.organisation";
+  String ORGANIZATION = "sonar.organisation";
```
julienlancelot: It's not possible to update an API like this... You need to deprecate ORGANISATION, and create a new ORGANIZATION parameter

server/sonar-server/src/main/java/org/sonar/server/platform/ServerIdLoader.java
```diff
@@ -45,11 +45,11 @@ public ServerIdLoader(Settings settings, ServerIdGenerator idGenerator) {
       return Optional.absent();
     }
 
-    String organisation = settings.getString(CoreProperties.ORGANISATION);
+    String organization = settings.getString(CoreProperties.ORGANIZATION);
```
julienlancelot: I don't think that we could do that. Licenses are handled by a plugin, I don't know it but if it's using ORGANISATION it will fail.

PR #1263, opened 2016-09-19
SONAR-8031 Allow user login with 2 characters in Realm and IdentityProvider API

julienlancelot:
SONAR-8031 Allow user login with 2 characters in Realm and IdentityProvider API

### DIFF
--- a/sonar-plugin-api/src/main/java/org/sonar/api/server/authentication/UserIdentity.java
+++ b/sonar-plugin-api/src/main/java/org/sonar/api/server/authentication/UserIdentity.java
@@ -19,12 +19,6 @@
  */
 package org.sonar.api.server.authentication;
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.collect.FluentIterable.from;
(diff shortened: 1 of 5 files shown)


sonar-plugin-api/src/main/java/org/sonar/api/server/authentication/UserIdentity.java
```diff
@@ -193,7 +193,7 @@ private static void validateProviderLogin(String providerLogin) {
 
     private static void validateLogin(String login) {
       checkArgument(isNotBlank(login), "User login must not be blank");
-      checkArgument(login.length() <= 255 && login.length() >= 3, "User login size is incorrect (Between 3 and 255 characters)");
+      checkArgument(login.length() <= 255 && login.length() >= 2, "User login size is incorrect (Between 2 and 255 characters)");
```
teryk: I think it should be 2 constants that are called from non API code. So even if we duplicate the check logic we have the same limits

julienlancelot: UserIdentity is in the API, how to call code not from the API ?

PR #1270, opened 2016-09-21
SONAR-7950 WS ce/component fix typo in description

teryk:


teryk:
I doubted it, but I begin to be convinced we should test WS description. This is not the first time we have some typo/error or missing information. Adding an explicit test may prevent some errors

PR #1280, opened 2016-09-29
DB related fixes for 5.6.3

sns-seb:
This fixes:
* SONAR-8148 exception proof FeedFileSourcesBinaryData
* SONAR-8185 support at DAO level of null in FILES_SOURCES.LINES_HASHES

### DIFF
--- a/sonar-db/src/main/java/org/sonar/db/version/v51/FeedFileSourcesBinaryData.java
+++ b/sonar-db/src/main/java/org/sonar/db/version/v51/FeedFileSourcesBinaryData.java
@@ -32,6 +32,7 @@ import org.apache.commons.csv.CSVParser;
 import org.apache.commons.csv.CSVRecord;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
+import org.slf4j.LoggerFactory;
 import org.sonar.api.utils.DateUtils;
 import org.sonar.db.Database;
 import org.sonar.db.protobuf.DbFileSources;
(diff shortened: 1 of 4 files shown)


sonar-db/src/main/java/org/sonar/db/version/v51/FeedFileSourcesBinaryData.java
```diff
@@ -32,6 +32,7 @@
 import org.apache.commons.csv.CSVRecord;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
+import org.slf4j.LoggerFactory;
```
simonbrandhof: `org.sonar.api.utils.log.Logger` should be used.

sonar-db/src/main/java/org/sonar/db/version/v51/FeedFileSourcesBinaryData.java
```diff
@@ -147,7 +148,9 @@ public boolean handle(Select.Row row, SqlStatement update) throws SQLException {
       }
       return FileSourceDto.encodeSourceData(dataBuilder.build());
     } catch (Exception e) {
-      throw new IllegalStateException("Invalid FILE_SOURCES.DATA on row with ID " + fileSourceId + ": " + data, e);
+      LoggerFactory.getLogger(FeedFileSourcesBinaryData.class).error(
```
simonbrandhof: no need for `String.format`, SQ Logger supports `{}` arguments.

sonar-db/src/main/java/org/sonar/db/version/v51/FeedFileSourcesBinaryData.java
```diff
@@ -147,7 +148,9 @@ public boolean handle(Select.Row row, SqlStatement update) throws SQLException {
...
+        String.format("Invalid FILE_SOURCES.DATA on row with ID %s, data will be ignored: %s", fileSourceId, data), e);
```
simonbrandhof: Would it be possible to display the functional key instead of the id, so that user knows which projects must be re-analyzed without requesting DB ?

sonar-db/src/main/java/org/sonar/db/version/v51/FeedFileSourcesBinaryData.java
sns-seb: I'll fix it

sonar-db/src/main/java/org/sonar/db/version/v51/FeedFileSourcesBinaryData.java
sns-seb: it would require to change the select with a join but I want to do as little changes as possible to the code (don't want to introduce a regression)

PR #1281, opened 2016-09-29
SONAR-8151 add property anyone can create organization

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/organization/ws/CreateAction.java
```diff
@@ -21,9 +21,12 @@
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
+import org.sonar.api.CoreProperties;
```
SonarTech: MINOR Remove this unused import 'org.sonar.api.CoreProperties'. [rule](https://dory.sonarsource.com/sonarqube/coding_rules#rule_key=squid%3AUselessImportCheck)

SonarTech:
* MINOR 1 minor

server/sonar-server/src/test/java/org/sonar/server/organization/ws/CreateActionTest.java
```diff
@@ -61,16 +65,19 @@
   @Rule
   public ExpectedException expectedException = ExpectedException.none();
 
+  private Settings settings = new MapSettings()
```
simonbrandhof: I suggest to rely on default settings:
```private Settings settings = new MapSettings(new PropertyDefinitions(CorePropertyDefinitions.all()));```

then the property `ORGANIZATIONS_ANYONE_CAN_CREATE` is overridden in tests only when needed.

File: server/sonar-server/src/test/java/org/sonar/server/organization/ws/CreateActionTest.java [hunk 3]
```diff
@@ -109,14 +116,41 @@ public void verify_response_example() throws URISyntaxException, IOException {
...
     expectedException.expect(ForbiddenException.class);
     expectedException.expectMessage("Insufficient privileges");
 
     executeRequest("name");
   }
 
   @Test
+  public void request_succeeds_if_user_has_SYSTEM_ADMIN_permission_and_logged_in_user_can_not_create_organizations() {
+    giveUserSystemAdminPermission();
+    mockForSuccessfulInsert(SOME_UUID, SOME_DATE);
```
simonbrandhof: As `mockForSuccessfulInsert()` always mocks `System2.now()` with `SOME_DATE`, I suggest to use:
```
private System2 system2 = new TestSystem2().setNow(SOME_DATE);

File: sonar-core/src/main/java/org/sonar/core/config/CorePropertyDefinitions.java [hunk 4]
```diff
@@ -282,6 +285,14 @@ private CorePropertyDefinitions() {
         .category(CoreProperties.CATEGORY_EXCLUSIONS)
         .subCategory(CoreProperties.SUBCATEGORY_DUPLICATIONS_EXCLUSIONS)
         .multiValues(true)
+        .build(),
+
+      // ORGANIZATIONS
+      PropertyDefinition.builder(ORGANIZATIONS_ANYONE_CAN_CREATE)
+        .defaultValue(Boolean.toString(ORGANIZATIONS_ANYONE_CAN_CREATE_DEFAULT_VALUE))
+        .name("Allow any authenticated user to create new Organizations")
```
simonbrandhof: "create new" is a pleonasm :-)  Moreover why an upper-case O for orgs ?

sonar-core/src/main/java/org/sonar/core/config/CorePropertyDefinitions.java
```diff
@@ -282,6 +285,14 @@ private CorePropertyDefinitions() {
         .category(CoreProperties.CATEGORY_EXCLUSIONS)
         .subCategory(CoreProperties.SUBCATEGORY_DUPLICATIONS_EXCLUSIONS)
         .multiValues(true)
+        .build(),
+
+      // ORGANIZATIONS
+      PropertyDefinition.builder(ORGANIZATIONS_ANYONE_CAN_CREATE)
+        .defaultValue(Boolean.toString(ORGANIZATIONS_ANYONE_CAN_CREATE_DEFAULT_VALUE))
```
simonbrandhof: for readability I would recommend to put default value after the name. Moreover why using a constant ? For the same readability reason, it's better to have a full understanding of property when reading this block.

server/sonar-server/src/test/java/org/sonar/server/organization/ws/CreateActionTest.java
sns-seb: Your suggestion implies that the unit test care about the default value mechanism. I don't think it does.

That's an implementation detail of `Settings`.

I don't see any benefit to rely on it, especially considering that changing the default value in the definition would break the UT even though it's totally unrelated. This would also make UT's code more obscure (here, it's explicite default value for all tests methods is `false`).

Property `ORGANIZATIONS_ANYONE_CAN_CREATE` is already overridden in tests when needed.

server/sonar-server/src/test/java/org/sonar/server/organization/ws/CreateActionTest.java
sns-seb: I wonder...

your suggestion is not a 100% replacement. I won't set now in field initialization but only in `mockForSuccessfulInsert`. This way, any unexpected call will fail and that's good :)
It will also keep clear in test method that the date of the organization is the one provided at the beginning of the test.

sonar-core/src/main/java/org/sonar/core/config/CorePropertyDefinitions.java
sns-seb: constant avoids duplication the name of the property in code when it is read.

simonbrandhof: I don't get your point. My feedback is about having `.builder(ORGANIZATIONS_ANYONE_CAN_CREATE ).defaultValue(Boolean.toString(false))`.

sns-seb: Ok, I didn't get you were talking about the constant for the default value and not the one of the name.
I'll remove the constant for the default value

PR #1285, opened 2016-10-02
Detect encoding of a file from BOM

Godin:
This relates to https://jira.sonarsource.com/browse/SONAR-6337


sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/filesystem/InputFileBuilder.java
```diff
@@ -93,19 +98,52 @@ DefaultInputFile create(File file) {
   DefaultInputFile completeAndComputeMetadata(DefaultInputFile inputFile, InputFile.Type type) {
     inputFile.setType(type);
     inputFile.setModuleBaseDir(fs.baseDir().toPath());
-    inputFile.setCharset(fs.encoding());
+    Charset charset = detectCharset(inputFile.file(), fs.encoding());
```
henryju: Charset detection should be done after language detection. This would prevent opening every file, even if the file will be skipped.

sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/filesystem/InputFileBuilder.java
```diff
@@ -93,19 +98,52 @@ DefaultInputFile create(File file) {
...
+    inputFile.setCharset(charset);
```
henryju: This is more a question for PO/support but if we start having files with different encoding, we should probably log in DEBUG what is the detected charset. @bellingard WDYT? To avoid too verbose logs, we could maybe log in a single line detected language + charset?

Godin: FWIW IMO logs are already crazily verbose

sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/filesystem/InputFileBuilder.java
Godin: @henryju done

PR #1288, opened 2016-10-04
Improve exception handling

simonbrandhof:


server/sonar-process-monitor/src/main/java/org/sonar/process/monitor/Monitor.java
```diff
@@ -282,8 +282,12 @@ private RestartorThread() {
 
     @Override
     public void run() {
-      stopProcesses();
-      startProcesses();
+      try {
+        stopProcesses();
```
sns-seb: no need to wrap both `stopProcesses()` and `startProcesses()`, only the latter can throw `InterruptedException`
also, a comment in the `catch` would be wise (that we just ignore the interrupt for now)

server/sonar-process-monitor/src/main/java/org/sonar/process/monitor/WatcherThread.java
```diff
@@ -45,25 +45,29 @@
...
-        // terminate all other processes, but in another thread
-        monitor.stopAsync();
+        askForStop();
         stopped = true;
-      } catch (InterruptedException ignored) {
-        // continue to watch process
       }
+    } catch (InterruptedException ignored) {
+      askForStop();
+      Thread.currentThread().interrupt();
```
sns-seb: call to `Thread.currentThread().interrupt()` is useless since we dont free the thread and continue to loop
new code doesn't ensure anymore the watched thread is stopped
old code is correct

sns-seb: I even wonder if we don't end up in an infinite loop...

sns-seb: another option is to just drop the `while` but it still doesn't enforce the watched thread is stopped

server/sonar-process-monitor/src/main/java/org/sonar/process/monitor/Monitor.java
```diff
@@ -282,10 +282,12 @@ private RestartorThread() {
 
     @Override
     public void run() {
+      stopProcesses();
       try {
-        stopProcesses();
         startProcesses();
       } catch (InterruptedException e) {
+        // Startup was interrupted. Processes are supposed to be stopped.
```
sns-seb: that's not exactly true. Stop in case of failure inside `startProcesses` is asynchronous as handled by another thread.
So, at this point of the program, there is no garantie the processes are stopped (they are most likely not)

server/sonar-process-monitor/src/main/java/org/sonar/process/monitor/Monitor.java
```diff
@@ -282,10 +282,12 @@ private RestartorThread() {
...
+        // Cleaning the interruption state
```
sns-seb: `interrupt` is not cleaning the state, it's actually restoring it, ie. propagating it when applicable

PR #1293, opened 2016-10-07
SONAR-6724 : support analyzing root pom in multi-modules projects

lbndev:
This PR allows processing of files in aggregator modules.
When analyzing a multi-modules maven project, this allows the root pom to be indexed and analyzed.
It should also be beneficial to other project types (Gradle was mentioned in SONAR-6724).
This only applies to files, not folders : SONARPLUGINS-2295 introduced a warning for existing folders referenced in property sonar.sources of aggregator modules : it will still be displayed, and these folders will still be ignored.

### DIFF
--- a/sonar-scanner-engine/src/main/java/org/sonar/batch/scan/ProjectReactorBuilder.java
+++ b/sonar-scanner-engine/src/main/java/org/sonar/batch/scan/ProjectReactorBuilder.java
@@ -19,6 +19,8 @@
  */
 package org.sonar.batch.scan;
 
+import static java.util.stream.Collectors.*;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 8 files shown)


henryju:
Hi! Thanks for trying to make this very old topic move forward. What is the result on SonarQube side? How does the UI react when a project/module contains a mix of files and sub modules?
Also regarding measure aggregation, what is the behavior? For example, is ncloc properly aggregated when a project/module contains both source files and sub modules?

lbndev:
Hi,

On SonarQube side, everything looks fine to me :
- on the homepage, project size contains lines for java and xml (the only XML files I have in my test project are my POMs). In "code" view, my pom.xml appears as a file (with a pin icon) below the modules (with a folder icon) and its line count is properly added to the total lines count of the project. In Measures/Size view, my root pom.xml is present and shows different values for "lines" and "lines of code" (=ncloc ?). I guess all of this is handled automatically by the xml plugin.
 - in the issues page, I can find issues for my root pom as well as for other POMs and java files. I can use filters ("by file" filter contains my root pom.xml, "by module" filter contains the root project's name, "by rule" filter contains rules violated only by my parent pom, ...)
 - technical debt gets counted as usual and adds up with debt from child modules

henryju:
FYI we created a MMF to progress on this topic: https://jira.sonarsource.com/browse/MMF-509

lbndev:
Nice. Hopefully this will make it into SonarQube next major release.

Still, can I suggest that my PR be merged as a bugfix in the current LTS branch ?
It will be superseded by the improvements from MMF-509, but until then allow root poms to be analyzed. This is especially important for contexts where only LTS releases will be used (otherwise this bug will persist until next LTS - one every 18 months, right ? so some time around 12/2017 ?).

Thank you for your understanding

henryju:
This is unlikely we will merge that change in LTS. It has possibly many side effects regarding configuration. The point of LTS is stability, so no functional changes, only bug fixes.

lbndev:
Any news on this ?
Thanks

lbndev:
I ported those changes to branch-6.2, see PR #1613

janos-ss:
I'm closing this PR, because your other PR #1613 supersedes it.

PR #1294, opened 2016-10-10
SONAR-8221 Create WS api/components/search_projects projects are searched in DB

teryk:


server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
```diff
@@ -0,0 +1,132 @@
...
+  @Override
+  public void handle(Request httpRequest, Response httpResponse) throws Exception {
+    SearchProjectsWsResponse response = doHandle(toRequest(httpRequest));
+
+    writeProtobuf(response, httpRequest, httpResponse);
+  }
+
+  private SearchProjectsWsResponse doHandle(SearchProjectsRequest request) {
+    DbSession dbSession = dbClient.openSession(false);
+    try {
```
sns-seb: there is no reason not to use try-with-resource. Even Jacoco coverage bug isn't good enough of a reason.

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
```diff
@@ -0,0 +1,132 @@
...
+
+      return buildResponse(request, searchResult);
+
+    } finally {
+      dbClient.closeSession(dbSession);
+    }
+  }
+
+  private SearchResult searchProjects(DbSession dbSession, SearchProjectsRequest request) {
+    List<ComponentDto> allProjects = dbClient.componentDao().selectProjects(dbSession);
```
sns-seb: ouch! no search at all in DB? no limit, nothing?
This is a performance no go

sns-seb: in addition, I don't see anything done in Java which can't be implemented in DB. I don't understand why this is done like this.

Is this implementation only a draft?

Even so, I don't think it's a good idea to let such a performance issue go into master and will veto.

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
```diff
@@ -0,0 +1,132 @@
...
+      .stream()
+      .sorted((c1, c2) -> c1.name().compareTo(c2.name()))
+      .skip(offset(request.getPage(), request.getPageSize()))
+      .limit(request.getPageSize())
+      .collect(Collectors.toList());
+
+    return new SearchResult(projects, allProjects.size());
+  }
+
+  private static SearchProjectsRequest toRequest(Request httpRequest) {
```
sns-seb: `SearchProjectsRequest` is hardly useful when used only internally to class `SearchProjectsAction`. Why not use request directly?

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
```diff
@@ -0,0 +1,132 @@
...
+
+  private static SearchProjectsWsResponse buildResponse(SearchProjectsRequest request, SearchResult searchResult) {
+    SearchProjectsWsResponse.Builder response = SearchProjectsWsResponse.newBuilder();
+
+    response.setPaging(Common.Paging.newBuilder()
+      .setPageIndex(request.getPage())
+      .setPageSize(request.getPageSize())
+      .setTotal(searchResult.total));
+
+    Function<ComponentDto, Component.Builder> dbToWsComponent = dbComponent -> Component.newBuilder()
```
sns-seb: rather than a lambda assigned to a variable, use a method and a method reference
it will make `buildResponse` shorter, avoids creating a lamba object and assigning it to a variable.
methods can be documented with javadoc if need be

sns-seb: `Component.newBuilder()` should be created outside the lamba in order to be reused (don't forget to call `clear()` on it) and save useless creation of multiple builder instances (one for each entry in `searchResult`).
Builder are designed to be reused in protobuf

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
```diff
@@ -0,0 +1,132 @@
...
+    searchResult.projects
+      .stream()
+      .map(dbToWsComponent)
+      .sorted((component1, component2) -> component1.getName().compareTo(component2.getName()))
+      .forEach(response::addComponents);
+
+    return response.build();
+  }
+
+  private static class SearchResult {
```
sns-seb: that class should be called `SearchResults` as it represents multiple results.

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
```diff
@@ -0,0 +1,132 @@
...
+
+    Function<ComponentDto, Component.Builder> dbToWsComponent = dbComponent -> Component.newBuilder()
+      .setId(dbComponent.uuid())
+      .setKey(dbComponent.key())
+      .setName(dbComponent.name());
+
+    searchResult.projects
+      .stream()
+      .map(dbToWsComponent)
+      .sorted((component1, component2) -> component1.getName().compareTo(component2.getName()))
```
sns-seb: you can write this the Java8 way: `.sorted(Comparator.comparing(Component::getName))`

(even shorter with static import)

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
```diff
@@ -0,0 +1,132 @@
...
+    } finally {
+      dbClient.closeSession(dbSession);
+    }
+  }
+
+  private SearchResult searchProjects(DbSession dbSession, SearchProjectsRequest request) {
+    List<ComponentDto> allProjects = dbClient.componentDao().selectProjects(dbSession);
+    List<ComponentDto> projects = allProjects
+      .stream()
+      .sorted((c1, c2) -> c1.name().compareTo(c2.name()))
```
sns-seb: you can write this the Java8 way: `.sorted(Comparator.comparing(ComponentDto::name()))`

(but this should rather be done in SQL)

server/sonar-server/src/test/java/org/sonar/server/component/ws/SearchProjectsActionTest.java
```diff
@@ -0,0 +1,155 @@
...
+    componentDb.insertComponents(newProjectDto().setName("Maven"), newProjectDto().setName("Apache"), newProjectDto().setName("Guava"));
+
+    SearchProjectsWsResponse result = call(request);
+
+    assertThat(result.getComponentsList()).extracting(Component::getName)
+      .containsExactly("Apache", "Guava", "Maven");
+  }
+
+  @Test
+  public void paginate_result() {
```
sns-seb: calling a index and/or a page which returns 0 results is not tested

server/sonar-server/src/test/java/org/sonar/server/component/ws/SearchProjectsActionTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+    SearchProjectsWsResponse result = call(request);
+
+    assertThat(result.getComponentsCount()).isEqualTo(1);
+    assertThat(result.getComponents(0).getName()).isEqualTo("SonarQube");
+  }
+
+  @Test
+  public void fail_if_page_size_greater_than_500() {
+    expectedException.expect(IllegalArgumentException.class);
```
sns-seb: as this is not testing the error message, there is no guarantee the IAE is throw because of page size

server/sonar-server/src/test/java/org/sonar/server/component/ws/SearchProjectsActionTest.java
```diff
@@ -0,0 +1,155 @@
...
+    TestRequest httpRequest = ws.newRequest()
+      .setMediaType(MediaTypes.PROTOBUF);
+
+    httpRequest.setParam(Param.PAGE, String.valueOf(wsRequest.getPage()));
+    httpRequest.setParam(Param.PAGE_SIZE, String.valueOf(wsRequest.getPageSize()));
+
+    try {
+      return SearchProjectsWsResponse.parseFrom(httpRequest.execute().getInputStream());
+    } catch (IOException e) {
+      throw Throwables.propagate(e);
```
sns-seb: why this try and propagate ?
test method can declare thrown exceptions.
In case of error, the cause will be burried by this wrapping added to the one of the test framework
not worth it, imho

sonar-ws/src/main/java/org/sonarqube/ws/client/component/SearchProjectsRequest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public SearchProjectsRequest build() {
+      if (page == null) {
+        page = 1;
+      }
+      if (pageSize == null) {
+        pageSize = DEFAULT_PAGE_SIZE;
+      }
+
+      checkArgument(pageSize <= MAX_PAGE_SIZE);
```
sns-seb: put a message so that user know what the cause of the error was
in addition, it allows writing accurate UTs (see my other comment)

sonar-ws/src/test/java/org/sonarqube/ws/client/component/SearchProjectsRequestTest.java
```diff
@@ -0,0 +1,58 @@
...
+  public void handle_paging_limit_values() {
+    SearchProjectsRequest result = underTest.setPageSize(500).build();
+
+    assertThat(result.getPage()).isEqualTo(1);
+    assertThat(result.getPageSize()).isEqualTo(500);
+  }
+
+  @Test
+  public void fail_if_page_size_greater_than_500() {
+    expectedException.expect(IllegalArgumentException.class);
```
sns-seb: without testing the message, we don't know if IAE was caused by the page size check or anything else

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
teryk: yes it's a draft, I'm working on the ES index right now

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
teryk: `SearchProjectsRequest` will be used in the WS client and it's a convention in all WS

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
sns-seb: so, this review happens too early then?
Let me know when index search is implemented

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
teryk: with your 2nd refactoring, I'd need to pass the Component.Builder as a parameter. But I could extract in a private static class ResponseBuilder

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
sns-seb: then PR should include WSClient code.
there is no point pushing to master maybe-for-the-futur or pure-convention code which is in practice only overhead

server/sonar-server/src/test/java/org/sonar/server/component/ws/SearchProjectsActionTest.java
teryk: to avoid to add IOException in each of my tests. It doesn't really make the error message less visible

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
sns-seb: A ResponseBuilder is over engineered
KISS man!

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
sns-seb: no pb.
Early feedback is good.
I assumed the PR was ready for merge.

server/sonar-server/src/test/java/org/sonar/server/component/ws/SearchProjectsActionTest.java
sns-seb: I prefer the IOException declaration to the extra code which, imho, adds noise as it has no purpose toward what we are testing

sns-seb: at the end of the day, it's a matter of taste

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
teryk: as the pagination will be done in ES, I don't want to bother implement it in DB for this iteration

PR #1301, opened 2016-10-13
Fix core.properties by removing merging conflicts

henryju:


sonar-core/src/main/resources/org/sonar/l10n/core.properties
```diff
@@ -4091,3 +3135,4 @@ my_activity.my_favorite_projects=My Favorite Projects
 my_activity.no_favorite_projects=You don't have any favorite projects yet.
 my_activity.no_favorite_projects.engagement=Discover and favorite projects you are interested in to have a quick access to them.
 my_activity.explore_projects=Explore Projects
+
```
julienlancelot: It GitHub displaying the carriage return or is it a strange character ? 
I see only a space in the IDEA

henryju: Indeed there was an extra space.

PR #1305, opened 2016-10-15
Do not throw UnsupportedOperationException when httpRequest.getSession(false) is called

evernat:
When httpRequest.getSession(false) is called, the callers does not ask to create a http session.
(javadoc of HttpServletRequest: "If create is false and the request has no valid HttpSession, this method returns null.")

So throwing UnsupportedOperationException is not needed in this case and returning null is enough.

This will fix a [blocking issue](https://github.com/javamelody/sonar-javamelody/issues/4) in the [Sonar JavaMelody plugin](https://github.com/javamelody/javamelody/wiki/UserGuide#sonar-plugin).

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/platform/web/RootFilter.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/platform/web/RootFilter.java
@@ -147,6 +147,9 @@ public class RootFilter implements Filter {
 
     @Override
     public HttpSession getSession(boolean create) {
+      if (!create) {
+        return null;
+      }
       throw notSupported();
     }
 


julienlancelot:
@evernat Why do the Sonar JavaMeledoy plugin needs to get a HTTP session ? 
As Sonarqube is not allowing usage of HTTP session, it's a bad practice to not throw an exception.

evernat:
@julienlancelot 
No, Sonar JavaMelody plugin does not need to get a HTTP session.
In fact, the plugin calls httpRequest.getSession(false) to see if there is a session and if there is no session then no problem.
So the plugin does not need Sonarqube to allow usage of HTTP session, but the plugin needs that Sonarqube says there is no session.
The PR will not change the fact that Sonarqube refuses to create HTTP sessions.

julienlancelot:
Thanks @evernat, this PR has been merged !

PR #1306, opened 2016-10-16
SONAR-8291 Ability to mark a web service parameter as "internal"

simonbrandhof:


server/sonar-web/src/main/js/apps/web-api/components/Params.js
```diff
@@ -19,6 +19,8 @@
  */
 import React from 'react';
 
+import InternalBadge from './InternalBadge';
```
julienlancelot: Wouhou, we have a new JS developer !! (cc @teryk @stas-vilchik )

PR #1315, opened 2016-10-19
Added support for mariadb

leoleovich:
I have an exception with Mariadb about 
```
org.sonar.api.utils.MessageException: Unsupported mysql version: 5.5. Minimal supported version is 5.6.
```

This should help...

### DIFF
--- a/sonar-db/src/main/java/org/sonar/db/DatabaseChecker.java
+++ b/sonar-db/src/main/java/org/sonar/db/DatabaseChecker.java
@@ -45,6 +45,7 @@ public class DatabaseChecker implements Startable {
     // MsSQL 2014 is 12.x
     // https://support.microsoft.com/en-us/kb/321185
     MsSql.ID, Version.create(11, 0, 0),
+    MsSql.ID, Version.create(10, 0, 0),
     MySql.ID, Version.create(5, 6, 0),
     Oracle.ID, Version.create(11, 0, 0),
     PostgreSql.ID, Version.create(8, 0, 0));


leoleovich:
Actually when I just upgraded from mysql to mariadb I did not restart sonar. It worked just fine. The problem is only on validation side (as far as I can see)

simonbrandhof:
Hi, thanks for the pull request but MariaDB won't be supported. It's not a problem of codebase, but about effort to maintain and to validate.
See https://groups.google.com/forum/#!msg/sonarqube/GmbtZ_qf1ME/vCHk4sZVAwAJ for more details.
Regards

Exordian:
@simonbrandhof is replacing the error with some sort of warning an option? almost all other software handles it like this (confluence, jira, teamcity, ...) if you're running a non-optimal configuration, and you won't get any support for this. but i would actually appreciate not to be forced to some sort of software if it may work, and i know what i'm doing

simonbrandhof:
Hi @Exordian,
Firstly mimicking other softwares is not a valid reason, whatever their (good) quality.
About the argument to drop support in case of non-optimal configuration:

* mariadb is "not tested" but not "non-optimal". We don't make any judgement.
* any potential issues related to mariadb will give a wrong impression about SonarQube to users. We deal with this kind of problem by "fail-fasting".
* it's impossible to drop support for a configuration allowed by the software. Any questions on the Google group and Stackoverflow require effort to answer. This effort is reinforced for our commercial support team.

Finally logs are read by most of users only in case of errors. Warnings are almost useless. Decreasing the log level and removing fail-fast would be equivalent to support the feature.

Quix0r:
So SonarQube is no choice because of missing MariaDB? I have the same issue here, same message about "out-dated" (MariaDB 10) MySQL. Please reconsider your position towards MariaDB, there is a great connector and other great stuff to minimize maintenance support (such as database abstraction classes or even the JPA with JPQL).

sonar-db/src/main/java/org/sonar/db/DatabaseChecker.java
```diff
@@ -45,6 +45,7 @@
     // MsSQL 2014 is 12.x
     // https://support.microsoft.com/en-us/kb/321185
     MsSql.ID, Version.create(11, 0, 0),
+    MsSql.ID, Version.create(10, 0, 0),
```
Quix0r: `MsSql.ID` ? Maybe `MySQL.ID` is right?

StrangeWill:
This is a bit rough being as some distributions such as Debian 9 has made MariaDB their default and only MySQL provider... I've disabled SonarQube in our environment for now over this and have to basically consider running older distributions for SonarQube alone...

PR #1346, opened 2016-10-28
SONAR-5857 It should only be possible to use rating metrics defined in core for quality gates

julienlancelot:
SONAR-5857 It should only be possible to use rating metrics defined in core for quality gates


julienlancelot:
Manually merged

PR #1352, opened 2016-11-02
SONAR-8260 fix conflict between project and organization parameters

simonbrandhof:


server/sonar-server/src/main/java/org/sonar/server/permission/ws/AddUserAction.java
```diff
@@ -80,7 +81,11 @@ public void handle(Request request, Response response) throws Exception {
     try (DbSession dbSession = dbClient.openSession(false)) {
       UserId user = support.findUser(dbSession, request.mandatoryParam(PARAM_USER_LOGIN));
       Optional<ProjectId> projectId = support.findProject(dbSession, request);
-      OrganizationDto org = support.findOrganization(dbSession, request.param(PARAM_ORGANIZATION_KEY));
+      String organizationKey = request.param(PARAM_ORGANIZATION_KEY);
+      if (projectId.isPresent()) {
```
sns-seb: shouldn't this be

```java
checkArgument(!projectId.isPresent() || organizationKey == null, "Organization must not be set when project is set.");
```

simonbrandhof: yes it could. Proposal applied.

PR #1354, opened 2016-11-03
SONAR-6912 Prevent locking out administrator

simonbrandhof:


server/sonar-server/src/main/java/org/sonar/server/usergroups/ws/RemoveUserAction.java
```diff
@@ -83,6 +86,17 @@ public void handle(Request request, Response response) throws Exception {
     }
   }
 
+  /**
+   * Verify that there are still users with admin global permission if user is removed from the group.
+   */
+  private void checkNotTryingToDeleteLastAdmin(DbSession dbSession, GroupId group, UserDto user) {
```
sns-seb: since the method raises an exception (rather than returning a boolean which is usually implied by the method name check*), you could name it `ensureLastAdminIsNotRemoved` or something like that

server/sonar-server/src/test/java/org/sonar/server/usergroups/ws/RemoveUserActionTest.java
```diff
@@ -276,4 +310,13 @@ private boolean isUserInGroup(UserDto userDto, GroupDto groupDto) {
       .anyMatch(dto -> dto.getLogin().equals(userDto.getLogin()));
   }
 
+  private UserDto insertAnAdministratorInDefaultOrganization() {
+    return insertAnAdministrator(db.getDefaultOrganization());
+  }
+
+  private UserDto insertAnAdministrator(OrganizationDto org) {
```
sns-seb: you could add method `insertAdminByUserPermission` to `UserDbTester` in the same flavour there is `insertRootByUserPermission`

PR #1397, opened 2016-11-16
SONAR-8383 download all log files from System Info page

simonbrandhof:


server/sonar-server/src/main/java/org/sonar/ce/log/CeProcessLogging.java
```diff
@@ -28,7 +29,7 @@
 public class CeProcessLogging extends ServerProcessLogging {
 
   public CeProcessLogging() {
-    super("ce", "%X{ceTaskUuid}");
+    super(LogbackHelper.LogProcess.CE, "%X{ceTaskUuid}");
```
sns-seb: conflicting change with my PR, not so hard to fix, but I'd rather let you handle it since I have 5 branches with that change :)

server/sonar-process/src/main/java/org/sonar/process/LogbackHelper.java
```diff
@@ -52,6 +52,28 @@
  */
 public class LogbackHelper {
 
+  public enum LogProcess {
```
sns-seb: check with Java team, but I think classes should start with constants and code rather than embedded types

or maybe it's just a matter of taste to quickly see content of class rather than subtypes

(in this class, all other sub types are at the end)

simonbrandhof: indeed. Let's wait for SonarQube analysis !

sns-seb:
When log file is empty or missing (eg. in cluster mode) we get a completely blank page. It's hard to make a difference between this case and some error. Displaying a hardcoded content (such as "no logs") or using an HTTP status could be nice

server/sonar-server/src/main/java/org/sonar/ce/log/CeProcessLogging.java
simonbrandhof: it's even better after the merge : using existing ProcessId is far better than defining new enum LogProcess.

PR #1398, opened 2016-11-16
SONAR-8177 do not display messages of technical error in web services

simonbrandhof:


sns-seb:
second commit title is too long

server/sonar-server/src/main/java/org/sonar/server/ws/WebServiceEngine.java
```diff
@@ -120,7 +120,9 @@ public void execute(Request request, Response response) {
         return;
       }
       LOGGER.error("Fail to process request " + request, e);
-      sendErrors(response, 500, new Errors().add(Message.of(e.getMessage())));
+      // Sending exception message into response is a vulnerability. Error must be
+      // displayed only in logs.
+      sendErrors(response, 500, new Errors().add(Message.of("error_occurred")));
```
sns-seb: ok, that's "bourrin" and will affect all errors and not only ES ones, but that's good

PR #1401, opened 2016-11-17
SONAR-8363 Mark just created projects as favorite when there's a project creator permission

teryk:


server/sonar-server/src/main/java/org/sonar/server/permission/PermissionTemplateService.java
```diff
@@ -155,6 +156,13 @@ public void applyDefault(DbSession dbSession, ComponentDto component, @Nullable
     indexProjectPermissions(dbSession, asList(component.uuid()));
   }
 
+  public boolean hasProjectCreator(DbSession dbSession, ComponentDto component) {
```
simonbrandhof: A javadoc should be added to describe the goal. It's not straightforward to understand.

server/sonar-server/src/main/java/org/sonar/server/permission/PermissionTemplateService.java
```diff
@@ -203,7 +211,7 @@ private void copyPermissions(DbSession dbSession, PermissionTemplateDto template
    * template for the component qualifier.
    */
   @CheckForNull
-  private PermissionTemplateDto findDefaultTemplate(DbSession dbSession, ComponentDto component) {
+  public PermissionTemplateDto findDefaultTemplate(DbSession dbSession, ComponentDto component) {
```
simonbrandhof: why changing to `public` ?

server/sonar-server/src/test/java/org/sonar/server/computation/queue/ReportSubmitterTest.java
```diff
@@ -113,6 +113,7 @@ public void provision_project_if_does_not_exist() throws Exception {
     when(componentService.create(any(DbSession.class), any(NewComponent.class))).thenReturn(createdProject);
     when(permissionTemplateService.wouldUserHavePermissionWithDefaultTemplate(any(DbSession.class), anyLong(), eq(SCAN_EXECUTION), anyString(), eq(PROJECT_KEY), eq(Qualifiers.PROJECT)))
       .thenReturn(true);
+    when(permissionTemplateService.hasProjectCreator(any(DbSession.class), any(ComponentDto.class))).thenReturn(true);
```
simonbrandhof: A unit test is missing to verify that favorite is not created when the permission template does not include permissions for project creator.

simonbrandhof: In the same way an assertion is missing to verify that favorite is not created when submitting a report on an existing project.

server/sonar-server/src/main/java/org/sonar/server/permission/PermissionTemplateService.java
teryk: my bad, because of first implementation

server/sonar-server/src/main/java/org/sonar/server/permission/PermissionTemplateService.java
teryk: what do you think of `hasDefaultTemplateAProjectCreatorPermission` as a method name?

server/sonar-server/src/test/java/org/sonar/server/computation/queue/ReportSubmitterTest.java
teryk: it's indirectly tested when you check there's no interaction with the mock

PR #1408, opened 2016-11-22
Drop usage of the Selenese implementation provided by Orchestrator

simonbrandhof:


julienlancelot:
Very good job @simonbrandhof  !

PR #1412, opened 2016-11-25
SONAR-8089 Add missing condition on snapshots.islast=true

julienlancelot:


sonar-db/src/main/resources/org/sonar/db/measure/MeasureMapper.xml
```diff
@@ -104,7 +104,7 @@
     -- Add measures of base component
     union all
     select <include refid="measureColumns"/> from project_measures pm
-    inner join snapshots analysis on analysis.uuid = pm.analysis_uuid
+    inner join snapshots analysis on analysis.uuid = pm.analysis_uuid and analysis.islast=${_true}
```
simonbrandhof: I suggest to move this condition to `selectTreeByQueryFilters` so that it drops a duplication and it defines all conditions in the same place.
That is not blocker.

PR #1430, opened 2016-12-02
SONAR-8464 Create WS api/project_analyses/add_event

teryk:


sonar-db/src/test/java/org/sonar/db/event/EventDaoTest.java
```diff
@@ -60,15 +64,16 @@ public void select_by_component_uuid() {
...
     dbTester.prepareDbUnit(getClass(), "shared.xml");
 
-    List<EventDto> dtos = dao.selectByComponentUuid(dbTester.getSession(), "ABCD");
+    List<EventDto> dtos = underTest.selectByComponentUuid(dbTester.getSession(), "ABCD");
     assertThat(dtos).extracting("category").containsOnly(EventDto.CATEGORY_ALERT, EventDto.CATEGORY_PROFILE, EventDto.CATEGORY_VERSION);
   }
 
   @Test
   public void insert() {
     dbTester.prepareDbUnit(getClass(), "empty.xml");
```
julienlancelot: Nice to have, it would have been good to remove DbUnit, as the goal is to stop using it (too much a pain to maintain).
Note that for migration it can still makes sense, as migration never change.

server/sonar-web/src/main/webapp/WEB-INF/db/migrate/1452_make_uuid_not_null_on_events.rb
```diff
@@ -0,0 +1,30 @@
...
+#
+
+#
+# SonarQube 6.3
+#
+class MakeUuidNotNullOnEvents < ActiveRecord::Migration
+
+  def self.up
+    execute_java_migration('org.sonar.db.version.v63.MakeUuidNotNullOnEvents')
+    add_index 'events', :uuid, :name => 'events_uuid', :unique => true
```
julienlancelot: I thought there were already a Java component to add/remove index, but it's not the case, too bad.

PR #1433, opened 2016-12-05
SONAR-8506 prepare refactoring of Ruby migrations

simonbrandhof:
by removing some useless migrations as minimal version to upgrade from is 5.6.


simonbrandhof:
Note that commit names reference the root ticket SONAR-5471 but not the sub-task SONAR-8506.

PR #1436, opened 2016-12-07
SONAR-8465 Create WS api/project_analyses/delete_event

teryk:


server/sonar-server/src/main/java/org/sonar/server/projectanalysis/ws/DeleteEventAction.java
```diff
@@ -0,0 +1,89 @@
...
+    response.noContent();
+  }
+
+  private void doHandle(DeleteEventRequest request) {
+    try (DbSession dbSession = dbClient.openSession(false)) {
+      EventDto event = dbClient.eventDao().selectByUuid(dbSession, request.getEvent())
+        .orElseThrow(() -> new NotFoundException(format("Event '%s' not found", request.getEvent())));
+      checkPermissions(event);
+
+      dbClient.eventDao().delete(dbSession, request.getEvent());
```
julienlancelot: If the event is a version, and this is the latest version of the project, we should fail

julienlancelot: You forgot to take into this https://github.com/SonarSource/sonarqube/blob/master/server/sonar-web/src/main/webapp/WEB-INF/app/controllers/api/events_controller.rb#L186

PR #1441, opened 2016-12-08
SONAR-8465 Create WS api/project_analyses/delete_event

teryk:


server/sonar-server/src/main/java/org/sonar/server/projectanalysis/ws/DeleteEventAction.java
```diff
@@ -0,0 +1,111 @@
...
+import org.sonar.db.DbSession;
+import org.sonar.db.component.SnapshotDto;
+import org.sonar.db.event.EventDto;
+import org.sonar.server.exceptions.NotFoundException;
+import org.sonar.server.user.UserSession;
+import org.sonarqube.ws.client.projectanalysis.DeleteEventRequest;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.lang.String.format;
+import static org.sonarqube.ws.client.projectanalysis.CreateEventRequest.Category.OTHER;
```
julienlancelot: You should move Category from CreateEventRequest class to a public class Event Category

julienlancelot: Or better, put the Category class inside EventDto

sonar-db/src/main/resources/org/sonar/db/event/EventMapper.xml
```diff
@@ -47,9 +56,13 @@
       #{createdAt, jdbcType=BIGINT})
   </insert>
 
-  <delete id="delete">
+  <delete id="deleteById" parameterType="Long">
```
julienlancelot: As you've introduced UUID, I would recommend you to remove this deleteById method and use deleteByUuid (remember, the goal is to not use anymore ID on this table, it will make the removal of the column easier)

server/sonar-server/src/main/java/org/sonar/server/projectanalysis/ws/DeleteEventAction.java
teryk: I cannot do in EventDto because of the separation between sonar-ws and sonar-db

PR #1454, opened 2016-12-12
SONAR-5471 RoR used only for wrapping Java DB migrations

simonbrandhof:


sonar-db/src/main/java/org/sonar/db/version/v55/CreateInitialSchema.java
```diff
@@ -0,0 +1,918 @@
...
+        .addColumn(newBigIntegerColumnDefBuilder().setColumnName("started_at").build())
+        .addColumn(newBigIntegerColumnDefBuilder().setColumnName("created_at").setIsNullable(false).build())
+        .addColumn(newBigIntegerColumnDefBuilder().setColumnName("updated_at").setIsNullable(false).build())
+        .build());
+    addIndex(context, "ce_queue", "ce_queue_component_uuid", false, componentUuidCol);
+    addIndex(context, "ce_queue", "ce_queue_uuid", true, uuidCol);
+  }
+
+  private void createFileSources(Context context) throws SQLException {
+    VarcharColumnDef projectUuidCol = newVarcharColumnDefBuilder().setColumnName("project_uuid").setLimit(50).setIsNullable(false).build();
```
sns-seb: you could use `VarcharColumnDef.UUID_VARCHAR_SIZE` annd `VarcharColumnDef.UUID_SIZE` to avoid the duplication, but I understand this is extra work with limited valuue

sonar-db/src/main/java/org/sonar/db/version/v55/PopulateInitialSchema.java
```diff
@@ -0,0 +1,118 @@
...
+
+    context.prepareUpsert("insert into groups_users (user_id, group_id) values ((select id from users where login='" + ADMIN_USER + "'), (select id from groups where name=?))")
+      .setString(1, ADMINS_GROUP).addBatch()
+      .setString(1, USERS_GROUP).addBatch()
+      .execute()
+      .commit();
+  }
+
+  private static void truncateTable(Context context, String table) throws SQLException {
+    context.prepareUpsert("truncate table " + table).execute().commit();
```
sns-seb: Isn't truncate super slow on some SGBD and `delete *` should be used  instead?

sonar-db/src/main/java/org/sonar/db/version/v55/CreateInitialSchema.java
```diff
@@ -0,0 +1,918 @@
...
+        .addColumn(newIntegerColumnDefBuilder().setColumnName("resource_id").setIsNullable(true).build())
+        .addColumn(newClobColumnDefBuilder().setColumnName("text_value").setIsNullable(true).build())
+        .addColumn(newIntegerColumnDefBuilder().setColumnName("user_id").setIsNullable(true).build())
+        .build());
+    addIndex(context, "properties", "properties_key", false, propKey);
+  }
+
+  private void createProjectLinks(Context context) throws SQLException {
+    context.execute(
+      new CreateTableBuilder(getDialect(), "project_links")
```
sns-seb: you could factor `context.execute(new CreateTableBuilder(getDialect(), "XXXXX")` into something like `createTable(newTableBuilder("XXXX").[... no call to build...])`, a bit as you did for index creation

sonar-db/src/test/java/org/sonar/db/version/CreateTableBuilderTest.java
```diff
@@ -526,7 +526,7 @@ public void build_adds_DEFAULT_clause_on_varchar_column_on_MySQL() {
 
   @Test
   public void build_adds_DEFAULT_clause_on_varchar_column_on_Oracle() {
-    verifyDefaultClauseOnVarcharColumn(ORACLE, "CREATE TABLE table_42 (status VARCHAR (1) DEFAULT 'P' NOT NULL)");
+    verifyDefaultClauseOnVarcharColumn(ORACLE, "CREATE TABLE table_42 (status VARCHAR (1 CHAR) DEFAULT 'P' NOT NULL)");
```
sns-seb: this change does not seem related to the commit

sonar-db/src/main/java/org/sonar/db/version/v55/CreateInitialSchema.java
```diff
@@ -915,4 +915,8 @@ private void addIndex(Context context, String table, String index, boolean uniqu
     }
     context.execute(builder.build());
   }
+
+  private static VarcharColumnDef.Builder newVarcharBuilder(String column) {
```
sns-seb: I think it should be clear that this method creates a "bad" varchar builder.
you could either deprecate this method or call it something like `newLenientVarcharBuilder`
at least it should be documented with some javadoc to avoid any confusion in the future

sonar-db/src/main/java/org/sonar/db/version/v55/CreateInitialSchema.java
simonbrandhof: I think migrations should be isolated as much as possible. That's why I didn't use the constant which value is not guaranteed to be the same over time. I know that it's pure theoretical as we won't do such change...

sonar-db/src/main/java/org/sonar/db/version/v55/CreateInitialSchema.java
simonbrandhof: ok to rename but note that there won't be any future confusion as regular VARCHAR will never be used in this migration.

sonar-db/src/main/java/org/sonar/db/version/v55/PopulateInitialSchema.java
sns-seb: since, I believe, the dialect is available in the context, I was more thinking of using the best statement for each SGBD

sonar-db/src/main/java/org/sonar/db/version/v55/CreateInitialSchema.java
sns-seb: it will when we write the `CreateInitialSchema` for 6.X LTS

PR #1463, opened 2016-12-14
Typo: "componet" instead of "component"

otrosien:


henryju:
Thanks for the contribution!

PR #1469, opened 2016-12-14
SONAR-8467 Fix total number of analyses when filtering by category

teryk:


server/sonar-server/src/main/java/org/sonar/server/projectanalysis/ws/SearchResults.java
```diff
@@ -111,14 +111,13 @@ String getProjectUuid() {
...
       ListMultimap<String, String> eventCategoriesByAnalysisUuid = events.stream()
         .collect(Collectors.index(EventDto::getAnalysisUuid, EventDto::getCategory));
-      AtomicInteger count = new AtomicInteger();
+      Predicate<SnapshotDto> byCategory = a -> eventCategoriesByAnalysisUuid.get(a.getUuid()).contains(request.getCategory().getLabel());
+      this.countAnalyses = (int) analyses.stream().filter(byCategory).count();
       this.analyses = analyses.stream()
-        .filter(a -> eventCategoriesByAnalysisUuid.get(a.getUuid()).contains(request.getCategory().getLabel()))
-        .peek(a -> count.incrementAndGet())
+        .filter(byCategory)
         .skip(Paging.offset(request.getPage(), request.getPageSize()))
```
teryk: note that skip and limit and short-circuiting operations so peek was applied only on the subset of stream even if put before. That's where the bug comes from

PR #1489, opened 2016-12-30
Add aria labels to issue rule and filter buttons in issue list.

jareds:
Add aria labels to issue rule and filter buttons in issue list. This is required since these buttons do not have text describing what they do so are announced as "button" when using Jaws screen reading software. Note I was planning on opening up an issue on JIRA but am unable to do so. The current JIRA instance for Sonar does not appear to offer an audio captcha option, so since I am a totally blind screen reader user I am unable to create a user account.

### DIFF
--- a/server/sonar-web/src/main/js/apps/issues/templates/issues-issue-filter.hbs
+++ b/server/sonar-web/src/main/js/apps/issues/templates/issues-issue-filter.hbs
@@ -1,5 +1,5 @@
 <li class="issue-meta">
-  <button class="button-link issue-action issue-action-with-options js-issue-filter">
+  <button class="button-link issue-action issue-action-with-options js-issue-filter" aria-label="filter similar issues">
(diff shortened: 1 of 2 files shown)


sns-seb:
Hello @jareds and thanks for the contribution.

This PR adds aria-labels in only two locations. Does it mean they are the only locations where the whole SQ user interface fails to provide correct accessibility?

jareds:
I am not sure what other accessibility issues may exist in the user interface. I have only been using Sonar for a couple of weeks at my company and so far all I’ve been doing is basic triage on issues so have not used much of the user interface. I have contacted SonarSource to ask for a Jira account so I can report other issues I find. I’m a backend developer who is just starting to learn frontend development with a focus on accessibility so if I find more issues that I do not know how to fix I will handle them through Jira.

stas-vilchik:
@jareds I added a proper l10n and merged to master (see 6b5beb1).

Thanks for the contribution!

PR #1492, opened 2017-01-02
Clean-up and upgrade Maven plugins

simonbrandhof:


pom.xml
```diff
@@ -168,21 +163,21 @@
...
-          <artifactId>maven-jxr-plugin</artifactId>
-          <version>2.5</version>
+          <artifactId>maven-jar-plugin</artifactId>
+          <version>3.0.2</version>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-plugin-plugin</artifactId>
-          <version>3.3</version>
+          <artifactId>maven-jxr-plugin</artifactId>
```
henryju: Do we still use this plugin?

simonbrandhof: yes, in javadoc

PR #1503, opened 2017-01-05
SONAR-8592 Reverse order of locations in flows of issues

dbmeneses:


sonar-plugin-api/src/main/java/org/sonar/api/batch/sensor/issue/NewIssue.java
```diff
@@ -69,7 +69,8 @@
 
   /**
    * Register a flow for this issue. A flow is an ordered list of issue locations that help to understand the issue.
-   * It could be a path leading to the primary location. Several flows can be registered.
+   * It should be a <b>path that backtracks the issue from the its primary location to the start of the flow</b>.
```
henryju: typo: "from the its"

PR #1513, opened 2017-01-10
SONAR-8518 "sonar.profile" makes the analysis fail in preview mode

dbmeneses:


it/it-tests/src/test/java/it/analysis/IssuesModeTest.java
```diff
@@ -206,6 +206,22 @@ public void only_scan_changed_files_on_change() throws IOException {
...
+  // SONAR-8518
+  @Test
+  public void shoud_support_sonar_profile_prop() throws IOException {
+    restoreProfile("one-issue-per-line.xml");
+    restoreProfile("empty.xml");
+    orchestrator.getServer().provisionProject("sample", "xoo-sample");
+    orchestrator.getServer().associateProjectToQualityProfile("sample", "xoo", "empty");
+    
+    SonarScanner runner = configureRunner("shared/xoo-sample", 
+      "sonar.verbose", "true",
```
henryju: Is it required. Logs are already soooo verbose :)

dbmeneses: If the test fails, it's quite important: it will display the WS calls with the quality profiles fetched and used

PR #1531, opened 2017-01-13
SONAR-8652 Pass value of sonar.projectOrganization in the scanner report

henryju:


sonar-plugin-api/src/main/java/org/sonar/api/CoreProperties.java
```diff
@@ -127,6 +127,11 @@
   String PROJECT_VERSION_PROPERTY = "sonar.projectVersion";
 
   /**
+   * @since 6.3
+   */
+  String PROJECT_ORGANIZATION_PROPERTY = "sonar.projectOrganization";
```
sns-seb: I've tested the commit, it works! \o/

but please change the name of the property to `sonar.organization` as this is how it's specified in the MMF.

If you'd like to challenge it, you should get back to @bellingard .

PR #1536, opened 2017-01-17
SONAR-8574 Replace usage of api/resources/index by api/measures/compo…

julienlancelot:
…nent in ITs


server/sonar-server/src/main/java/org/sonar/server/ws/RemovedWebServiceHandler.java
```diff
@@ -0,0 +1,41 @@
...
+import org.sonar.api.server.ws.RequestHandler;
+import org.sonar.api.server.ws.Response;
+import org.sonar.server.exceptions.ServerException;
+
+import static java.net.HttpURLConnection.HTTP_GONE;
+
+/**
+ * Used to declare web services that are removed
+ */
+public class RemovedWebServiceHandler implements RequestHandler {
```
teryk: use an Enum

julienlancelot:
Cleanup of JRubyFacade as this class will be removed very soon

server/sonar-server/src/main/java/org/sonar/server/ws/RemovedWebServiceHandler.java
julienlancelot: ok

PR #1542, opened 2017-01-18
SONAR-7282 fix project deletion

bartfastiel:


server/sonar-server/src/main/java/org/sonar/server/component/index/ComponentIndexer.java
```diff
@@ -91,11 +95,18 @@ public void deleteByProjectUuid(String uuid) {
...
   private void deleteComponentsByProjectUuid(String projectUuid) {
-    esClient
-      .prepareDelete(INDEX_COMPONENTS, TYPE_COMPONENT, projectUuid)
-      .setRouting(projectUuid)
-      .setRefresh(true)
-      .get();
+    BulkIndexer.delete(esClient, INDEX_COMPONENTS, projectUuidSearch(projectUuid));
+  }
+
+  private SearchRequestBuilder projectUuidSearch(String projectUuid) {
```
simonbrandhof: Having three private methods breaks readability IMO. `projectUuidFilter()` and `projectUuidSearch()` are used only once and so do not aim to share code. The following seems more simple for me because I can read the full Elasticsearch request : 
```
private void deleteComponentsByProjectUuid(String projectUuid) {
    BulkIndexer.delete(esClient, INDEX_COMPONENTS, esClient.prepareSearch(INDEX_COMPONENTS)
      .setQuery(boolQuery()
        .filter(
          termQuery(ComponentIndexDefinition.FIELD_PROJECT_UUID, projectUuid))));
  }
```

server/sonar-server/src/test/java/org/sonar/server/component/index/ComponentIndexerTest.java
```diff
@@ -121,16 +121,36 @@ public void reindex_project() {
     assertMatches("NewName", 1);
   }
 
+  @Test
+  public void reindex_project_with_files() {
```
simonbrandhof: The test name should mention that a file is deleted/updated.

PR #1547, opened 2017-01-19
SONAR-7305 Create WS api/measures/search_history

teryk:


server/sonar-server/src/test/java/org/sonar/server/measure/ws/SearchHistoryActionTest.java
```diff
@@ -0,0 +1,386 @@
...
+    wsRequest.setMetrics(newArrayList(complexityMetric.getKey(), nclocMetric.getKey(), "METRIC_42", "42_METRIC"));
+
+    expectedException.expect(IllegalArgumentException.class);
+    expectedException.expectMessage("Metrics 42_METRIC, METRIC_42 are not found");
+
+    call();
+  }
+
+  @Test
+  public void fail_if_not_enough_permissions() {
```
julienlancelot: I don't understand the content of this test, from where is coming the ForbiddenException ?
Anonymous user should have access to the project if the "Browse" permission is set to the project.
It's not clear for me what is really tested here, reading the test I'm thinking that this WS requires to be authenticated which is not the case.

server/sonar-server/src/main/java/org/sonar/server/measure/ws/SearchHistoryAction.java
```diff
@@ -0,0 +1,179 @@
...
+      component.uuid(),
+      analyses.stream().map(SnapshotDto::getUuid).collect(Collectors.toList()),
+      metrics.stream().map(MetricDto::getId).collect(Collectors.toList()));
+  }
+
+  private List<SnapshotDto> searchAnalyses(DbSession dbSession, SearchHistoryRequest request, ComponentDto component) {
+    SnapshotQuery dbQuery = new SnapshotQuery()
+      .setComponentUuid(component.projectUuid())
+      .setStatus(STATUS_PROCESSED)
+      .setSort(SORT_FIELD.BY_DATE, SORT_ORDER.ASC);
```
teryk: as discussed it doesn't seem obvious to me, we should discuss with Atlas

server/sonar-server/src/main/java/org/sonar/server/measure/ws/TimeMachineWs.java
```diff
@@ -1,71 +0,0 @@
...
- */
-package org.sonar.server.measure.ws;
-
-import com.google.common.io.Resources;
-import org.sonar.api.server.ws.RailsHandler;
-import org.sonar.api.server.ws.WebService;
-
-import static org.sonar.server.ws.KeyExamples.KEY_PROJECT_EXAMPLE_001;
-
-public class TimeMachineWs implements WebService {
```
teryk: is it what you did in your PR about Ruby?

server/sonar-server/src/test/java/org/sonar/server/measure/ws/SearchHistoryActionTest.java
teryk: `userSession.anonymous()` cancels the current userSession rights so it doesn't have any specific permission

PR #1550, opened 2017-01-23
SONAR-7282 when updating a project, clear ALL components from index

bartfastiel:


server/sonar-server/src/main/java/org/sonar/server/component/index/ComponentIndexer.java
```diff
@@ -58,44 +62,43 @@ public ComponentIndexer(DbClient dbClient, EsClient esClient) {
...
    */
   public void indexByProjectUuid(String projectUuid) {
+    requireNonNull(projectUuid);
+    deleteComponentsByProjectUuid(projectUuid);
+    doIndexByProjectUuid(projectUuid);
+  }
+
+  private void doIndexByProjectUuid(@Nullable String projectUuid) {
+    BulkIndexer bulk = new BulkIndexer(esClient, INDEX_COMPONENTS);
+    bulk.setLarge(true);
```
simonbrandhof: The attribute `setLarge` must be enabled only during server startup because it disables replicas

PR #1552, opened 2017-01-23
Add `hasFilename` predicate with caching

janos-ss:
To be merged later, not now, to `master` instead of this branch.

Let me know if this is going in the right direction. In particular, I'm not too thrilled about adding methods in `FileSystem.Index` and in `ModuleInputComponentStore` but I don't see a better solution. Let me know if I missed something.

I plan to follow the same pattern for querying by extension, so one more method in these interfaces/classes, and one more multimap.


janos-ss:
Thanks, this will be part of another pull request, probably into `master`.

PR #1556, opened 2017-01-24
SONAR-8678 allow search for components by multiple, independent words

bartfastiel:


server/sonar-server/src/test/java/org/sonar/server/component/index/ComponentIndexTest.java
```diff
@@ -259,6 +259,22 @@ public void should_find_item_despite_two_missing_characters_and_lowercase_and_in
...
+
+    assertSearchResults("plugin java", file);
+  }
+
+  @Test
+  public void should_search_for_two_words_in_any_order() {
+    ComponentDto project = indexProject("key-1", "SonarQube");
+    ComponentDto file = indexFile(project, "src/main/java/AbstractPluginFactory.java", "AbstractPluginFactory.java");
+
+    assertSearchResults("java plugin", file);
```
simonbrandhof: You can complete the two tests with tokens that do not represent the file extension: `abstract factory` and `factory abstract`.

simonbrandhof: You can also add a test with one missing word: `factory something` should return the file. `something else` should return nothing.

PR #1562, opened 2017-01-24
SONAR-8397 Exploring a failed quality condition on rating is not usable

stas-vilchik:


julienlancelot:
Sounds good and more usable !

PR #1580, opened 2017-01-27
6.3 hardening 2

julienlancelot:


it/it-plugins/server-plugin/src/main/resources/widgets/widget-displaying-properties.html.erb
```diff
@@ -1,12 +0,0 @@
-<table>
```
teryk: 👍

server/sonar-server/src/test/java/org/sonar/server/authentication/InitFilterTest.java
```diff
@@ -173,6 +173,18 @@ public void redirect_when_failing_because_of_UnauthorizedExceptionException() th
     assertThat(authenticationException.getPublicMessage()).isEqualTo("Email john@email.com is already used");
   }
 
+  @Test
+  public void redirect_with_context_path_when_failing_because_of_UnauthorizedExceptionException() throws Exception {
```
teryk: ExceptionExceptionExceptionException? ^^

julienlancelot: I'll fix this

PR #1599, opened 2017-01-31
SONAR-8694 boost prefix and case sensitive prefix matches in search

bartfastiel:


server/sonar-server/src/main/java/org/sonar/server/es/DefaultIndexSettings.java
```diff
@@ -28,6 +28,9 @@
   /** Maximum length of ngrams. */
   public static final int MAXIMUM_NGRAM_LENGTH = 15;
 
+  /** Pattern, that splits the user search input **/
+  public static final String SEARCH_TERM_TOKENIZER_PATTERN = "[\\s]+";
```
simonbrandhof: This constant is specific to the context of `ComponentIndexSearchFeature` and should move accordingly.

server/sonar-server/src/test/java/org/sonar/server/component/index/ComponentIndexFeaturePrefixTest.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Test
+  public void should_be_able_to_match_case() {
+    features.set(ComponentIndexSearchFeature.PREFIX);
+    assertNoFileMatches("cOmPoNeNt.Js", "CoMpOnEnT.jS");
+  }
+
+  @Test
+  public void should_prefer_matching_case() {
```
simonbrandhof: nice !

PR #1601, opened 2017-01-31
SONAR-8523 refactor user index

bartfastiel:


server/sonar-server/src/main/java/org/sonar/server/user/index/UserIndexDefinition.java
```diff
@@ -56,39 +55,14 @@ public void define(IndexDefinitionContext context) {
...
-      .put("index.analysis.analyzer.search_ngrams.type", "custom")
-      .put("index.analysis.analyzer.search_ngrams.tokenizer", "whitespace")
-      .putArray("index.analysis.analyzer.search_ngrams.filter", "trim", "lowercase");
-
     // type "user"
     NewIndex.NewIndexType mapping = index.createType(TYPE_USER);
-    mapping.stringFieldBuilder(FIELD_LOGIN).addSubField(SEARCH_SUB_SUFFIX, buildGramSearchField()).build();
-    mapping.stringFieldBuilder(FIELD_NAME).addSubField(SEARCH_SUB_SUFFIX, buildGramSearchField()).build();
-    mapping.stringFieldBuilder(FIELD_EMAIL).addSubField(SEARCH_SUB_SUFFIX, buildGramSearchField()).enableSorting().build();
+    mapping.stringFieldBuilder(FIELD_LOGIN).enable(USER_SEARCH_GRAMS_ANALYZER).build();
```
simonbrandhof: The fact that a sub-field is created does not appear anymore with this method, whereas it's important when using it (see `USER_SEARCH_GRAMS_ANALYZER.subField(FIELD_LOGIN)` in `UserIndex`).

bartfastiel: I agree and I renamed "enable" to "addSubFields".

simonbrandhof: that does not resolve my concern. Names of sub-fields become magic.

PR #1605, opened 2017-01-31
SONAR-8651 add organization to api/projects/create

sns-seb:


server/sonar-server/src/test/java/org/sonar/server/tester/UserSessionRule.java
```diff
@@ -102,22 +102,22 @@ public static UserSessionRule forServerTester(ServerTester serverTester) {
   /**
    * Log in with the default login {@link #DEFAULT_LOGIN}
    */
-  public UserSessionRule login() {
-    return login(DEFAULT_LOGIN);
+  public UserSessionRule log_in() {
```
simonbrandhof: Introducing an underscore in a Java method does not improve clarity. Please rename to `logIn()`.

server/sonar-server/src/test/java/org/sonar/server/project/ws/CreateActionTest.java
```diff
@@ -71,24 +73,26 @@
...
-          componentUpdater,
-          defaultOrganizationProvider));
+    new CreateAction(
+      new ProjectsWsSupport(db.getDbClient()),
+      db.getDbClient(), userSession,
+      componentUpdater,
+      defaultOrganizationProvider));
 
   @Test
   public void create_project() throws Exception {
```
simonbrandhof: You can also test that project cannot be created in an organization if user has provisioning permission in another organization.

PR #1615, opened 2017-02-01
SONAR-8664 add organization to response of api/notifications/list

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/notification/ws/ListAction.java
```diff
@@ -146,35 +150,53 @@ private boolean isDispatcherAuthorized(PropertyDto prop, String dispatcher) {
...
-    return dbClient.componentDao().selectByIds(dbSession,
-      properties.stream()
-        .filter(prop -> prop.getResourceId() != null)
-        .map(PropertyDto::getResourceId)
-        .distinct()
-        .collect(toList()))
+    Set<String> authorizedComponentUuids = new HashSet<>(dbClient.authorizationDao().selectAuthorizedRootProjectsUuids(dbSession, userSession.getUserId(), UserRole.USER));
+    Set<Long> componentIds = properties.stream()
+      .filter(prop -> prop.getResourceId() != null)
+      .map(PropertyDto::getResourceId)
```
simonbrandhof: could also be written `.map(PropertyDto::getResourceId).filter(Objects::nonNull)` to remove the condition `!=null` :-)

sonar-ws/src/main/protobuf/ws-notifications.proto
```diff
@@ -35,6 +35,7 @@ message ListResponse {
 message Notification {
   optional string channel = 1;
   optional string type = 2;
+  optional string organization = 5;
```
simonbrandhof: having 5 before 3 and 4 is error-prone for future additions.

sns-seb: but the error doesn't last past the compilation, so...

I think it's a tradeoff between backward compatibility and field sorting that functionally makes sense

server/sonar-server/src/main/java/org/sonar/server/notification/ws/ListAction.java
sns-seb: legacy code, but yeah, I can improve it

sonar-ws/src/main/protobuf/ws-notifications.proto
simonbrandhof: I was just talking about moving the line, not changing the field number !

sns-seb: that's my point, functionally it makes more sense to have the organization before the component (key), ie. sort from the most generic to the most specific. Yet, organization can be number '3' for backward compatibility reasons.

PR #1616, opened 2017-02-01
Search on Users case insensitive

teryk:
- only search on name is case insensitive
- I still have to do a local performance test to check it scales for 10'000 users

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/permission/ws/UsersAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/permission/ws/UsersAction.java
@@ -121,9 +121,7 @@ public class UsersAction implements PermissionsWsAction {
       .setPageIndex(request.mandatoryParamAsInt(Param.PAGE))
       .setPageSize(request.mandatoryParamAsInt(Param.PAGE_SIZE))
       .setSearchQuery(textQuery);
-    if (project.isPresent()) {
-      permissionQuery.setComponentUuid(project.get().getUuid());
-    }
+    project.ifPresent(projectId -> permissionQuery.setComponentUuid(projectId.getUuid()));
     if (permission != null) {
       if (project.isPresent()) {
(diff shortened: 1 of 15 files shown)


server/sonar-server/src/main/java/org/sonar/server/permission/ws/UsersAction.java
```diff
@@ -121,9 +121,7 @@ private static PermissionQuery buildPermissionQuery(Request request, Optional<Pr
       .setPageIndex(request.mandatoryParamAsInt(Param.PAGE))
       .setPageSize(request.mandatoryParamAsInt(Param.PAGE_SIZE))
       .setSearchQuery(textQuery);
-    if (project.isPresent()) {
-      permissionQuery.setComponentUuid(project.get().getUuid());
-    }
+    project.ifPresent(projectId -> permissionQuery.setComponentUuid(projectId.getUuid()));
```
julienlancelot: 👍

server/sonar-server/src/test/java/org/sonar/server/usergroups/ws/UsersActionTest.java
```diff
@@ -250,9 +248,18 @@ public void filtering() throws Exception {
       .execute()
       .assertJson(getClass(), "all.json");
 
-    newUsersRequest()
-      .setParam("id", group.getId().toString())
-      .setParam("q", "love")
+    newUsersRequest().setParam("id", group.getId().toString())
+      .setParam("q", ".logi")
+      .execute()
+      .assertJson(getClass(), "all_ada.json");
```
julienlancelot: You could have done some refactoring to remove JSON usage :)

teryk: 😥 I know…

PR #1627, opened 2017-02-03
SONAR-8736 add missing createdAt in example response ws/rule/search

bartfastiel:


server/sonar-server/src/main/resources/org/sonar/server/rule/ws/example-search.json
```diff
@@ -7,6 +7,7 @@
       "key": "squid:S1067",
       "repo": "squid",
       "name": "Expressions should not be too complex",
+      "createdAt": "2013-03-27T08:52:40+0100",
```
julienlancelot: "createdAt" is not a field that is always returned, it's only returned when the parameter 'f=createdAt' is set.
The example should not contain the exhaustive list of fields, so in my opinion it's not mandatory to put it, but why not !

bartfastiel: There have been "createdAt" fields on _some_ rules. I think it is better to have it on all or none rule.

PR #1630, opened 2017-02-05
SONAR-8438 New web service api/system/ping

simonbrandhof:
SONAR-8438 New web service api/system/ping


server/sonar-server/src/main/java/org/sonar/server/platform/ws/PingAction.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static org.apache.commons.io.IOUtils.write;
+
+public class PingAction implements SystemWsAction {
+  @Override
+  public void define(WebService.NewController controller) {
+    controller.createAction("ping")
+      .setDescription("Answers \"pong\" as plain-text")
+      .setSince("6.3")
+      .setResponseExample(getClass().getResource("example-ping.txt"))
```
julienlancelot: In order for the example to be more visible, we're trying to start the file by the name of the action -> ping-example.txt
(I know that it's not done everywhere yet...)

simonbrandhof: ok I didn't know. I will fix but still I suggest you to do a mass change.

simonbrandhof: fixed in next fixup commit

julienlancelot: Perfect ! You're right we should do some mass update

PR #1640, opened 2017-02-08
SONAR-8754 add support for guarded organizations

sns-seb:


server/sonar-server/src/test/java/org/sonar/server/organization/ws/DeleteActionTest.java
```diff
@@ -126,7 +126,7 @@ public void request_fails_with_NotFoundException_if_organization_with_specified_
   }
 
   @Test
-  public void request_fails_with_ForbiddenException_when_user_has_no_System_Administer_permission() {
+  public void request_fails_with_ForbiddenException_when_user_has_no_System_Administer_permission_for_non_guarded_organization() {
```
simonbrandhof: to prepare renaming of constant `GlobalPermissions.SYSTEM_ADMIN` to something like `ADMINISTER_ORGANIZATION`, I suggest to not use the term "System Administer", for instance `request_fails_with_ForbiddenException_when_user_is_not_administrator_of_non_guarded_organization`

simonbrandhof: (this feedback applies to other tests of this file too)

PR #1650, opened 2017-02-09
SONAR-8745 rule activation in scanner report

bartfastiel:
Please notice, that there are three commits.

1. A minor test refactoring.
2. A bigger refactoring of a scanner related test. It used to use binary files as input - I changed it to use a generated input.
3. The main change: the scanner should include the rule activation dates in the scanner report.


sonar-ws/src/main/protobuf/ws-rules.proto
```diff
@@ -154,6 +154,7 @@ message Active {
   optional string qProfile = 1;
   optional string inherit = 2;
   optional string severity = 3;
+  optional string createdAt = 6;
```
henryju: Would be more readable to sort fields by index

sonar-plugin-api/src/main/java/org/sonar/api/batch/rule/ActiveRule.java
```diff
@@ -56,6 +54,8 @@
    */
   Map<String, String> params();
 
+  long createdAt();
```
henryju: I would not add that to the public API. This is technical stuff.

sonar-scanner-engine/src/main/java/org/sonar/scanner/report/ActiveRulesPublisher.java
```diff
@@ -43,6 +43,7 @@ public void publish(ScannerReportWriter writer) {
         builder.setRuleRepository(input.ruleKey().repository());
         builder.setRuleKey(input.ruleKey().rule());
         builder.setSeverity(Constants.Severity.valueOf(input.severity()));
+        builder.setCreatedAt(input.createdAt());
```
henryju: You can cast to DefaultActiveRule if needed since this is internal use.

bartfastiel: I used `.map(input -> (DefaultActiveRule) input)` for that. Ok?

server/sonar-server/src/test/java/org/sonar/server/computation/task/projectanalysis/qualityprofile/ActiveRulesHolderImplTest.java
```diff
@@ -48,7 +48,7 @@ public void get_inactive_rule() throws Exception {
 
   @Test
   public void get_active_rule() throws Exception {
-    underTest.set(asList(new ActiveRule(RULE_KEY, Severity.BLOCKER, Collections.<String, String>emptyMap())));
+    underTest.set(asList(new ActiveRule(RULE_KEY, Severity.BLOCKER, Collections.<String, String>emptyMap(), 1_000L)));
```
sns-seb: we've come to use constants to avoid non descriptive magic numbers (and duplication, but that's not the main point).
Here, you could name it `SOME_DATE`

sonar-scanner-engine/src/main/java/org/sonar/scanner/report/ActiveRulesPublisher.java
sns-seb: or you could use a method reference to `DefaultActiveRule.class::cast`

(much cleaner, imho)

PR #1657, opened 2017-02-10
SONAR-8736 keep leak period clean, if rule is activated

bartfastiel:


server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/issue/IssueCreationDateCalculator.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @Override
+  public void onIssue(Component component, DefaultIssue issue) {
+    if (issue.isNew() && ruleIsNew(issue)) {
+
+      // get scm information of file, if available
+      getChangeSet(component)
+        .map(scmInfo -> Optional
+
+          // if possible determine line change
```
sns-seb: when you need to document a basic if then else, don't you think the optional API becomes pointless?

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/issue/IssueCreationDateCalculator.java
```diff
@@ -0,0 +1,123 @@
...
+    this.analysisMetadataHolder = analysisMetadataHolder;
+    this.changeContext = createScan(new Date(analysisMetadataHolder.getAnalysisDate()));
+    this.activeRulesHolder = activeRulesHolder;
+  }
+
+  @Override
+  public void onIssue(Component component, DefaultIssue issue) {
+    if (issue.isNew() && ruleIsNew(issue)) {
+
+      // get scm information of file, if available
```
sns-seb: hint of this being abusive: 
when there is no changeset, date of issue is not altered... this is hidden in the last line `ifPresent` while for understanding by reader, this should be written as a fail fast condition

functional programming is about making code closer to the algorithm... this is quite the opposite

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/issue/IssueCreationDateCalculator.java
```diff
@@ -0,0 +1,123 @@
...
+    // special case: this is the first analysis of the project: use scm dates for all issues
+    return true;
+  }
+
+  private Optional<Long> lastAnalysis() {
+    return Optional.ofNullable(analysisMetadataHolder.getBaseAnalysis()).map(Analysis::getCreatedAt);
+  }
+
+  private long ruleCreation(DefaultIssue issue) {
+    ActiveRule activeRule = toJavaUtilOptional(activeRulesHolder.get(issue.getRuleKey()))
```
sns-seb: we need to discuss whether this is more readable that the following (ping @simonbrandhof )
```
Optional<ActiveRule> activeRule = activeRulesHolder.get(issue.getRuleKey());
checkArgument(activeRule.isPresent(), "The rule %s raised an issue, but is not one of the active rules.", issue.getRuleKey().rule());
return activeRule.get();
```

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/container/ProjectAnalysisTaskContainerPopulator.java
```diff
@@ -136,7 +137,7 @@ public void populateContainer(TaskContainer container) {
    * List of all objects to be injected in the picocontainer dedicated to computation stack.
    * Does not contain the steps declared in {@link ReportComputationSteps#orderedStepClasses()}.
    */
-  private static List componentClasses() {
+  private static List<Object> componentClasses() {
```
sns-seb: commit title "SONAR-8736 backdate issues that are caused by quality profile changes" is ambiguous.
"quality profile change" could both apply to swtiching from one QP to another to adding rule to a profile

could be "SONAR-8736 backdate issues caused by just enabled rules"

PR #1658, opened 2017-02-10
Functional version of SonarQube and plugins displayed

teryk:
release of pom parent, maven packager and update center need to be done

### DIFF
--- a/it/it-plugins/access-secured-props-plugin/pom.xml
+++ b/it/it-plugins/access-secured-props-plugin/pom.xml
@@ -10,7 +10,7 @@
   <artifactId>access-secured-props-plugin</artifactId>
   <packaging>sonar-plugin</packaging>
   <name>Plugins :: Access Secured Properties</name>
-  <version>1.0-SNAPSHOT</version>
+  <version>1.0.2.42</version>
   <description>Access .secured properties</description>
 
   <dependencies>
@@ -27,7 +27,7 @@
       <plugin>
         <groupId>org.sonarsource.sonar-packaging-maven-plugin</groupId>
         <artifactId>sonar-packaging-maven-plugin</artifactId>
-        <version>1.15</version>
+        <version>1.18.0.372</version>
         <extensions>true</extensions>
         <configuration>
           <pluginClass>AccessSecuredPropsPlugin</pluginClass>
--- a/it/it-tests/src/test/java/it/plugins/VersionPluginTest.java
+++ b/it/it-tests/src/test/java/it/plugins/VersionPluginTest.java
@@ -1,7 +1,7 @@
 /*
  * SonarQube
- * Copyright (C) 2009-2016 SonarSource SA
- * mailto:contact AT sonarsource DOT com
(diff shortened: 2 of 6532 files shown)


server/sonar-server/src/main/java/org/sonar/server/ui/ws/GlobalAction.java
```diff
@@ -122,7 +125,27 @@ private void writeQualifiers(JsonWriter json) {
   }
 
   private void writeVersion(JsonWriter json) {
-    json.prop("version", server.getVersion());
+    String technicalVersion = server.getVersion();
+    String functionalVersion = technicalVersion;
```
julienlancelot: I'm not sure to understand why are you using 2 variables. Only one should be enough (just replace technicalVersion by server.getVersion() in next line) => it will make the code more clear IMO.

julienlancelot:
The WS api/server/upgrades has been forgotten => it will always display the technical version

it/it-tests/src/test/java/it/plugins/VersionPluginTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.sonarqube.ws.client.GetRequest;
+import org.sonarqube.ws.client.WsClient;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static util.ItUtils.newAdminWsClient;
+
+public class VersionPluginTest {
+
+  @ClassRule
+  public static Orchestrator orchestrator = Category3Suite.ORCHESTRATOR;
```
julienlancelot: You have forgotten to add this test in the suite classes of Category3Suite

it/it-plugins/access-secured-props-plugin/pom.xml
```diff
@@ -10,7 +10,7 @@
   <artifactId>access-secured-props-plugin</artifactId>
   <packaging>sonar-plugin</packaging>
   <name>Plugins :: Access Secured Properties</name>
-  <version>1.0-SNAPSHOT</version>
+  <version>1.0.2.42</version>
```
teryk: as discussed, kept as is

server/sonar-server/src/main/java/org/sonar/server/ui/ws/GlobalAction.java
teryk: I've updated, I'm not sure the code is clearer because I split the functional version

server/sonar-server/src/main/java/org/sonar/server/ui/VersionFormatter.java
```diff
@@ -0,0 +1,54 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.server.ui;
+
+import com.google.common.base.Splitter;
+import java.util.List;
+
+public class VersionFormatter {
```
julienlancelot: Please add UTs

PR #1662, opened 2017-02-10
use webpack 2

stas-vilchik:


stas-vilchik:
@gregaubert I've rebased that PR on master, so it has a fresh state. You can have a look, but I'm not going to merge it for the moment. There is an issue with webpack tree shaking (webpack/webpack#2867), which removes the benefit of using webpack 2.

PR #1664, opened 2017-02-12
SONAR-8772 Performance hotspot in verification of project permissions

simonbrandhof:


sonar-db/src/main/java/org/sonar/db/permission/AuthorizationDao.java
```diff
@@ -117,6 +133,10 @@ public int countUsersWithGlobalPermissionExcludingUserPermission(DbSession dbSes
       });
   }
 
+  /**
+   * @deprecated it loads too many results and there's no functional need.
+   */
+  @Deprecated
   public Collection<String> selectAuthorizedRootProjectsUuids(DbSession dbSession, @Nullable Integer userId, String role) {
```
SonarTech: INFO Do not forget to remove this deprecated code someday. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS1133)

SonarTech:
* ![INFO][INFO] 1 info

[INFO]: https://raw.githubusercontent.com/SonarCommunity/sonar-github/master/images/severity-info.png 'Severity: INFO'

server/sonar-server/src/main/java/org/sonar/server/user/ServerUserSession.java
```diff
@@ -152,21 +149,29 @@ protected boolean hasOrganizationPermissionImpl(String organizationUuid, String
...
+    if (permissionsByProjectUuid == null) {
+      permissionsByProjectUuid = HashMultimap.create();
     }
-    return projectUuidByPermission.get(permission).contains(projectUuid);
+    Set<String> permissions;
+    if (permissionsByProjectUuid.containsKey(projectUuid)) {
+      permissions = permissionsByProjectUuid.get(projectUuid);
+    } else {
+      permissions = loadProjectPermissions(projectUuid);
+      permissionsByProjectUuid.putAll(projectUuid, permissions);
```
sns-seb: according to javadoc of `putAll`: 
> In particular, this is a no-op if {@code values} is empty.

I'm wondering there is actually any caching when there is no permission for the projectUuid.
If it's a no-op, then the key won't be added to the Multimap and `containsKey` will always return `false`, hence no caching the next time method is called with the same `projectUuid`.

In which case, a `Map<String, Set<String>>` should be used instead (and the call to `containsKey` dropped in favor of a single call to `get`)

server/sonar-server/src/main/java/org/sonar/server/user/ServerUserSession.java
```diff
@@ -152,21 +149,29 @@ protected boolean hasOrganizationPermissionImpl(String organizationUuid, String
...
-        Collection<String> projectUuids = dbClient.authorizationDao().selectAuthorizedRootProjectsUuids(dbSession, getUserId(), permission);
-        addProjectPermission(permission, projectUuids);
-      }
+    if (permissionsByProjectUuid == null) {
+      permissionsByProjectUuid = HashMultimap.create();
     }
-    return projectUuidByPermission.get(permission).contains(projectUuid);
+    Set<String> permissions;
+    if (permissionsByProjectUuid.containsKey(projectUuid)) {
+      permissions = permissionsByProjectUuid.get(projectUuid);
```
sns-seb: why not `return permissionsByProjectUuid.get(projectUuid).contains(permission)` and drop the `else` and the undefined variable?

server/sonar-server/src/main/java/org/sonar/server/user/ServerUserSession.java
```diff
@@ -152,21 +149,29 @@ protected boolean hasOrganizationPermissionImpl(String organizationUuid, String
...
   }
 
-  private void addProjectPermission(String permission, Collection<String> authorizedProjectUuids) {
-    for (String key : authorizedProjectUuids) {
-      projectUuidByPermission.put(permission, key);
+  private Set<String> loadProjectPermissions(String projectUuid) {
+    Set<String> permissions;
+    try (DbSession dbSession = dbClient.openSession(false)) {
+      if (userDto != null && userDto.getId() != null) {
+        permissions = dbClient.authorizationDao().selectProjectPermissions(dbSession, projectUuid, userDto.getId());
```
sns-seb: I favor multiple returns over undefined variable at the beginning of method. So old-school procedural style... :p

server/sonar-server/src/test/java/org/sonar/server/user/ServerUserSessionTest.java
```diff
@@ -80,6 +80,15 @@ public void setUp() throws Exception {
   }
 
   @Test
+  public void anonymous_is_not_logged_in_and_does_not_have_login() throws Exception {
```
sns-seb: I don't see any test of the caching mechanism, except for `hasSystemAdminister` and `getGroups`. From statically looking at the code, this is a miss (see my other comment).

server/sonar-server/src/main/java/org/sonar/server/user/ServerUserSession.java
simonbrandhof: I'm surprised myself.

server/sonar-server/src/main/java/org/sonar/server/user/ServerUserSession.java
```diff
@@ -52,12 +49,11 @@
   private final DbClient dbClient;
   private final OrganizationFlags organizationFlags;
   private final DefaultOrganizationProvider defaultOrganizationProvider;
-  private final Supplier<List<GroupDto>> groups = Suppliers.memoize(this::loadGroups);
+  private final Supplier<Collection<GroupDto>> groups = Suppliers.memoize(this::loadGroups);
```
sns-seb: I think this should be a Set and equals/hashcode should be defined in GroupDto (based on the id)

It's important we know the groups are not ordered, but also as important that there is no duplicate.

WDYT?

simonbrandhof: yes, good idea. Out of scope of this PR so I will address it just after.

PR #1674, opened 2017-02-15
SONAR-7282 only reindex components if the index is empty

bartfastiel:


simonbrandhof:
As discussed the test must be refactored in order minimize number of mocks.

PR #1682, opened 2017-02-16
Create api/projects/search

julienlancelot:


server/sonar-server/src/main/java/org/sonar/server/project/ws/SearchAction.java
```diff
@@ -0,0 +1,153 @@
...
+    this.support = support;
+  }
+
+  @Override
+  public void define(WebService.NewController context) {
+    WebService.NewAction action = context.createAction(ACTION_SEARCH)
+      .setSince("6.3")
+      .setDescription("Search for projects or views.<br>" +
+        "Requires 'System Administrator' permission")
+      .setInternal(true)
```
teryk: why internal?

server/sonar-server/src/main/java/org/sonar/server/project/ws/SearchAction.java
```diff
@@ -0,0 +1,153 @@
...
+      .addPagingParams(100)
```
teryk: no check on pagination

server/sonar-server/src/main/java/org/sonar/server/project/ws/SearchAction.java
julienlancelot: Why not ?

server/sonar-server/src/main/java/org/sonar/server/project/ws/SearchAction.java
julienlancelot: where ?

PR #1683, opened 2017-02-16
SONAR-8769 Display project and organization keys in the analysis logs

dbmeneses:


sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/ProjectScanContainer.java
```diff
@@ -238,6 +238,10 @@ protected void doAfterStart() {
     if (StringUtils.isNotEmpty(organization)) {
       LOG.info("Organization key: {}", organization);
     }
+    String branch = tree.root().definition().getBranch();
+    if (branch != null) {
```
henryju: What if branch is empty? Is it possible?

dbmeneses: getBranch() handles that case and returns null. Put I will add a medium test for it.

PR #1689, opened 2017-02-20
set embedded database timeout to 30s for slow build machines

bartfastiel:


server/sonar-server/src/test/java/org/sonar/server/platform/db/EmbeddedDatabaseTest.java
```diff
@@ -50,7 +50,7 @@
   @Rule
   public TemporaryFolder temporaryFolder = new TemporaryFolder();
   @Rule
-  public Timeout timeout = Timeout.seconds(10);
+  public Timeout timeout = Timeout.seconds(30);
```
sns-seb: Travis is our reference of "slow" and 10s is ok there
the problem is most likely not on the UT's side

bartfastiel: I still would love to change this. It won't change the result of unit tests on any of our machines - but might help users with slow machines to build.

At no cost we can improve the tests.

PR #1712, opened 2017-02-27
SONAR-8092 use BulkProcessor and calculate concurrent requests with actual number of shards

bartfastiel:


bartfastiel:
Replaced by #1716

PR #1723, opened 2017-03-01
Add tags to WS of components domain

teryk:


server/sonar-server/src/test/java/org/sonar/server/component/ws/ShowActionTest.java
```diff
@@ -78,6 +79,7 @@ public void json_example() throws IOException {
       .getInput();
 
     assertJson(response).isSimilarTo(getClass().getResource("show-example.json"));
+    assertThat(response).containsOnlyOnce("\"tags\"");
```
julienlancelot: I would prefer having a dedicated unit test on tags

PR #1734, opened 2017-03-02
Fix/branch preview mode

henryju:


sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/DefaultComponentTree.java
```diff
@@ -41,13 +38,8 @@ public void index(InputComponent component, InputComponent parent) {
...
     parents.put(component, parent);
-    Set<InputComponent> list = children.get(parent);
-    if (list == null) {
-      list = new LinkedHashSet<>();
-      children.put(parent, list);
-    }
-
-    list.add(component);
+    children.putIfAbsent(parent, new LinkedHashSet<>());
+    children.get(parent).add(component);
```
janos-ss: Could be written as `children.computeIfAbsent(parent, k -> new LinkedHashSet<Integer>()).add(component);`

sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/report/JSONReport.java
```diff
@@ -130,10 +135,15 @@ private void writeJsonIssues(JsonWriter json, Set<RuleKey> ruleKeys, Set<String>
     json.name("issues").beginArray();
     for (TrackedIssue issue : getIssues()) {
       if (issue.resolution() == null) {
+        InputComponent component = componentStore.getByKey(issue.componentKey());
+        String componentKey = getModule(component).definition().getKeyWithBranch();
```
janos-ss: It seems this assignment should go in the `else` block of the next `if` statement

sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/DefaultComponentTree.java
henryju: No. Because it returns null in case entry was missing.

sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/report/JSONReport.java
henryju: Nope: first I get module key. And then in case it is a dir or a file, I'll append relativepath to form the final key.

sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/DefaultComponentTree.java
janos-ss: If the entry is missing it will put a new `LinkedHashSet`. It will have the same outcome as what you wrote, but in case the entry is present, it won't create a new `LinkedHashSet` unnecessarily.

sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/report/JSONReport.java
janos-ss: Ok I see, the `componentKey` is always needed.

sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/DefaultComponentTree.java
henryju: Sorry I've read too quickly. I though you proposed to use `children.putIfAbsent(parent, new LinkedHashSet<>()).add(component);`

PR #1737, opened 2017-03-02
SONAR-8875 Add language facet on projects page

gregaubert:
The following additional filtering criterion should be added:
 * Project languages (Java, JavaScript, PHP, ... )
 * This filter should offer all the languages with LoC in the instance.
 * The bar graph on the right of the filter should represent the count of projects with LoC of that language
 * Multiple languages should be selectable at once as OR criteria

### DIFF
--- a/server/sonar-web/src/main/js/apps/projects/filters/Filter.js
+++ b/server/sonar-web/src/main/js/apps/projects/filters/Filter.js
@@ -44,6 +44,27 @@ export default class Filter extends React.Component {
     halfWidth: false
   };
 
+  isSelected (option) {
+    const { value } = this.props;
+    return Array.isArray(value) ? value.includes(option) : option === value;
+  }
+
+  getPath (option) {
+    const { property, value } = this.props;
+    let urlOption;
+
+    if (Array.isArray(value)) {
+      if (this.isSelected(option)) {
+        urlOption = value.length > 1 ? value.filter(val => val !== option).join(',') : null;
+      } else {
+        urlOption = value.concat(option).join(',');
+      }
+    } else {
+      urlOption = this.isSelected(option) ? null : option;
+    }
+    return this.props.getFilterUrl({ [property]: urlOption });
+  }
+
   renderHeader () {
     return (
         <div className="search-navigator-facet-header projects-facet-header">
@@ -66,22 +87,20 @@ export default class Filter extends React.Component {
   }
 
   renderOption (option) {
-    const { property, value, facet, getFacetValueForOption } = this.props;
+    const { facet, getFacetValueForOption } = this.props;
     const className = classNames('facet', 'search-navigator-facet', 'projects-facet', {
-      'active': option === value,
+      'active': this.isSelected(option),
       'search-navigator-facet-half': this.props.halfWidth
     }, this.props.optionClassName);
 
-    const path = option === value ?
-        this.props.getFilterUrl({ [property]: null }) :
-        this.props.getFilterUrl({ [property]: option });
+    const path = this.getPath(option);
 
     const facetValue = (facet && getFacetValueForOption) ? getFacetValueForOption(facet, option) : null;
 
     return (
         <Link key={option} className={className} to={path} data-key={option}>
           <span className="facet-name">
-            {this.props.renderOption(option, option === value)}
+            {this.props.renderOption(option, this.isSelected(option))}
           </span>
           {facetValue != null && (
               <span className="facet-stat">
--- a/server/sonar-web/src/main/js/apps/projects/filters/Filter.js
+++ b/server/sonar-web/src/main/js/apps/projects/filters/Filter.js
@@ -26,7 +26,7 @@ export default class Filter extends React.Component {
   static propTypes = {
     value: React.PropTypes.any,
     property: React.PropTypes.string.isRequired,
-    options: React.PropTypes.array.isRequired,
+    getOptions: React.PropTypes.func.isRequired,
     maxFacetValue: React.PropTypes.number,
     optionClassName: React.PropTypes.string,
 
@@ -96,7 +96,7 @@ export default class Filter extends React.Component {
   renderOptions () {
     return (
         <div className="search-navigator-facet-list">
-          {this.props.options.map(option => this.renderOption(option))}
+          {this.props.getOptions(this.props.facet).map(option => this.renderOption(option))}
         </div>
     );
   }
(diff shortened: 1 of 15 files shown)


server/sonar-web/src/main/js/apps/projects/filters/LanguageFilter.js
```diff
@@ -0,0 +1,48 @@
...
+import LanguageFilterOption from './LanguageFilterOption';
+
+export default class LanguageFilter extends React.Component {
+  renderOption = option => {
+    return (
+      <LanguageFilterOption languageKey={option}/>
+    );
+  };
+
+  getFacetValueForOption = (facet, option) => {
```
SonarTech: MINOR Remove curly braces and "return" from this arrow function body. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=javascript%3AS3524)

server/sonar-web/src/main/js/apps/projects/filters/LanguageFilterOption.js
```diff
@@ -0,0 +1,36 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import { connect } from 'react-redux';
+import React from 'react';
+
+class LanguageFilterOption extends React.Component {
+  render () {
+    const languageName = this.props.language ? this.props.language.name : this.props.languageKey;
+    return (
+      <span>{languageName}</span>
```
stas-vilchik: @gregaubert you need to know that we use a special value `<null>` for unknown language. In this case you need to display `translate('unknown')`.

server/sonar-web/src/main/js/apps/projects/filters/Filter.js
```diff
@@ -96,7 +115,7 @@ export default class Filter extends React.Component {
   renderOptions () {
     return (
         <div className="search-navigator-facet-list">
-          {this.props.options.map(option => this.renderOption(option))}
+          {this.props.getOptions(this.props.facet).map(option => this.renderOption(option))}
```
stas-vilchik: @gregaubert because now the list of options is dynamic, it can be empty. In this case I'd display a message like "no results found".

server/sonar-web/src/main/js/apps/projects/filters/LanguageFilterOption.js
gregaubert: I should display it instead of the language key ? So we could have 3 `Unknown` languages  with different stats in the facet, is that alright ?

stas-vilchik: It should be displayed only if the project key is exactly `<null>`.

server/sonar-web/src/main/js/apps/projects/filters/LanguageFilterOption.js
```diff
@@ -0,0 +1,36 @@
...
+    );
+  }
+}
+
+const mapStateToProps = (state, ownProps) => ({
+  language: state.languages[ownProps.languageKey]
```
gregaubert: I found `getLanguages` and was wondering what was the `key` parameter for ?
```js
export const getLanguages = (state, key) => (
    fromLanguages.getLanguages(state.languages, key)
);
```
It seems like it is not implemented nor used. Should I remove this parameter when I create `getLanguageByKey` ?

server/sonar-web/src/main/js/apps/projects/store/utils.js
```diff
@@ -32,14 +32,29 @@ const getAsLevel = value => {
...
 };
 
+const getAsString = value => {
+  if (!value) {
+    return null;
+  }
+  return value;
+};
+
+const getAsArray = getAsType => (values => {
```
gregaubert: yeah why not... I don't know why I did it weird like that! I'll change it.

server/sonar-web/src/main/js/apps/projects/store/utils.js
```diff
@@ -136,6 +151,14 @@ export const convertToFilter = (query, isFavorite) => {
     conditions.push(convertIssuesRating('sqale_rating', query['maintainability']));
   }
 
+  if (query['language'] != null) {
+    if (!Array.isArray(query['language']) || query['language'].length < 2) {
```
gregaubert: They proposed both on the api doc, so I thought maybe there are some performance improvements server side using the `=` syntax for 1 element.
But I think I'll still need to check if 'query['language']` is an array or not so I'm not sure I can completely remove the condition.

server/sonar-web/src/main/js/apps/projects/filters/LanguageFilterOption.js
stas-vilchik: yeah, I think this is a typo that there is this {{key}} parameter, you can clean it up

server/sonar-web/src/main/js/apps/projects/store/utils.js
stas-vilchik: ok, fine

server/sonar-web/src/main/js/apps/projects/filters/LanguageFilter.js
```diff
@@ -0,0 +1,52 @@
...
+    );
+  };
+
+  getFacetValueForOption = (facet, option) => facet[option];
+
+  render () {
+    return (
+      <FilterContainer
+          property="language"
+          getOptions={facet => facet ? Object.keys(facet) : []}
```
stas-vilchik: options should be sorted

server/sonar-web/src/main/js/apps/projects/filters/LanguageFilter.js
```diff
@@ -0,0 +1,52 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import React from 'react';
+import FilterContainer from './FilterContainer';
+import LanguageFilterOption from './LanguageFilterOption';
+
+export default class LanguageFilter extends React.Component {
+  static propTypes = {
+    query: React.PropTypes.string.isRequired,
```
stas-vilchik: you have a propTypes violation

> Warning: Failed prop type: Invalid prop `query` of type `object` supplied to `LanguageFilter`, expected `string`.

stas-vilchik:
@gregaubert LGTM!

PR #1741, opened 2017-03-03
SONAR-8876 Language facet searchbox on projects page

gregaubert:
@julienlancelot We will be able to verify if it really work when the api is updated to return languages that are in the filter but not top 6.


server/sonar-web/src/main/js/apps/projects/filters/LanguageFilterFooter.js
```diff
@@ -0,0 +1,84 @@
...
+    if (facet) {
+      options = difference(options, Object.keys(facet));
+    }
+    return options.map(key => ({ label: languages[key].name, value: key }));
+  }
+
+  render () {
+    return (
+      <Select
+          onChange={option => this.handleLanguageChange(option.value)}
```
stas-vilchik: let's just pass `onChange={this.handleLanguageChange}`

PR #1746, opened 2017-03-04
Fix check of version in IssueJsonReportTest

simonbrandhof:


it/it-tests/src/test/java/it/analysis/IssueJsonReportTest.java
```diff
@@ -276,9 +279,13 @@ public void test_json_report_on_root_module() throws Exception {
     JSONAssert.assertEquals(expectedJson, json, false);
   }
 
-  private String expected(String path) throws IOException, FileNotFoundException {
-    return sanitize(IOUtils.toString(getResourceInputStream(path)))
-      .replaceAll(Pattern.quote(SONAR_VERSION_PLACEHOLDER), orchestrator.getServer().version().toString());
+  private String expected(String path) throws IOException {
+    try (WsResponse response = newWsClient(orchestrator).wsConnector().call(new GetRequest("api/server/version"))) {
```
Godin: @simonbrandhof yesterday I realized that `orchestrator.getServer().version()` extracts version from JAR (`sonar-application-<version>.jar`) and made corresponding fixes for build with Gradle.

So given this change, out of curiosity: does it mean that there is inconsistency between responce of `api/server/version`, version extracted from JAR, version in JSON?

PR #1752, opened 2017-03-06
SONAR-8840 Create WS api/project_tags/search to search for tags

teryk:


server/sonar-server/src/test/java/org/sonar/server/projecttag/ws/SetActionTest.java
```diff
@@ -51,20 +56,24 @@
   @Rule
   public DbTester db = DbTester.create();
   private DbClient dbClient = db.getDbClient();
-  public WsActionTester ws = new WsActionTester(new SetAction(dbClient, new ComponentFinder(dbClient), userSession));
   private DbSession dbSession = db.getSession();
   private ComponentDto project;
 
+  private ProjectIndexer indexer = mock(ProjectIndexer.class);
```
julienlancelot: Unit tests of WS should not rely on mocks for DB/Index

server/sonar-server/src/main/java/org/sonar/server/projecttag/ws/SearchAction.java
```diff
@@ -0,0 +1,63 @@
...
+import org.sonar.api.server.ws.Response;
+import org.sonar.api.server.ws.WebService;
+import org.sonar.server.measure.index.ProjectMeasuresIndex;
+import org.sonarqube.ws.WsProjectTags;
+
+import static org.sonar.api.server.ws.WebService.Param.PAGE_SIZE;
+import static org.sonar.api.server.ws.WebService.Param.TEXT_QUERY;
+import static org.sonar.server.ws.WsUtils.writeProtobuf;
+
+public class SearchAction implements ProjectTagsWsAction {
```
julienlancelot: This class is missing in ProjectTagsWsModule => The WS is not available.

server/sonar-server/src/main/java/org/sonar/server/measure/index/ProjectMeasuresIndex.java
```diff
@@ -304,6 +312,34 @@ private static QueryBuilder toValueQuery(MetricCriterion criterion) {
     }
   }
 
+  public List<String> searchTags(@Nullable String textQuery, int pageSize) {
```
julienlancelot: Why only pageSize ? Either you should have page and pageSize, either you should return everything

server/sonar-server/src/test/java/org/sonar/server/projecttag/ws/SearchActionTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+  @Test
+  public void json_example() throws IOException {
+    when(index.searchTags(anyString(), anyInt())).thenReturn(newArrayList("official", "offshore", "playoff"));
+
+    String result = ws.newRequest().execute().getInput();
+
+    assertJson(ws.getDef().responseExampleAsString()).withStrictArrayOrder().isSimilarTo(result);
+  }
+
```
julienlancelot: Unit tests missing on some common use cases : search with text query, with page size, etc.

server/sonar-server/src/main/java/org/sonar/server/projecttag/ws/SearchAction.java
teryk: 😭

server/sonar-server/src/main/java/org/sonar/server/measure/index/ProjectMeasuresIndex.java
teryk: it's the same for rules and issues tags. You cannot have pagination for an ES facet. That's what we're using here to return tags. Still, we want to be able to change the number of tags returned.

server/sonar-server/src/test/java/org/sonar/server/projecttag/ws/SearchActionTest.java
teryk: WDYM? Isn't it tested in ProjectMeasuresIndex? I don't want to test the same thing twice

server/sonar-server/src/test/java/org/sonar/server/projecttag/ws/SetActionTest.java
teryk: ok, main use case tested end to end

teryk:
tickets for WS api/issues/tags and api/rules/tags have been created

server/sonar-server/src/main/java/org/sonar/server/measure/index/ProjectMeasuresIndex.java
```diff
@@ -304,6 +312,34 @@ private static QueryBuilder toValueQuery(MetricCriterion criterion) {
...
+    checkArgument(pageSize <= 100, "Page size must be lower than or equals to " + 100);
+    if (pageSize == 0) {
```
julienlancelot: Other WS on rules and issues tags do not have the same behavior => we should fix them (create JIRA tickets) or change this one

teryk: the ticket have already been created

PR #1756, opened 2017-03-07
SONAR-8840 Restrict project tags search to one organization

teryk:


server/sonar-server/src/main/java/org/sonar/server/exceptions/NotFoundException.java
```diff
@@ -19,11 +19,12 @@
...
 package org.sonar.server.exceptions;
 
+import static java.lang.String.format;
 import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 
 public class NotFoundException extends ServerException {
 
-  public NotFoundException(String message) {
-    super(HTTP_NOT_FOUND, message);
+  public NotFoundException(String message, Object... messageArguments) {
```
julienlancelot: Why ? Because as you know, if the message contains '%' something without arguments, the WS will badly fail....
If you want to keep it like this, you should forbid having no args => but It makes no sense in my opinion.

server/sonar-server/src/test/java/org/sonar/server/projecttag/ws/SearchActionTest.java
```diff
@@ -86,12 +103,33 @@ public void search_by_query_and_page_size() {
...
       newDoc().setTags(newArrayList("offshore", "playoff")));
 
-    SearchResponse result = call("off", 2);
+    SearchResponse result = call(ORG.getKey(), "off", 2);
 
     assertThat(result.getTagsList()).containsOnly("offshore", "playoff");
   }
 
   @Test
+  public void search_on_default_org_is_not_provided() {
```
julienlancelot: typo => `search_on_default_org_if_not_provided`

PR #1760, opened 2017-03-09
SONAR-8452 open the last state of projects page

stas-vilchik:


stas-vilchik:
@gregaubert please review :)

stas-vilchik:
@gregaubert the bug you showed me is fixed

PR #1766, opened 2017-03-09
SONAR-8878 Add ut and it for Language facet, sorting and projects search

gregaubert:


it/it-tests/src/test/java/pageobjects/projects/FacetItem.java
```diff
@@ -38,4 +38,14 @@ public FacetItem shouldHaveValue(String key, String value) {
   public void selectValue(String key) {
     this.elt.$(".facet[data-key=\"" + key + "\"]").click();
   }
+
+  public FacetItem selectOptionItem(String value) {
+    this.elt.$(".Select-input input").val(value).pressEnter();
+    return this;
+  }
+
+  public FacetItem sortList(Integer idx) {
```
stas-vilchik: I'd not expose the fact that the first link sorts ascending, the second one descending.

Instead let's expose two methods: `sortListAsc` and `sortListDesc`. Internally they can rely on this *private* method.

gregaubert: Ok I'll change it, but I don't think I can rely on that method then. Because not every facets have the left button that does a descending sort and the right button an ascending sort.

it/it-tests/src/test/java/pageobjects/projects/FacetItem.java
```diff
@@ -44,8 +46,23 @@ public FacetItem selectOptionItem(String value) {
...
+    this.getSortingButton("-").click();
     return this;
   }
+
+  public FacetItem sortListAsc() {
+    this.getSortingButton("[a-zA-Z ]").click();
+    return this;
+  }
+
+  public SelenideElement getSortingButton(String selector) {
```
gregaubert: Ok I'll update it and merge

PR #1773, opened 2017-03-10
Handle membership when creating/removing user

julienlancelot:
- SONAR-8901 Associate new users to organizations
- SONAR-8930 When removing a user, remove him from any organization he belongs to

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/organization/OrganizationCreationImpl.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/organization/OrganizationCreationImpl.java
@@ -33,6 +33,7 @@ import org.sonar.db.DbClient;
 import org.sonar.db.DbSession;
 import org.sonar.db.organization.DefaultTemplates;
 import org.sonar.db.organization.OrganizationDto;
+import org.sonar.db.organization.OrganizationMemberDto;
 import org.sonar.db.permission.GroupPermissionDto;
 import org.sonar.db.permission.OrganizationPermission;
 import org.sonar.db.permission.UserPermissionDto;
@@ -105,6 +106,7 @@ public class OrganizationCreationImpl implements OrganizationCreation {
     OrganizationPermission.all()
       .forEach(p -> insertUserPermissions(dbSession, newUser, organization, p));
     insertPersonalOrgDefaultTemplate(dbSession, organization);
+    insertOrganizationMember(dbSession, organization, newUser);
 
     dbSession.commit();
 
@@ -251,5 +253,11 @@ public class OrganizationCreationImpl implements OrganizationCreation {
     dbClient.userGroupDao().insert(
       dbSession,
       new UserGroupDto().setGroupId(group.getId()).setUserId(createUserId));
+  }
+
+  private void insertOrganizationMember(DbSession dbSession, OrganizationDto organizationDto, UserDto userDto) {
+    dbClient.organizationMemberDao().insert(dbSession, new OrganizationMemberDto()
+      .setOrganizationUuid(organizationDto.getUuid())
+      .setUserId(userDto.getId()));
   }
 }
--- a/server/sonar-server/src/main/java/org/sonar/server/user/UserUpdater.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/user/UserUpdater.java
@@ -386,14 +386,15 @@ public class UserUpdater {
     }
     String defOrgUuid = defaultOrganizationProvider.get().getUuid();
     List<GroupDto> userGroups = dbClient.groupDao().selectByUserLogin(dbSession, userDto.getLogin());
-    if (!userGroups.stream().anyMatch(g -> defOrgUuid.equals(g.getOrganizationUuid()) && g.getName().equals(defaultGroupName))) {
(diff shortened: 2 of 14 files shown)


teryk:
When removing a user, do you remove the personal organization? What is the logic to keep an personal organization without an activated user? 
Remember the user_id column in the organizations table?
It may be addressed but it's not obvious from the commits.

julienlancelot:
You're talking about something that is not related to this MMF => this should have been done with the introduction of personal org but has been forgotten so we'll handled it in the MMF about closing its own account, the related ticket is https://jira.sonarsource.com/browse/SONAR-8929

server/sonar-server/src/main/java/org/sonar/server/organization/OrganizationCreationImpl.java
```diff
@@ -252,4 +254,10 @@ private void addCurrentUserToGroup(DbSession dbSession, GroupDto group, int crea
       dbSession,
       new UserGroupDto().setGroupId(group.getId()).setUserId(createUserId));
   }
+
+  private void insertOrganizationMember(DbSession dbSession, OrganizationDto organizationDto, UserDto userDto) {
```
teryk: what is this OrganizationCreation class? Is it similar to the Updater we had? I know you didn't create the class

server/sonar-server/src/main/java/org/sonar/server/user/UserUpdater.java
```diff
@@ -386,14 +386,15 @@ private void addDefaultGroup(DbSession dbSession, UserDto userDto) {
...
     List<GroupDto> userGroups = dbClient.groupDao().selectByUserLogin(dbSession, userDto.getLogin());
-    if (!userGroups.stream().anyMatch(g -> defOrgUuid.equals(g.getOrganizationUuid()) && g.getName().equals(defaultGroupName))) {
-      Optional<GroupDto> groupDto = dbClient.groupDao().selectByName(dbSession, defOrgUuid, defaultGroupName);
-      if (!groupDto.isPresent()) {
-        throw new ServerException(HttpURLConnection.HTTP_INTERNAL_ERROR,
-          format("The default group '%s' for new users does not exist. Please update the general security settings to fix this issue.",
-            defaultGroupName));
-      }
-      dbClient.userGroupDao().insert(dbSession, new UserGroupDto().setUserId(userDto.getId()).setGroupId(groupDto.get().getId()));
+    if (userGroups.stream().anyMatch(g -> defOrgUuid.equals(g.getOrganizationUuid()) && g.getName().equals(defaultGroupName))) {
```
teryk: suggestion: extract method to describe its purpose. I know you didn't write this code

server/sonar-server/src/main/java/org/sonar/server/user/UserUpdater.java
```diff
@@ -386,14 +386,15 @@ private void addDefaultGroup(DbSession dbSession, UserDto userDto) {
...
+      return;
+    }
+    Optional<GroupDto> groupDto = dbClient.groupDao().selectByName(dbSession, defOrgUuid, defaultGroupName);
```
teryk: use `orElseThrow`

server/sonar-server/src/main/java/org/sonar/server/organization/OrganizationCreationImpl.java
julienlancelot: It's a class that handle creation of Organization

server/sonar-server/src/test/java/org/sonar/server/organization/OrganizationCreationImplTest.java
```diff
@@ -348,6 +348,19 @@ public void createForUser_creates_default_template_for_new_organization() throws
   }
 
   @Test
+  public void createForUser_add_current_user_as_member_of_organization() throws OrganizationCreation.KeyConflictException {
+    UserDto user = dbTester.users().insertUser(dto -> dto.setLogin(A_LOGIN).setName(A_NAME));
+    when(organizationValidation.generateKeyFrom(A_LOGIN)).thenReturn(SLUG_OF_A_LOGIN);
+    mockForSuccessfulInsert(SOME_UUID, SOME_DATE);
+    enableCreatePersonalOrg(true);
+
+    underTest.createForUser(dbSession, user);
```
teryk: question: are you comfortable with the mocks in this case?

server/sonar-server/src/main/java/org/sonar/server/organization/OrganizationCreationImpl.java
teryk: ok, I wasn't explicit enough apparently. If this class has the same responsibility and spirit than XUpdater classes, then it should be named accordingly. We don't need several names for the same thing.
I asked the question to confirm I wasn't missing something

server/sonar-server/src/main/java/org/sonar/server/user/UserUpdater.java
```diff
@@ -89,14 +89,6 @@ public UserUpdater(NewUserNotifier newUserNotifier, Settings settings, DbClient
     this.organizationCreation = organizationCreation;
   }
 
-  public UserDto create(NewUser newUser) {
```
teryk: 👍

PR #1774, opened 2017-03-10
format code using prettier

stas-vilchik:


gregaubert:
I think we can update and merge that now that the pull request about tags has been merged

stas-vilchik:
Yes, will do.

PR #1776, opened 2017-03-13
SONAR-8521 Improve size, duplication and coverage facets style

gregaubert:


server/sonar-web/src/main/js/apps/projects/filters/Filter.js
```diff
@@ -40,19 +40,25 @@ export default class Filter extends React.PureComponent {
     getFacetValueForOption: React.PropTypes.func,
 
     halfWidth: React.PropTypes.bool,
-    highlightUnder: React.PropTypes.func
+    highlightUnder: React.PropTypes.any
```
stas-vilchik: I'm not comfortable with `any`. It's hard to tell what the component expects as an input. Can you stick to a function please?

gregaubert: I think I can just put `number`. I was using `any` because I wanted a number or the default boolean, but in the end I just check if the value exists so I don't need the default boolean value.

server/sonar-web/src/main/js/apps/projects/filters/Filter.js
```diff
@@ -40,19 +40,24 @@ export default class Filter extends React.PureComponent {
...
+    halfWidth: false
   };
 
   isSelected (option) {
     const { value } = this.props;
     return Array.isArray(value) ? value.includes(option) : option === value;
   }
 
+  highlightUnder (option) {
+    return typeof this.props.highlightUnder === 'number' &&
```
stas-vilchik: and now I'd simplify the condition to `this.props. highlightUnder  != null`

PR #1786, opened 2017-03-14
SONAR-7860 Remove CodeColorizerFormat API

dbmeneses:


simonbrandhof:
I wonder if `server/sonar-web/src/main/less/sonar-colorizer.less` can be dropped too.
cc @stas-vilchik

stas-vilchik:
@simonbrandhof of course not, the file is actually controlling the code styles.

PR #1790, opened 2017-03-14
SONAR-8922 Add tags facet on project page

gregaubert:


server/sonar-web/src/main/js/apps/projects/filters/SearchFilterContainer.js
```diff
@@ -25,7 +26,7 @@ import SearchFilter from './SearchFilter';
 
 type Props = {
   query: {},
-  router: { push: (string) => void },
+  router: { push: (path: string, query?: {}) => void },
```
stas-vilchik: I don't think you need `query` here

server/sonar-web/src/main/js/apps/projects/filters/TagsFilterContainer.js
```diff
@@ -0,0 +1,33 @@
...
+  getProjectsAppFacetByProperty,
+  getProjectsAppMaxFacetValue
+} from '../../../store/rootReducer';
+
+const mapStateToProps = (state, ownProps) => ({
+  value: ownProps.query['tags'],
+  facet: getProjectsAppFacetByProperty(state, 'tags'),
+  maxFacetValue: getProjectsAppMaxFacetValue(state)
+});
+export default connect(mapStateToProps)(withRouter(TagsFilter));
```
gregaubert: mmmh I don't think so, because `TagsFilter` needs at least the `facet` props to be able to compute the list of options for the searchbox...

PR #1801, opened 2017-03-16
SONAR-8843 Add the tags on the projects page and project homepage

gregaubert:


server/sonar-web/src/main/js/components/ui/TagsList.css
```diff
@@ -0,0 +1,29 @@
...
+  font-size: 11px;
+  padding-left: 4px;
+  opacity: 0.6;
+  display: inline-block;
+  max-width: 220px;
+  vertical-align: text-top;
+  margin-top: 2px;
+}
+
+.single-line {
```
stas-vilchik: you can partially reuse the class `.text-ellipsis`

server/sonar-web/src/main/js/components/ui/TagsList.css
```diff
@@ -0,0 +1,29 @@
...
+
+.tags-list i {
+  padding-left: 4px;
+}
+
+.tags-list i::before {
+  font-size: 12px;
+}
+
+.tags-list span {
```
stas-vilchik: I think `.note` is enough here, it gives you 12px font-size and the color.

server/sonar-web/src/main/js/components/ui/TagsList.js
```diff
@@ -0,0 +1,59 @@
...
+import './TagsList.css';
+
+type Props = {
+  tags?: Array<string>,
+  allowUpdate?: boolean,
+  hideWhenNoTags?: boolean,
+  allowMultiLine?: boolean
+};
+
+export default class SearchableFilterFooter extends React.PureComponent {
```
stas-vilchik: change the class name

server/sonar-web/src/main/js/apps/projects/components/ProjectCard.js
```diff
@@ -74,6 +78,7 @@ export default class ProjectCard extends React.PureComponent {
               {project.name}
             </Link>
           </h2>
+          <TagsList tags={this.props.project && this.props.project.tags} hideWhenNoTags={true}/>
```
stas-vilchik: you don't need to check `this.props.project`, it is already done before (line 49)

server/sonar-web/src/main/js/apps/projects/components/ProjectCard.js
```diff
@@ -35,7 +36,10 @@ export default class ProjectCard extends React.PureComponent {
     project?: {
       analysisDate?: string,
       key: string,
-      name: string
+      name: string,
+      isFavorite?: boolean,
+      organization?: string,
+      tags?: Array<string>
```
stas-vilchik: can you add a test with tags please?

stas-vilchik: I believe the WS always returns an array (possibly empty), so you can drop `?`

server/sonar-web/src/main/js/components/ui/TagsList.js
```diff
@@ -0,0 +1,59 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+// @flow
+import React from 'react';
+import classNames from 'classnames';
+import { translate } from '../../helpers/l10n';
+import './TagsList.css';
+
+type Props = {
```
stas-vilchik: remove all `?` and add default values for all booleans, it's better for typing

stas-vilchik: based on my experience, it's better to extract the checking of `tags` existence outside of the component. That keeps you component much simpler

server/sonar-web/src/main/js/components/ui/TagsList.css
gregaubert: Ok I'll remove the `font-size: 11px` in `.tags-list span` but I don't think I can use `.note` for the icon.

server/sonar-web/src/main/js/apps/projects/components/ProjectCard.js
gregaubert: In the end I didn't drop the `?` because for the moment we have a case where the `tags` doesn't exists. It's when I click on a project without tags to see its detail and then come back to the projects page.

PR #1824, opened 2017-03-21
Deprecate WsTester, advice to use WsActionTester instead

bartfastiel:


simonbrandhof:
The perfect strategy would be to drop the class and refactor all the tests using it :-)

PR #1835, opened 2017-03-22
use "restore_rofile" from sonar ws in ITs

bartfastiel:


it/it-tests/src/test/java/util/ItUtils.java
```diff
@@ -338,9 +341,35 @@ public static ComponentNavigation getComponentNavigation(Orchestrator orchestrat
     return ComponentNavigation.parse(content);
   }
 
-  public static File findFileInClasspath(String classpathRelativeFilename) throws URISyntaxException {
-    ResourceLocation location = FileLocation.ofClasspath("/organization/IssueAssignTest/one-issue-per-file-profile.xml");
-    return new File(ItUtils.class.getResource(location.getPath()).toURI());
+  public static void restoreProfile(Orchestrator orchestrator, String classpathRelativeFilename) {
```
simonbrandhof: The String parameter is misleading. I suggest to replace it by URL so that caller does simply:
```
restoreProfile(orchestrator, getClass().getResource("FooTest/profile.xml"));
```
(or `getClass().getResourceAsStream("FooTest/profile.xml")`). That would allow to remove the method `findFileInClasspath()` and to allow setting byte[] or Reader types to `RestoreWsRequest`.

it/it-tests/src/test/java/util/ItUtils.java
```diff
@@ -338,9 +341,35 @@ public static ComponentNavigation getComponentNavigation(Orchestrator orchestrat
...
+    adminWsClient
+      .qualityProfiles()
+      .restoreProfile(
+        RestoreWsRequest.builder()
+          .setBackup(findFileInClasspath(classpathRelativeFilename))
+          .setOrganization(organization)
+          .build());
+  }
+
+  public static File findFileInClasspath(String classpathRelativeFilename) {
```
simonbrandhof: Using `FileLocation` does not bring any value here. Using `Class#getResource()` is good enough for tests.

it/it-tests/src/test/java/util/ItUtils.java
bartfastiel: Done.

it/it-tests/src/test/java/util/ItUtils.java
bartfastiel: Done.

it/it-tests/src/test/java/util/ItUtils.java
```diff
@@ -338,9 +340,28 @@ public static ComponentNavigation getComponentNavigation(Orchestrator orchestrat
...
+          .setOrganization(organization)
+          .build());
+  }
+
+  private static File resourceToFile(URL resource) {
+    URI uri;
+    try {
+      uri = resource.toURI();
+    } catch (URISyntaxException e) {
+      throw new IllegalArgumentException("Cannot find quality profile xml file '" + resource + "' in classpath");
```
simonbrandhof: this method is generic, so message should not be specific to quality profiles. I suggest `Cannot find '" + resource + "' in classpath

it/it-tests/src/test/java/it/qualityProfile/QualityProfilesRestoreAndSearchTest.java
```diff
@@ -0,0 +1,57 @@
...
+  @Before
+  public void init() {
+    orchestrator.resetData();
+  }
+
+  @Test
+  public void restore_and_search_in_default_organization() {
+    ItUtils.restoreProfile(orchestrator, getClass().getResource("/authorisation/one-issue-per-line-profile.xml"));
+    QualityProfiles.SearchWsResponse results = ItUtils.newAdminWsClient(orchestrator).qualityProfiles().search(new SearchWsRequest());
+    System.out.println(results.getProfilesList());
```
simonbrandhof: System.out should be removed.

PR #1839, opened 2017-03-22
SONAR-9002 Add a log when a node is waiting for initialization

ehartmann:


server/sonar-process-monitor/src/main/java/org/sonar/application/SchedulerImpl.java
```diff
@@ -117,6 +117,8 @@ private void tryToStartWeb() {
       tryToStartProcess(process, () -> javaCommandFactory.createWebCommand(false));
     } else if (appState.tryToLockWebLeader()) {
       tryToStartProcess(process, () -> javaCommandFactory.createWebCommand(true));
+    } else {
+      LOG.info("Waiting for back-end initialization from " + appState.getLeaderHostName());
```
simonbrandhof: I think the term "back-end" can be removed.

ehartmann:
QA seems to be a false positive.

PR #1847, opened 2017-03-23
add flow-typed definitions

stas-vilchik:


server/sonar-web/src/main/js/apps/projects/filters/SearchFilter.js
```diff
@@ -63,7 +63,7 @@ export default class SearchFilter extends React.PureComponent {
   render() {
     const { userQuery } = this.state;
     const inputClassName = classNames('input-super-large', {
-      touched: userQuery && userQuery.length < 2
+      touched: userQuery != null && userQuery.length < 2
```
gregaubert: :thinking: Why is this better than just checking for a truthy value ?

stas-vilchik: type definition of the `classNames` function tells that it expects `boolean` here

gregaubert: oh okay! :+1:

PR #1852, opened 2017-03-24
SONAR-9006 SONAR-8864 add canCustomizeRule and canCreateCustomerRule to rules/app

sns-seb:


server/sonar-server/src/test/java/org/sonar/server/rule/ws/AppActionTest.java
```diff
@@ -209,6 +209,51 @@ public void canWrite_is_false_if_user_is_not_profile_administrator_of_default_or
     assertJson(json).isSimilarTo("{ \"canWrite\": false }");
   }
 
+  @Test
+  public void canCustomizeRule_is_true_if_user_is_profile_administrator_of_default_organization_and_no_organization_is_specified() {
+    userSession.addPermission(OrganizationPermission.ADMINISTER_QUALITY_PROFILES, db.getDefaultOrganization());
+
```
bartfastiel: _minor code style_ 🐜 
These blank lines should probably separate test setup, test execution and assertions. If each of those parts is as short as in this case, the blank lines could be avoided in my opinion.

server/sonar-server/src/main/java/org/sonar/server/rule/ws/AppAction.java
```diff
@@ -88,6 +88,7 @@ private void addPermissions(OrganizationDto organization, JsonWriter json) {
     boolean isAdminister = userSession.hasPermission(OrganizationPermission.ADMINISTER_QUALITY_PROFILES, organization);
     json.prop("canWrite", isAdminister);
     json.prop("canCustomizeRule", isAdminister && organization.getUuid().equals(defaultOrganizationProvider.get().getUuid()));
+    json.prop("canCreateCustomRule", isAdminister && organization.getUuid().equals(defaultOrganizationProvider.get().getUuid()));
```
bartfastiel: Don't repeat yourself

PR #1857, opened 2017-03-27
SONAR-8844 Fix project tags selector

gregaubert:
* Limit list size to 10 elements max
* Fix long tags display


server/sonar-web/src/main/less/components/menu.less
```diff
@@ -78,6 +78,8 @@
   padding: 4px 16px 0;
 
   .search-box-input { font-size: @smallFontSize; }
+  
+  .search-box-submit { vertical-align: baseline; }
```
stas-vilchik: I have an utility class `abs-width-300`, you can use it

PR #1858, opened 2017-03-27
MMF-721 apply feedback

stas-vilchik:


gregaubert:
Ok I just saw the color legend, all good!

PR #1867, opened 2017-03-29
Small fixes relating to project tags

teryk:
* restrict tag to project only
* typo in api/components/search_projects description


server/sonar-server/src/main/java/org/sonar/server/projecttag/ws/SetAction.java
```diff
@@ -92,7 +93,7 @@ public void handle(Request request, Response response) throws Exception {
 
     try (DbSession dbSession = dbClient.openSession(false)) {
       ComponentDto project = componentFinder.getByKey(dbSession, projectKey);
-      checkRequest(project.isRootProject(), "Component must be a project");
+      checkRequest(Qualifiers.PROJECT.equals(project.qualifier()), "Component '%s' is not a project", project.key());
```
julienlancelot: Please use static import

PR #1879, opened 2017-03-31
SONAR-8994 Fix groups count when the user is just added

gregaubert:


gregaubert:
I added the IT for this particular case and hopefully fixed the other ITs that failed sometime.

PR #1889, opened 2017-04-03
Automatically remove nil values from api calls

gregaubert:


gregaubert:
There is  a problem with the `/api/measures/seach` that I still need to fix...

gregaubert:
Ok... :thinking: I'm not sure it's a good idea anymore...
The problem is related to the `q` parameter we have in a lot of apis... Some of our api won't accept an empty string as value for this parameter and others do... So with the previous way we were doing things there was no problem because we almost always checked the params like : 
```
if (query) {
data.q = query
}
```
But now, I tried to check which api required a value and which not and changed the code accordingly but it's feels a bit "messy"... I don't know if there is a better way of doing things ? Or if there are other parameters that have the same problems ?

gregaubert:
Maybe a better solution than doing this under the hood is to create a helper function like for exemple `filterNil` that we would use when we want. That would allow us to still handle some special cases if needed.

Another point is to maybe also filter empty string along with nil, I can't see a real reason to send an empty parameter like `?q=&otherparam=test`. And it would fix the problem with `q` parameters.

gregaubert:
Ok, as it is now we can't filter empty string by default because we actually have some "valid" use cases : 
* Issues assign : the `assign` parameter is empty when you want to unassign
* Issues set tags : the `tags` parameter is a list, and an empty list means remove all tags (it's this way to prevent the users to remove all tags by mistake if the forget to define the tags parameter)

PR #1921, opened 2017-04-11
BUILD: setup ramfs for travis

tomverin:


travis.sh
```diff
@@ -1,6 +1,28 @@
...
+printf "${RED}disk size before build${NC}\n"
+df -h
+du -sh $HOME
+
+printf "${RED}move original home${NC}\n"
+sudo mv /home/travis /home/travis.ori
+printf "${RED}create ramdisk mount point${NC}\n"
+sudo mkdir -p /home/travis
+printf "${RED}create ramdisk${NC}\n"
+sudo mount -t tmpfs -o size=5120m ext2 /home/travis
```
ehartmann: @tomverin I was just having a eye on this when looking for another pull request.
I suppose you wanted to do something like :
sudo mount -t ramfs -o size=5120m ramfs /home/travis

Here is a good article between the differences : https://www.jamescoyle.net/knowledge/951-the-difference-between-a-tmpfs-and-ramfs-ram-disk
Using tmpfs may use the swap memory (ie. hard drive) if there is no enough memory and so this optimization may be useless in this case.

tomverin: I tried both to see the differences, thanks Eric

PR #1926, opened 2017-04-12
(Simon) minor hardening

simonbrandhof:


sonar-scanner-engine/src/main/java/org/sonar/scanner/rule/ActiveRulesProvider.java
```diff
@@ -57,19 +56,15 @@ private static ActiveRules load(ActiveRulesLoader loader, ModuleQProfiles qProfi
     Collection<String> qProfileKeys = getKeys(qProfiles);
     Map<RuleKey, LoadedActiveRule> loadedRulesByKey = new HashMap<>();
 
-    try {
-      for (String qProfileKey : qProfileKeys) {
-        Collection<LoadedActiveRule> qProfileRules;
-        qProfileRules = load(loader, qProfileKey);
+    for (String qProfileKey : qProfileKeys) {
```
julienlancelot: You could have used stream :)

simonbrandhof: I fixed only the issues found by SonarJava :-)

dbmeneses: I guess a lambda with forEach, not stream. But in case like this one personally I only see the advantage if it can turn into a one-liner (not the case).

PR #1907, opened 2017-04-06
MMF-838 make rules ws organization aware

bartfastiel:

PR #1934, opened 2017-04-13
SONAR-9115 let api/rules/create fail, if organizations are enabled

bartfastiel:
🌊  This PR is based on #1907, which should be reviewed first.


server/sonar-server/src/main/java/org/sonar/server/rule/ws/CreateAction.java
```diff
@@ -62,19 +63,22 @@
...
   }
 
   @Override
   public void define(WebService.NewController controller) {
     WebService.NewAction action = controller
       .createAction("create")
       .setDescription("Create a custom rule. <br/>" +
-        "Since 5.5, it's no more possible to create manual rule.")
+        "Since 5.5, it's no more possible to create manual rule." +
+        "Custom rules are currently not supported, if the organization feature is enabled. In that case, the webservice will fail.")
```
simonbrandhof: I would say "Custom rules are not supported if the organization feature is enabled.".

PR #1940, opened 2017-04-14
quality flaws

bartfastiel:


server/sonar-server/src/main/java/org/sonar/server/rule/index/RuleQuery.java
```diff
@@ -260,12 +260,11 @@ public String getRuleKey() {
     return ruleKey;
   }
 
-  @CheckForNull
   public String getOrganizationUuid() {
```
sns-seb: nothing actually prevents this method to return `null`
actually, `new RuleQuery().getOrganizationUuid()` returns `null`

PR #1945, opened 2017-04-14
6.4 hardening 1

julienlancelot:


server/sonar-server/src/test/java/org/sonar/server/authentication/JwtSerializerTest.java
```diff
@@ -74,12 +74,27 @@ public void generate_token_with_expiration_date() throws Exception {
...
+    assertThat(claims.getExpiration().getTime()).isGreaterThanOrEqualTo(now.getTime() + expirationTimeInSeconds * 1000L - 1000L);
+  }
+
+  @Test
+  public void generate_token_with_big_expiration_date() throws Exception {
+    setSecretKey(A_SECRET_KEY);
+    underTest.start();
+    Date now = new Date();
+
+    long oneYearInSeconds = 12 * 30 * 24 * 60 * 60L;
```
teryk: is it 3 months or one year the max authentication session timeout?
where have you put `sonar.web.sessionTimeoutInMinutes` to 129600 in sonar.properties

sonar-plugin-api/src/main/java/org/sonar/api/web/ServletFilter.java
```diff
@@ -102,6 +102,24 @@ public String getUrl() {
...
 
+    public String label() {
+      return "UrlPattern{" +
+        "inclusions=[" + convertPatternsToString(inclusions) + "]" +
+        ", exclusions=[" + convertPatternsToString(exclusions) + "]" +
+        '}';
+    }
+
+    private static String convertPatternsToString(List<String> input) {
+      StringBuilder toStringBuilder = new StringBuilder();
```
teryk: suggestion: rename to `output`

sonar-plugin-api/src/main/java/org/sonar/api/web/ServletFilter.java
```diff
@@ -102,6 +102,24 @@ public String getUrl() {
       throw new IllegalStateException("this method is deprecated and should not be used anymore");
     }
 
+    public String label() {
+      return "UrlPattern{" +
```
teryk: remark: there are string builder libraries in case you're interested

server/sonar-server/src/test/java/org/sonar/server/authentication/JwtSerializerTest.java
julienlancelot: As discussed, this class doesn't check the value of the session, so the limit should be LONG#MAX_VALUE

sonar-plugin-api/src/main/java/org/sonar/api/web/ServletFilter.java
julienlancelot: ok

PR #1946, opened 2017-04-18
Deprecate parameter “ps” in api/rules/tags

bartfastiel:
The page size parameter probably has not been used much, because the results were not ordered and there is no corresponding parameter "p".


bartfastiel:
@julienlancelot and me discussed shortly about this topic. Let’s keep things as they are for the moment.

PR #1947, opened 2017-04-18
Expose glamor to be able to access it from any extension

gregaubert:


server/sonar-web/src/main/less/init/type.less
```diff
@@ -128,9 +128,13 @@ small,
 
 // Transformation
 
+.text-capitalize   { text-transform: capitalize; }
 .text-uppercase    { text-transform: uppercase; }
 .text-lowercase    { text-transform: lowercase; }
 .text-no-transform { text-transform: none; }
+.text-upperfirstletter:first-letter {
```
stas-vilchik: I suggest `text-upper-first-letter`

PR #1948, opened 2017-04-18
6.4 hardening 2

julienlancelot:


it/it-tests/src/test/java/it/qualityGate/QualityGateTest.java
```diff
@@ -250,6 +251,28 @@ public void ad_hoc_build_break_strategy() throws IOException {
     }
   }
 
+  @Test
+  public void does_not_fail_when_condition_is_on_removed_metric() throws IOException {
```
teryk: you can reproduce with a medium test, why do you add an IT? We have to be careful to not add to many ITs

julienlancelot: It's not that simple as the analysis was failing, that's why I've added an IT

PR #1932, opened 2017-04-13
MMF-468 public private projects

sns-seb:
Shared branch of the atlas team.
:grey_exclamation::busts_in_silhouette::busts_in_silhouette::grey_exclamation:

PR #1958, opened 2017-04-20
MMF-468 SONAR-9096 add private flag to response: api/components/search_projects

bartfastiel:
🌊  This PR is based on #1932 - the shared branch for all private vs. public changes.


bartfastiel:

PR #1960, opened 2017-04-20
MMF-468 SONAR-9099 add private flag to response of api/components/show

bartfastiel:
🌊  This PR is based on #1932 - the shared branch for all private vs. public changes.


server/sonar-server/src/main/java/org/sonar/server/component/ws/ComponentDtoToWsComponent.java
```diff
@@ -57,6 +57,9 @@ private ComponentDtoToWsComponent() {
     setNullable(emptyToNull(dto.language()), wsComponent::setLanguage);
     setTags(dto, wsComponent);
     lastAnalysis.ifPresent(analysis -> wsComponent.setAnalysisDate(formatDateTime(analysis.getCreatedAt())));
+    if (Qualifiers.PROJECT.equals(dto.qualifier())) {
+      wsComponent.setPrivate(dto.isPrivate());
```
sns-seb: should be set for views also
it's currently always false, but views are a root component and have a private flag

bartfastiel:

PR #1949, opened 2017-04-18
SONAR-9132 hide template rules when organizations are enabled

bartfastiel:

PR #1962, opened 2017-04-20
SONAR-9132 remove redundant "org-enabled" check for custom rule create/delete

bartfastiel:
As of #1949, template rules and custom rules will be removed, when you enable organizations. The checks on api/rules/create and api/rules/delete are now redundant.


bellingard:
What's the status of this P/R? @sns-seb approved it 2 days ago, so I would expect it to be merged so that I can test it. Thanks.

PR #1966, opened 2017-04-20
6.4 hardening 5

julienlancelot:


sonar-ws/src/main/protobuf/ws-organizations.proto
```diff
@@ -38,7 +38,8 @@ message UpdateWsResponse {
 
 // WS api/organizations/search
 message SearchWsResponse {
-  repeated Organization organizations = 1;
+  optional sonarqube.ws.commons.Paging paging = 1;
+  repeated Organization organizations = 2;
```
henryju: This is not backward compatible with SQ 6.3... ok I know, the WS is internal, but still we are relying on it in SonarLint. Why not keep organizations filed as first index?

julienlancelot: Indeed good point, I've forgotten that this WS were already used by the Scanner, I will change it

PR #1973, opened 2017-04-24
MMF-468 SONAR-9151 add private flag to api/navigation/component

bartfastiel:


bartfastiel:

PR #1975, opened 2017-04-24
MMF-468 SONAR-9088 add default visibility flag for organizations

bartfastiel:


bartfastiel:

PR #1977, opened 2017-04-24
MMF-468 SONAR-9156 apply orgs default visibility on first analysis

bartfastiel:
🌊  This PR is based on #1975, which should be reviewed first.


bartfastiel:

PR #1982, opened 2017-04-26
Billing plugin integration

gregaubert:


server/sonar-web/src/main/less/components/columns.less
```diff
@@ -32,6 +32,13 @@
   overflow: visible !important;
 }
 
+.column-full {
```
stas-vilchik: I think it's better to not add utility css classes to the sonarqube. It'll be hard to know if they are still used in future.

So please move them to the plugin, and moreover prefix them with `billing-` or something similar to avoid collisions.

server/sonar-web/src/main/js/app/components/extensions/OrganizationPageExtension.js
```diff
@@ -27,6 +27,7 @@ import type { Organization } from '../../../store/organizations/duck';
 
 type Props = {
   organization: Organization,
+  location: { query: { page: string } },
```
stas-vilchik: you can simply write `location: {}`, because not every organization extension will have a page parameter

server/sonar-web/src/main/less/components/columns.less
gregaubert: Ok, you want me to move all of them ?
I did that because I thought those class might be useful in other part than the billing plugin.

stas-vilchik: yes, it's better to move all of them

stas-vilchik:
not related to this PR, but when I'm on the Billing page, I expect Administration menu entry to be selected (with orange underline)

gregaubert:
@stas-vilchik I removed everything related to less exept the one related to the table, to inverse the `zebra` class... Because it seemed complicated to do from two separate places and not worth the hassle.

stas-vilchik:
looks good!

PR #1985, opened 2017-04-26
Fix some issues related to organization and their members

gregaubert:


server/sonar-web/src/main/js/apps/users/components/UsersSelectSearchOption.js
```diff
@@ -64,8 +64,8 @@ export default class UsersSelectSearchOption extends React.PureComponent {
         title={user.name}>
         <div className="little-spacer-bottom little-spacer-top">
           <Avatar hash={user.avatar} email={user.email} size={AVATAR_SIZE} />
-          <strong className="spacer-left">{user.login}</strong>
-          <span className="note little-spacer-left">{this.props.children}</span>
+          <strong className="spacer-left">{this.props.children}</strong>
```
stas-vilchik: same needs to be applied for the current select value

server/sonar-web/src/main/js/apps/account/organizations/CreateOrganizationForm.js
```diff
@@ -229,7 +229,7 @@ class CreateOrganizationForm extends React.PureComponent {
           </div>
 
           <footer className="modal-foot">
-            {this.state.processing
+            {this.state.loading
```
stas-vilchik: usually I disable the submit button and display the spinner, let's keep the consistency

server/sonar-web/src/main/js/apps/users/components/UsersSelectSearchOption.js
gregaubert: Good catch! Thanks!

server/sonar-web/src/main/js/apps/account/organizations/CreateOrganizationForm.js
gregaubert: I think it was done this way to prevent the use to use the cancel button ? Since you can't really cancel the current operation...
So should I display the spinner and disable both buttons ?

stas-vilchik: No, only the submit button. There is nothing really bad if you click "Cancel", because it just closes the modal, and not sends a request.

gregaubert:
Fixed, thanks!

PR #1987, opened 2017-04-27
SONAR-9054 WS api/navigation/organization filter out billing pages for personal organization

teryk:


server/sonar-server/src/test/java/org/sonar/server/ui/ws/OrganizationActionTest.java
```diff
@@ -120,6 +120,25 @@ public void filter_out_admin_pages_when_user_is_not_admin() {
   }
 
   @Test
+  public void filter_out_pages_for_personal_organizations() {
```
julienlancelot: You should add a test checking that billing page is displayed on non perso org

PR #1989, opened 2017-04-27
SQ 6.4 web hardening

gregaubert:
Addressing the following issues : 
* [Blank page when requesting pages starting with 2 slashes](https://jira.sonarsource.com/browse/SONAR-8784) : the problem is [related to a bug](https://github.com/ReactTraining/react-router/issues/3184) in v2 versions of  react-router and history
* [Fix XSS](https://jira.sonarsource.com/browse/SONAR-9003)
* [Gratuitous call to google fonts](https://jira.sonarsource.com/browse/SONAR-9047) : use `webfontloader` to only load the font when the component that needs it is rendered


server/sonar-web/src/main/js/components/SelectList/index.js
```diff
@@ -84,12 +85,13 @@ const SelectListItemView = Backbone.View.extend({
   },
 
   initialize(options) {
+    console.log(`testlog`, options); // eslint-disable-line
```
stas-vilchik: 🙀

server/sonar-web/src/main/js/apps/groups/users-view.js
```diff
@@ -31,8 +32,8 @@ export default Modal.extend({
       width: '100%',
       readOnly: false,
       focusSearch: false,
-      format(item) {
-        return `${item.name}<br><span class="note">${item.login}</span>`;
+      dangerouslyUnescapedHtmlFormat(item) {
```
stas-vilchik: I like the name 👍

server/sonar-web/src/main/js/apps/about/components/AboutAppForSonarQubeDotCom.js
```diff
@@ -45,72 +46,97 @@ type Props = {
...
-              </Link>}
-          </div>
+export default class AboutAppForSonarQubeDotCom extends React.PureComponent {
+  props: Props;
+  state: State;
+
+  constructor(props: Props) {
+    super(props);
+    this.state = { loadingFont: true };
+    WebFont.load({
```
stas-vilchik: this works, however I wonder if we could avoid pulling an extra dependency by using webpack's built-in `require.ensure` function to asynchronously load files

Check any `routes.js` for example.

We could have a wrapper component that will call `require.ensure` to fetch the `AboutAppForSonarQubeDotCom` component in its `componentDidUpdate`, and then render it.

gregaubert: Oh yeah, good idea! I'll check!

gregaubert:
@stas-vilchik I changed the loading of `the AboutAppForSonarQubeDotCom` component!

server/sonar-web/src/main/js/apps/about/components/AboutAppForSonarQubeDotComLazyLoader.js
```diff
@@ -0,0 +1,47 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      AboutAppForSonarQubeDotCom: null
+    };
+  }
+
+  componentDidMount() {
+    if (!this.state.AboutAppForSonarQubeDotCom) {
+      require.ensure(['./AboutAppForSonarQubeDotCom'], require => {
```
stas-vilchik: This is a bit weird api, but you don't have to put anything for the first parameter (`[]` is fine). Webpack will parse the code inside the `require.ensure`, find all calls to `require` and make sure dependencies are available.

PR #1995, opened 2017-04-28
SQ 6.4 web hardening 2

gregaubert:
Addressing the following issues : 
* [Enter key is catched by the button to close the global message popup](https://jira.sonarsource.com/browse/SONAR-8854)
* [Project page extension link is wrong](https://jira.sonarsource.com/browse/SONAR-8975)
* Wrong nav menu highlighted for admin extensions in organizations (related to a comment in #1982]

### DIFF
--- a/server/sonar-web/src/main/js/app/components/nav/component/ComponentNavMenu.js
+++ b/server/sonar-web/src/main/js/app/components/nav/component/ComponentNavMenu.js
@@ -302,7 +302,7 @@ export default class ComponentNavMenu extends React.PureComponent {
     );
   }
 
-  renderExtension = ({ key, name }, isAdmin = false) => {
+  renderExtension = ({ key, name }, isAdmin) => {
     const pathname = isAdmin ? `/project/admin/extension/${key}` : `/project/extension/${key}`;
     return (
       <li key={key}>
@@ -336,7 +336,7 @@ export default class ComponentNavMenu extends React.PureComponent {
           <i className="icon-dropdown" />
         </a>
         <ul className="dropdown-menu">
-          {withoutGovernance.map(this.renderExtension)}
+          {withoutGovernance.map(e => this.renderExtension(e, false))}
         </ul>
       </li>
     );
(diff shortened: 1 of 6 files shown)


server/sonar-web/src/main/js/app/components/nav/component/ComponentNavMenu.js
```diff
@@ -303,7 +303,9 @@ export default class ComponentNavMenu extends React.PureComponent {
   }
 
   renderExtension = ({ key, name }, isAdmin = false) => {
-    const pathname = isAdmin ? `/project/admin/extension/${key}` : `/project/extension/${key}`;
+    const pathname = isAdmin === true
+      ? `/project/admin/extension/${key}`
+      : `/project/extension/${key}`;
```
stas-vilchik: IMO the code is misleading, I'd better change `withoutGovernance.map(this.renderExtension)` to explicitly pass `isAdmin`

server/sonar-web/src/main/js/apps/component-measures/details/history/MeasureHistory.js
```diff
@@ -110,9 +110,10 @@ export default class MeasureHistory extends React.PureComponent {
     }
 
     const data = snapshots.map(snapshot => {
+      const valueAsNumber = Number(snapshot.value);
       return {
         x: snapshot.date,
-        y: Number(snapshot.value)
+        y: isNaN(valueAsNumber) ? snapshot.value : valueAsNumber
```
stas-vilchik: This could potentially hide an error, when a number is badly converted for some reason. Instead I'd make this condition more clear by checking the metric type, something like 
```javascript
this.props.metric.type === 'LEVEL' ? snapshot.value : Number(snapshot.value)
```

server/sonar-web/src/main/js/app/components/nav/component/ComponentNavMenu.js
gregaubert: Yes it is misleading, I was wondering between both solutions. I'll update it.

gregaubert:
It's fixed!

PR #2001, opened 2017-05-01
SQ 6.4 web hardening 3

gregaubert:


server/sonar-web/src/main/js/components/issue/actions.js
```diff
@@ -48,6 +48,7 @@ export const updateIssue = (
       }
     },
     error => {
+      console.log(`testlog`, error); // eslint-disable-line
```
stas-vilchik: if we played a game, you would have the last life 😉

gregaubert: :man_facepalming: I need to find a way to better check this before my commits... Maybe a hook or something...

PR #2003, opened 2017-05-02
SONAR-8865 remove old documents from rules/extension index when updating

bartfastiel:


server/sonar-server/src/main/java/org/sonar/server/rule/index/RuleExtensionDoc.java
```diff
@@ -39,8 +39,7 @@ public RuleExtensionDoc() {
 
   @Override
   public String getId() {
-    // the Id is generated by elasticsearch
-    return null;
+    return getRuleKey() + "|" + getScope().getScope();
```
simonbrandhof: using StringBuilder would be better

it/it-tests/src/test/java/it/rule/RulesPerOrganizationTest.java
```diff
@@ -0,0 +1,103 @@
...
+    createOrganization();
+  }
+
+  @AfterClass
+  public static void tearDown() throws Exception {
+    deleteOrganizationsIfExists(orchestrator, ORGANIZATION_FOO);
+    deleteOrganizationsIfExists(orchestrator, ORGANIZATION_BAR);
+  }
+
+  private static void createOrganization() {
```
bartfastiel: 👍 , but I went the other way around, and made the organization's name a parameter

server/sonar-server/src/main/java/org/sonar/server/rule/index/RuleExtensionDoc.java
bartfastiel: IntelliJ inspections tell me, that this code is not faster as the string concatenation:

```
    return new StringBuilder(getRuleKey().toString())
      .append('|')
      .append(getScope().getScope())
      .toString();

```

So I'll leave everything as it is. Any objections?

PR #2004, opened 2017-05-02
fix loading problem on permission templates page

stas-vilchik:


stas-vilchik:
@gregaubert you'll see my early attempts to have something like redux 😉

PR #2005, opened 2017-05-02
fix extensions using backbone router

stas-vilchik:


gregaubert:
How can I test this ?

stas-vilchik:
You go to Administration -> Configuration -> Portfolios and see an error in the console.

PR #2011, opened 2017-05-03
apply feedback on private projects ui

stas-vilchik:


stas-vilchik:
@gregaubert the first commit here is by mistake, it's already in master

PR #2015, opened 2017-05-03
SONAR-9122 prevent setting a project as private

stas-vilchik:


stas-vilchik:
To be used with real billing plugin or a [fake one].(https://github.com/SonarSource/sonarqube/tree/master/it/it-plugins/fake-billing-plugin)

PR #2020, opened 2017-05-03
SQ 6.4 web hardening 4

gregaubert:


gregaubert:
Oh ok, yeah it makes more sense since these links work if the login is not mandatory... I'll fix it then!

PR #2022, opened 2017-05-04
SONAR-9166 allow to change default project visibility on-premise

stas-vilchik:


stas-vilchik:
it's already done in master :)

PR #2026, opened 2017-05-04
SQ 6.4 web hardening 5

gregaubert:


server/sonar-web/src/main/js/app/components/nav/component/ComponentNav.js
```diff
@@ -69,25 +69,26 @@ export default class ComponentNav extends React.PureComponent {
...
+                favorite={this.props.component.isFavorite}
+              />
 
-            <TooltipsContainer options={{ delay: { show: 0, hide: 2000 } }}>
-              <ComponentNavMeta
-                {...this.props}
-                {...this.state}
-                version={this.props.component.version}
-                snapshotDate={this.props.component.snapshotDate}
+              <ComponentNavBreadcrumbs
```
stas-vilchik: prevent wrapping breadcrumbs into two lines

server/sonar-web/src/main/js/components/shared/drilldown-link.js
```diff
@@ -55,7 +55,9 @@ export class DrilldownLink extends React.PureComponent {
   propsToIssueParams = () => {
     const params = {};
 
-    if (this.props.periodDate) {
+    if (this.props.sinceLeakPeriod) {
```
stas-vilchik: some components are still using the `periodDate` property, for example `overview/main/Coverage`, please change them all

gregaubert: Ok, I thought it was only for quality gates conditions.
Then I'll also update `Coverage` and `Duplications`.
But there is something weird with those components, because they set the `period` property instead of `periodDate`...

gregaubert:
@stas-vilchik It's updated, only the 1st and last commit changed.

server/sonar-web/src/main/js/apps/overview/qualityGate/QualityGateCondition.js
```diff
@@ -124,8 +123,7 @@ export default class QualityGateCondition extends React.PureComponent {
           className={className}
           component={component.key}
           metric={condition.measure.metric.key}
-          period={condition.period}
-          periodDate={periodDate}>
+          sinceLeakPeriod={true}>
```
stas-vilchik: `true` must not be passed always, but only if `condition.period` is defined

For example, add a quality gate condition `issues > 0` and analyze a project with issues. The link will be wrong pointing only to the leak period.

gregaubert: Oh yeah, good catch! :+1: Thanks!

PR #2027, opened 2017-05-04
SONAR-9077 let api/components/suggestions return name instead longName

bartfastiel:


bartfastiel:
Will be covered by #2039

PR #2029, opened 2017-05-05
SQ 6.4 web hardening 6

gregaubert:


server/sonar-web/src/main/js/apps/issues/components/AppContainer.js
```diff
@@ -36,13 +39,28 @@ const mapStateToProps = (state, ownProps) => ({
...
-    return { ...response, issues: parsedIssues };
-  }, onFail(dispatch));
+  searchIssues({ ...query, additionalFields: '_all' })
+    .then(response => {
+      const parsedIssues = response.issues.map(issue =>
+        parseIssueFromResponse(issue, response.components, response.users, response.rules)
+      );
+      return { ...response, issues: parsedIssues };
+    })
+    .then(response => fetchIssueOrganizations(response.issues)(dispatch).then(() => response))
```
stas-vilchik: would be better:
```javascript
dispatch(fetchIssueOrganizations(response.issues))
```

server/sonar-web/src/main/js/apps/issues/sidebar/ProjectFacet.js
```diff
@@ -28,7 +28,8 @@ import FacetFooter from './components/FacetFooter';
 import type { ReferencedComponent, Component } from '../utils';
 import Organization from '../../../components/shared/Organization';
 import QualifierIcon from '../../../components/shared/QualifierIcon';
-import { searchComponents, getTree } from '../../../api/components';
+import { convertToFilter } from '../../../apps/projects/store/utils';
```
stas-vilchik: I'd not introduce a dependency between `issues` and `projects` apps for such small detail

server/sonar-web/src/main/js/components/SourceViewer/SourceViewerBase.js
```diff
@@ -248,6 +250,8 @@ export default class SourceViewerBase extends React.PureComponent {
       if (this.mounted) {
         if (response.status === 403) {
           this.setState({ component, loading: false, notAccessible: true });
+        } else if (response.status === 404) {
+          this.setState({ component, loading: false, sourceRemoved: true });
```
stas-vilchik: `sourceRemoved` should be reset to `false` when sources exist

otherwise when you're switching between issues using arrow key, if you open a removed file, then you'll see the warning for every other file

server/sonar-web/src/main/js/apps/issues/components/AppContainer.js
gregaubert: Ok I can do that, but then I need to remove the typings for `dispatch` or update the `Dispatch` definition to make it work. Because flow complains that `.then` doesn't exists on `dispatch`.

server/sonar-web/src/main/js/apps/issues/sidebar/ProjectFacet.js
gregaubert: :+1: true

PR #2031, opened 2017-05-05
SONAR-9168 do not fail when indexing measure with unexpected text value

simonbrandhof:
Contrary to what's suggested in ticket, this PR simply ignores numeric measures that don't have any numeric value


server/sonar-db-dao/src/main/java/org/sonar/db/measure/ProjectMeasuresIndexerIterator.java
```diff
@@ -220,9 +220,7 @@ private void readMeasure(ResultSet rs, Measures measures) throws SQLException {
     }
     if (NCLOC_LANGUAGE_DISTRIBUTION_KEY.equals(metricKey)) {
       readTextValue(rs, measures::setLanguages);
-      return;
```
teryk: I'd keep the return; for consistency or replace with a switch

PR #2041, opened 2017-05-09
SONAR-9114 allow to search in “_all_” orgs in api/components/search

bartfastiel:


bartfastiel:
Simon and me discussed, that we do not want a magic value, but an additional boolean parameter instead.

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchAction.java
```diff
@@ -95,6 +101,13 @@ public void define(WebService.NewController context) {
       .setExampleValue("my-org")
       .setSince("6.3");
     action
+      .createParam(PARAM_ALL_ORGANIZATIONS)
+      .setDescription("Whether to return components of all organizations. Must not be set, if '" + PARAM_ORGANIZATION + "' is set.")
```
simonbrandhof: The incompatibility with parameter "organization" is not enforced. Not a problem so far, but in this case it deserves to be documented.

simonbrandhof: My bad, validation is done by SearchWsRequest validate(). I didn't see because we try to not do any validation of request in WS client.

bartfastiel:
The bug has been solved by switching to another webservice. So *for the moment* there is no need to change api/components/search .

PR #2047, opened 2017-05-10
SONAR-9174 Display issue box on the last line of its primary location

stas-vilchik:


gregaubert:
:cry: I can't find an issue on multiple lines to test it!

stas-vilchik:
try
```javascript
alert('foo' +
  'bar');
```

PR #2057, opened 2017-05-11
SONAR-9073 fix response example of api/components/suggestions

bartfastiel:


server/sonar-server/src/main/resources/org/sonar/server/component/ws/components-example-suggestions.json
```diff
@@ -31,5 +56,6 @@
       "name": "Default Organization"
     }
   ],
-  "warning": "short_input"
+  "projects": [
```
simonbrandhof: there should be a project, correct ?

bartfastiel: In the current example, there should not be any projects. This node is only containing project key/name mappings for referenced projects of modules, files and test files.

Maybe a more sophisticated example with suggestions in several qualifiers would be easier to understand, but for this internal api I would like to keep things simple for the moment.

simonbrandhof: ok

PR #2059, opened 2017-05-11
SONAR-9217 if two component suggestions have the same relevancy, prefer A over Z

bartfastiel:
api/components/suggestions used to only use "relevance" (score) in sorting results. If two suggestions have exactly the same relevance (this is a rare case), then we will now also take alphabetical order into account.


server/sonar-server/src/test/java/org/sonar/server/component/index/ComponentIndexScoreTest.java
```diff
@@ -146,6 +146,13 @@ public void do_not_match_wrong_file_extension() {
   }
 
   @Test
+  public void rank_a_higher_then_b() {
```
simonbrandhof: this test should be renamed so that we understand that ordering fallbacks on name field is score is the same between two results.

PR #2061, opened 2017-05-11
SONAR-9072 for api/components/suggestions use all valid search tokens

bartfastiel:
The search for suggestions should ignore tokens, that are only one character long (with warning), but use all other tokens for the search.

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/component/ws/SuggestionsAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/component/ws/SuggestionsAction.java
@@ -22,6 +22,7 @@ package org.sonar.server.component.ws;
 import com.google.common.collect.ListMultimap;
 import com.google.common.html.HtmlEscapers;
 import com.google.common.io.Resources;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashSet;
@@ -30,6 +31,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import org.sonar.api.server.ws.Change;
 import org.sonar.api.server.ws.Request;
@@ -46,7 +48,7 @@ import org.sonar.server.component.index.ComponentHitsPerQualifier;
 import org.sonar.server.component.index.ComponentIndex;
 import org.sonar.server.component.index.ComponentIndexQuery;
 import org.sonar.server.component.index.ComponentIndexResults;
-import org.sonar.server.es.textsearch.ComponentTextSearchFeatureRepertoire;
+import org.sonar.server.es.DefaultIndexSettings;
 import org.sonar.server.favorite.FavoriteFinder;
 import org.sonar.server.user.UserSession;
 import org.sonarqube.ws.WsComponents.SuggestionsWsResponse;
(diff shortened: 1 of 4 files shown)


server/sonar-server/src/main/java/org/sonar/server/component/ws/SuggestionsAction.java
```diff
@@ -200,6 +202,12 @@ private SuggestionsWsResponse loadSuggestionsWithoutSearch(int skip, int limit,
   }
 
   private SuggestionsWsResponse loadSuggestionsWithSearch(String query, int skip, int limit, Set<String> recentlyBrowsedKeys, List<String> qualifiers) {
+    if (split(query).noneMatch(token -> token.length() >= MINIMUM_NGRAM_LENGTH)) {
+      SuggestionsWsResponse.Builder queryBuilder = newBuilder();
+      getWarning(query).ifPresent(queryBuilder::setWarning);
```
simonbrandhof: Having a single call to `getWarning()` in this method would be better but ok, it's acceptable.

server/sonar-server/src/main/java/org/sonar/server/es/textsearch/ComponentTextSearchFeatureRepertoire.java
```diff
@@ -53,35 +57,54 @@ public QueryBuilder getQuery(ComponentTextSearchQuery query) {
   },
   PREFIX(CHANGE_ORDER_OF_RESULTS) {
     @Override
-    public QueryBuilder getQuery(ComponentTextSearchQuery query) {
-      return prefixAndPartialQuery(query.getQueryText(), query.getFieldName(), SEARCH_PREFIX_ANALYZER)
+    public Stream<QueryBuilder> getQueries(ComponentTextSearchQuery query) {
+      List<String> tokens = split(query.getQueryText());
```
simonbrandhof: The split is executed multiple times, once per search criteria. It should be part of `ComponentTextSearchQuery`.

server/sonar-server/src/main/java/org/sonar/server/es/textsearch/ComponentTextSearchFeatureRepertoire.java
```diff
@@ -53,35 +57,54 @@ public QueryBuilder getQuery(ComponentTextSearchQuery query) {
...
     }
   },
   KEY(GENERATE_RESULTS) {
     @Override
-    public QueryBuilder getQuery(ComponentTextSearchQuery query) {
-      return matchQuery(SORTABLE_ANALYZER.subField(query.getFieldKey()), query.getQueryText())
-        .boost(50f);
+    public Stream<QueryBuilder> getQueries(ComponentTextSearchQuery query) {
+      String queryText = query.getQueryText();
+      if (ComponentKeys.isValidModuleKey(queryText)) {
```
simonbrandhof: Is it expected to not search for file keys anymore ? Here only projects and modules are supported.

server/sonar-server/src/main/java/org/sonar/server/component/ws/SuggestionsAction.java
bartfastiel: Not trivial to refactor, so I'll leave this for the moment.

server/sonar-server/src/main/java/org/sonar/server/es/textsearch/ComponentTextSearchFeatureRepertoire.java
bartfastiel: I've dropped this change.

server/sonar-server/src/main/java/org/sonar/server/es/textsearch/ComponentTextSearchFeatureRepertoire.java
bartfastiel: Done.

PR #2062, opened 2017-05-11
SONAR-9219 Deprecate parameters "developer{Id|Key}" in web services api/measures/component and api/measures/component_tree

simonbrandhof:


server/sonar-server/src/main/java/org/sonar/server/measure/ws/MeasuresWsParametersBuilder.java
```diff
@@ -51,12 +49,13 @@ static NewParam createMetricKeysParameter(NewAction action) {
...
-      .setDescription("Developer id. If set, developer's measures are returned. Only with Developer Cockpit plugin.")
-      .setExampleValue(Uuids.UUID_EXAMPLE_01);
-    action.createParam(PARAM_DEVELOPER_KEY)
-      .setDescription("Developer key. If set, developer's measures are returned. Only with Developer Cockpit plugin.")
-      .setExampleValue(KeyExamples.KEY_DEVELOPER_EXAMPLE_001);
+    deprecateDeveloperParameter(action, PARAM_DEVELOPER_ID);
+    deprecateDeveloperParameter(action, PARAM_DEVELOPER_KEY);
   }
 
+  private static void deprecateDeveloperParameter(NewAction action, String key) {
```
bartfastiel: One day, we have to change `WebService`, so that we do not have to modify mutable parameters (like `action` in this case) anymore.

PR #2065, opened 2017-05-12
integration tests: avoid false positive, if tests are executed very fast

bartfastiel:


bartfastiel:
These false positives did not occur for some time. No action required at the moment.

PR #2070, opened 2017-05-12
SONAR-9197 add index to active_rule_parameters

simonbrandhof:
in order to support big volume when organizations are enabled


server/sonar-db-core/src/main/resources/org/sonar/db/version/schema-h2.ddl
```diff
@@ -370,6 +370,7 @@ CREATE TABLE "ACTIVE_RULE_PARAMETERS" (
   "RULES_PARAMETER_KEY" VARCHAR(128),
   "VALUE" VARCHAR(4000)
 );
+CREATE INDEX "ix_arp_on_active_rule_id" ON "ACTIVE_RULE_PARAMETERS" ("ACTIVE_RULE_ID");
```
sns-seb: uppercase is used everywhere else in the file

PR #2078, opened 2017-05-18
Cluster ITs

ehartmann:


it/it-tests/pom.xml
```diff
@@ -39,6 +39,12 @@
     </dependency>
     <dependency>
       <groupId>org.sonarsource.sonarqube</groupId>
+      <artifactId>sonar-process</artifactId>
```
simonbrandhof: Integration tests should not call SQ code.

it/it-tests/src/test/java/it/cluster/Cluster.java
```diff
@@ -0,0 +1,243 @@
...
+import static org.sonar.process.ProcessProperties.CLUSTER_NETWORK_INTERFACES;
+import static org.sonar.process.ProcessProperties.CLUSTER_PORT;
+import static org.sonar.process.ProcessProperties.CLUSTER_SEARCH_DISABLED;
+import static org.sonar.process.ProcessProperties.CLUSTER_SEARCH_HOSTS;
+import static org.sonar.process.ProcessProperties.CLUSTER_WEB_DISABLED;
+import static org.sonar.process.ProcessProperties.SEARCH_HOST;
+import static org.sonar.process.ProcessProperties.SEARCH_JAVA_OPTS;
+import static org.sonar.process.ProcessProperties.SEARCH_PORT;
+import static org.sonar.process.ProcessProperties.WEB_JAVA_OPTS;
+import static org.sonar.process.ProcessProperties.WEB_PORT;
```
simonbrandhof: See previous comment, all these import should be removed.

pom.xml
```diff
@@ -69,7 +69,7 @@
     <slf4j.version>1.7.21</slf4j.version>
     <tomcat.version>8.5.11</tomcat.version>
     <elasticsearch.version>2.4.4</elasticsearch.version>
-    <orchestrator.version>3.15.0.1038</orchestrator.version>
+    <orchestrator.version>3.15-SNAPSHOT</orchestrator.version>
```
simonbrandhof: SNAPSHOT versions are not allowed.

it/it-tests/src/test/java/it/cluster/NetworkUtils.java
```diff
@@ -0,0 +1,58 @@
...
+import java.net.NetworkInterface;
+import java.net.SocketException;
+import java.util.Collections;
+import java.util.Enumeration;
+
+public class NetworkUtils {
+  // TODO : move it to orchestrator ?
+  // Same usage for ClusterSettingsLoopbackTest
+
+  public static InetAddress getNonloopbackIPv4Address()  {
```
simonbrandhof: Instead of introducing a new class `NetworkUtils`, this method should be moved to the class `Cluster`.

pom.xml
ehartmann: Yes I've just set this to be able to use the MsSQL automatic collation. I'll update this.

it/it-tests/pom.xml
ehartmann: I'll remove this, I've seen some usage of direct calls, so I've used it.

ehartmann:
This test must use category 5.

ehartmann:
This PR has been picked in https://github.com/SonarSource/sonarqube/pull/2367

PR #2081, opened 2017-05-18
SONAR-9286 Fix ES error at startup

ehartmann:


ehartmann:
Will be fixed directly in branch-6.4

PR #2082, opened 2017-05-19
SONAR-8596 Fix a bug with cached bundles

gregaubert:
* There is a bug happening when the bundle is already cached, the 304 response is not handled correctly.
* And a second problem : if the navigator is set as `en-US` we don't detect that it's the same language as `en` that is returned by the WS as the effective language. And therefore we don't use the caching capabilities.

### DIFF
--- a/server/sonar-web/src/main/js/helpers/l10n.js
+++ b/server/sonar-web/src/main/js/helpers/l10n.js
@@ -19,7 +19,7 @@
  */
 /* @flow */
 import moment from 'moment';
-import { request } from './request';
+import { getJSON } from './request';
 
 let messages = {};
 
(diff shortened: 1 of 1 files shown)


server/sonar-web/src/main/js/helpers/l10n.js
```diff
@@ -92,34 +67,49 @@ function checkCachedBundle() {
...
 }
 
+function getL10nBundle(params) {
+  const url = '/api/l10n/index';
+  return getJSON(url, params);
+}
+
 export function requestMessages() {
-  const currentLocale = getPreferredLanguage();
+  const browserLocal = getPreferredLanguage();
```
stas-vilchik: browserLocale

PR #2084, opened 2017-05-19
SONAR-9253 Create an option panel on top of projects page

gregaubert:


server/sonar-web/src/main/js/apps/projects/components/AllProjects.js
```diff
@@ -28,112 +30,116 @@ import { parseUrlQuery } from '../store/utils';
...
-    const { query } = this.state;
+    const { query, optionBarOpen } = this.state;
     const isFiltered = Object.keys(query).some(key => query[key] != null);
 
-    const view = query.view || 'list';
+    const view = query.view || 'overall';
     const visualization = query.visualization || 'risk';
 
-    const top = this.props.organization ? 95 : 30;
+    const top = (this.props.organization ? 95 : 30) + (optionBarOpen ? 45 : 0);
```
stas-vilchik: Once the page is scrolled, the sidebar looks weird, it doesn't stick to the top of the page any more. The perfect solution would be to make sidebar `position: fixed` only when is scrolled more than 45px.

sonar-core/src/main/resources/org/sonar/l10n/core.properties
```diff
@@ -862,8 +862,9 @@ projects.explore_projects=Explore Projects
 projects.not_analyzed=Project is not analyzed yet.
 projects.search=Search by project name or key
 projects.sort_list=Sort list by
-projects.view.list=List
-projects.view.visualizations=Visualizations
+projects.perspective=Perspective
+projects.view_settings=Settings
+projects.view.overall=Overall status
```
stas-vilchik: `Overall Status` (should be capitalized)

server/sonar-web/src/main/js/apps/projects/components/PerspectiveSelectOption.js
```diff
@@ -0,0 +1,71 @@
...
+    const { option } = this.props;
+    return (
+      <div
+        className={this.props.className}
+        onMouseDown={this.handleMouseDown}
+        onMouseEnter={this.handleMouseEnter}
+        onMouseMove={this.handleMouseMove}
+        title={option.label}>
+        <div>
+          {option.type === 'view' && <i className="icon-list little-spacer-right" />}
```
stas-vilchik: make this icon a react component

server/sonar-web/src/main/js/apps/projects/components/ProjectOptionBar.js
```diff
@@ -0,0 +1,74 @@
...
+    evt.preventDefault();
+    this.props.onToggleOptionBar(false);
+  };
+
+  render() {
+    const { open } = this.props;
+    return (
+      <div className="projects-topbar">
+        <a className="button projects-topbar-open" href="#" onClick={this.openBar}>
+          {translate('projects.view_settings')}
```
stas-vilchik: On mockups there is an icon here. Did you decide to not use it?

server/sonar-web/src/main/js/apps/projects/components/AllProjects.js
gregaubert: Ah yes, good catch! Or maybe the top bar should be fixed ? And not hidden when scrolling ?

server/sonar-web/src/main/js/apps/projects/components/ProjectOptionBar.js
gregaubert: Yeah after discussion with Laura we decided to remove it. Mainly because there is nearly nowhere icons in buttons in the application.

server/sonar-web/src/main/js/apps/projects/components/PerspectiveSelectOption.js
gregaubert: Ok, I can, but what's the reason for preferring having them as components ? I also saw that the sonar font was created from an online tool using icons from other fonts, why not continue like this ?

I think I'll move all the icon components in an icon folder if you are ok with it ?

gregaubert:
@stas-vilchik I applied the change, you can check only the two last commits.

server/sonar-web/src/main/js/components/icons-components/CloseIcon.js
```diff
@@ -0,0 +1,40 @@
...
+
+export default function CloseIcon({ className, size = 16 }: Props) {
+  /* eslint-disable max-len */
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+      height={size}
+      width={size}
+      viewBox="0 0 16 20">
```
stas-vilchik: the ratio of width/height is not the same as ratio of the viewBox: 16px/16px to 16px/20px, this might cause problems

server/sonar-web/src/main/js/components/icons-components/ListIcon.js
```diff
@@ -0,0 +1,40 @@
...
+
+export default function ListIcon({ className, size = 16 }: Props) {
+  /* eslint-disable max-len */
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+      height={size}
+      width={size}
+      viewBox="0 0 16 16">
```
stas-vilchik: it's good here

server/sonar-web/src/main/js/components/icons-components/CloseIcon.js
gregaubert: Ok, so for the height I'll multiply the size given in props by the ratio

PR #2092, opened 2017-05-23
SONAR-8913 Not analyzed projects on projects page

julienlancelot:
Replace "snapshotDate" by "analysisDate" in api/navigation/component


gregaubert:
@stas-vilchik You can review only my commit. It's fixing an issue where we had the "Not analyzed yet" message in the project card when switching from the projects page to a project detail and back to the projects page.

PR #2096, opened 2017-05-23
SONAR-9259 allow to sort by last analysis date in project search

bartfastiel:


server/sonar-server/src/main/java/org/sonar/server/component/ws/ProjectMeasuresQueryValidator.java
```diff
@@ -29,13 +30,17 @@
...
+import static java.util.Arrays.asList;
 import static org.sonar.core.util.stream.MoreCollectors.toHashSet;
 import static org.sonar.core.util.stream.MoreCollectors.toSet;
 import static org.sonar.server.measure.index.ProjectMeasuresQuery.MetricCriterion;
+import static org.sonar.server.measure.index.ProjectMeasuresQuery.SORT_BY_LAST_ANALYSIS_DATE;
 import static org.sonar.server.measure.index.ProjectMeasuresQuery.SORT_BY_NAME;
 
 public class ProjectMeasuresQueryValidator {
 
+  private static final Set<String> NON_METRIC_SORT_KEYS = new HashSet<>(asList(SORT_BY_NAME, SORT_BY_LAST_ANALYSIS_DATE));
```
julienlancelot: I think it would be better to use an ImmutableSet

server/sonar-server/src/test/java/org/sonar/server/component/ws/SearchProjectsActionTest.java
```diff
@@ -918,6 +921,28 @@ public void sort_by_quality_gate_then_by_name() throws Exception {
...
+    ComponentDto project3 = db.components().insertPublicProject(organization);
+    db.components().insertSnapshot(project3, snapshot -> snapshot.setCreatedAt(20_000_000_000L).setLast(true));
+    authorizationIndexerTester.allowOnlyAnyone(project3);
+    ComponentDto project4 = db.components().insertPublicProject(organization);
+    db.components().insertSnapshot(project4, snapshot -> snapshot.setCreatedAt(10_000_000_000L).setLast(false));
+    db.components().insertSnapshot(project4, snapshot -> snapshot.setCreatedAt(30_000_000_000L).setLast(true));
+    authorizationIndexerTester.allowOnlyAnyone(project4);
+    projectMeasuresIndexer.indexOnStartup(null);
+
+    assertThat(call(request.setSort(ANALYSIS_DATE).setAsc(true)).getComponentsList()).extracting(Component::getKey)
```
julienlancelot: I think we should also test by descending sort (by only duplicating this line with setAsc(false) and change expected project keys)

server/sonar-server/src/main/java/org/sonar/server/component/ws/ProjectMeasuresQueryValidator.java
bartfastiel: I prefer to avoid guava, as it is an external reference.

server/sonar-server/src/test/java/org/sonar/server/component/ws/SearchProjectsActionTest.java
bartfastiel: 👍

PR #2099, opened 2017-05-23
SONAR-9295 Add the default port for ES

ehartmann:


server/sonar-server/src/main/java/org/sonar/server/es/EsClientProvider.java
```diff
@@ -56,13 +56,14 @@ public EsClient provide(Settings settings) {
         esSettings.put("client.transport.sniff", true);
         nativeClient = TransportClient.builder().settings(esSettings).build();
         Arrays.stream(settings.getStringArray(ProcessProperties.CLUSTER_SEARCH_HOSTS))
-          .map(Host::parse)
-          .forEach(h -> h.addTo(nativeClient));
+          .map(HostAndPort::fromString)
```
ehartmann: @sns-seb I'll add ITs for this. The main issue was that EsClientProvider *in cluster mode* required to have `sonar.cluster.search.hosts=192.168.0.10:9001,192.168.0.11:9001` and `sonar.cluster.search.hosts=192.168.0.10,192.168.0.11` was not allowed making configuration of cluster more harder.
The main modification is here with HostAndPort::fromString function that will automatically add the default port.

When SonarQube is installed standalone there is no need for such change.

sns-seb: ok, I understand. I missed that the target of the change was really `sonar.cluster.search.hosts` when cluster is enabled and the current node has ES.

But I think I still have  a point.

Only it applies to when clustering is not enabled and search is configured with properties `ProcessProperties.SEARCH_HOST` and `ProcessProperties.SEARCH_PORT`. Also, it's not a regression nor a bug, but rather an **inconsistency**: 
* when clustering is enabled, "no port" means "use the default port" while
* when clustering is not enabled, "no port" means "bind to any port" (car port 0 => attribution automatique)

This should be discussed with @christophelevis

ehartmann: I don't think it's inconsistency there but this reflects the two modes to connect to ElasticSearch in a cluster environment:

- If Elasticsearch is activated on the node, SonarQube is not trying to connect to Elasticsearch as a cluster but connect to the local Elasticsearch node (so using `SEARCH_HOST` ie. the interface bound and `SEARCH_PORT`, and I think 0 should never be used in production mode)
- If Elasticsearch is not activated on the node, SonarQube is using `CLUSTER_SEARCH_HOSTS` that contains the list of Elasticsearch nodes (with or without port number). And in this case omitting port number should not lead to an exception but using the default Elasticsearch port.

PR #2097, opened 2017-05-23
SONAR-9245 Return leak period in api/components/search_projects

julienlancelot:

PR #2100, opened 2017-05-23
SONAR-9245 Create projects leak view

gregaubert:
[SONAR-9257](https://jira.sonarsource.com/browse/SONAR-9257) : 
* [X] Add a "Display" option to the new top panel to switch between Overall and Leak view
* [X] Create/Update Project cards for the leak
* [x] Add leak period start date on the right with the analysis date
* [X] Add facets from the Overall view :
  * Add Quality Gate facet
  * Add Languages facet
  * Add Tags facet

[SONAR-9256](https://jira.sonarsource.com/browse/SONAR-9256)  :
* [X] Update design of project cards

~:warning: The leak period start date is not yet there, I'm waiting on #2097 to be merged before adding it~


gregaubert:
@stas-vilchik You can already review it right now or wait on the leak period start date before reviewing it.
In any case I will add the leak period start date as a separate commit.

server/sonar-web/src/main/js/apps/projects/components/ProjectCardOverallMeasures.js
```diff
@@ -111,11 +108,11 @@ export default function ProjectCardMeasures({ measures }: Props) {
       </div>
 
       {measures['ncloc'] != null &&
-        <div className="project-card-measure" data-key="ncloc">
+        <div className="project-card-measure pull-right" data-key="ncloc">
           <div className="project-card-measure-inner">
             <div className="project-card-measure-number">
               <span className="spacer-right">
-                <SizeRating value={formatMeasure(measures['ncloc'], 'INT')} />
+                <SizeRating value={parseInt(measures['ncloc'])} />
```
stas-vilchik: should be either `parseInt(measures['ncloc'], 10)` or `Number(measures['ncloc'])`

server/sonar-web/src/main/js/apps/projects/styles.css
```diff
@@ -136,10 +137,18 @@
   content: "";
 }
 
+.project-card-measure.pull-right {
```
stas-vilchik: instead of doing this you could add both `pull-right` and `text-right` classes to the element

server/sonar-web/src/main/js/apps/projects/styles.css
```diff
@@ -136,10 +137,18 @@
...
+  text-align: right;
+}
+
 .project-card-measure.pull-right:before {
   display: none;
 }
 
+.project-card-measure.pull-right .project-card-measure-inner {
+  text-align: right;
```
stas-vilchik: same here

server/sonar-web/src/main/js/apps/projects/styles.css
gregaubert: In fact I didn't use .`text-right` because it's overwritten by the `text-align: center` of `.project-card-measure`.
So either I let it like right now, or I add a `!important` to `.text-right`, but I'm sure it's a good idea.

gregaubert:
@stas-vilchik I added the leak period start date, you can review only the last commit.

server/sonar-web/src/main/js/apps/projects/components/ProjectCard.js
```diff
@@ -99,6 +89,31 @@ export default function ProjectCard({ measures, organization, project, type }: P
...
+        </div>
+        {isLeakView &&
+          isProjectAnalyzed &&
+          <div
+            className={classNames('project-card-dates note text-right pull-right', {
+              'width-100': isPrivate || hasTags
+            })}>
+            {hasLeakPeriodStart &&
+              <span>
+                {translateWithParameters(
```
stas-vilchik: there is a little inconsistency between two messages: 

> Leak Period started 7 months ago

> Last analysis: April 26, 2017 11:38 AM

Note that the second one has a colon, but the first one doesn't. This is a bit annoying.

server/sonar-web/src/main/js/apps/projects/components/ProjectCard.js
```diff
@@ -99,6 +89,31 @@ export default function ProjectCard({ measures, organization, project, type }: P
...
+                  'projects.leak_period_x',
+                  moment(project.leakPeriodDate).fromNow()
+                )}
+              </span>}
+            {isProjectAnalyzed &&
```
stas-vilchik: this block is included in `{isLeakView &&`, so you don't see the analysis date in the "Overall" mode

gregaubert: Yes, this is kind of "wanted", check this : https://jira.sonarsource.com/browse/SONAR-9256

PR #2101, opened 2017-05-23
SONAR-9296 Do not check sonar.search.host if search is disabled

ehartmann:


server/sonar-process-monitor/src/main/java/org/sonar/application/config/ClusterSettings.java
```diff
@@ -65,8 +65,15 @@ private static void checkProperties(Props props) {
       throw new MessageException(format("Property [%s] is forbidden", CLUSTER_WEB_LEADER));
     }
 
+    if (props.valueAsBoolean(ProcessProperties.CLUSTER_ENABLED) &&
+      !props.valueAsBoolean(ProcessProperties.CLUSTER_SEARCH_DISABLED, false)
+      ) {
+      // If cluster enabled and search not disabled
```
ehartmann: Indeed, if it's clear for everyone I can remove it. 
I've just written it because it's a double negation : "not disabled".

PR #2106, opened 2017-05-26
Fix unconsistant IT's in organization membership bis

gregaubert:


it/it-tests/src/test/java/it/organization/OrganizationMembershipTest.java
```diff
@@ -152,57 +151,58 @@ public void user_creating_an_organization_becomes_member_of_this_organization()
...
   }
 
   @Test
   public void search_for_members() {
     String orgKey = createOrganization();
     String user1 = createUser();
-    String user2 = createUser();
-    createUser();
+    String user2 = createUser("sameprefixuser");
+    createUser("sameprefixuser");
```
bartfastiel: 🐜  You might consider rearranging the order of the lines, to make it more obvious, that one of the users is not made a member.

✨  Alternative: Create a method `createMember`, that creates a user + makes it a member.

it/it-tests/src/test/java/it/organization/OrganizationMembershipTest.java
```diff
@@ -152,57 +151,58 @@ public void user_creating_an_organization_becomes_member_of_this_organization()
...
       .shouldNotHaveActions();
   }
 
   @Test
   public void search_for_members() {
     String orgKey = createOrganization();
     String user1 = createUser();
-    String user2 = createUser();
-    createUser();
+    String user2 = createUser("sameprefixuser");
```
bartfastiel: 🐞  If you want to increase the scope of this test, you could also use `util.user.UserRule#createUser(java.lang.String, java.lang.String, java.lang.String, java.lang.String)` to provide a name, different from the user's login.

it/it-tests/src/test/java/it/organization/OrganizationMembershipTest.java
```diff
@@ -152,57 +151,58 @@ public void user_creating_an_organization_becomes_member_of_this_organization()
   @Test
   public void should_display_members_page() {
     String orgKey = createOrganization();
-    userRule.createUser("foo", "pwd");
-    userRule.createUser("bar", "pwd");
+    String userFoo = createUser("foo");
+    String userBar = createUser("bar");
```
bartfastiel: 🐜  You might consider rearranging the order of the lines, to make it more obvious, that one of the users is not made a member.

it/it-tests/src/test/java/it/organization/OrganizationMembershipTest.java
```diff
@@ -84,7 +84,6 @@ public void tearDown() throws Exception {
   @Test
   public void new_user_should_not_become_member_of_default_organization() throws Exception {
     String login = createUser();
-
```
bartfastiel: ❓  Out of curiosity: What does `bis` (in the commit message) mean?

gregaubert: It's because I already tried to fix this test in a previous commit with the same message, but I missed some cases and we still had some FP.

PR #2107, opened 2017-05-26
Display About page faster

gregaubert:
Related to https://groups.google.com/a/sonarsource.com/d/topic/dogfood-sonarqube/KST_kILd_Uc/discussion


gregaubert:
:+1: It does look better with only one spinner each.
I added the global loading spinner in the index page too. I had to move the style to a less file to have it available there.

There is an "issue" though, the global loading blink when the js files are loaded.

gregaubert:
Ok, with this the blink is a little less noticeable but we can still see it

gregaubert:
@stas-vilchik Do you think there is something else we could do to remove or reduce the blink ?

stas-vilchik:
@gregaubert I think it's fine now.

gregaubert:
Ok I'll merge then

PR #2123, opened 2017-05-29
rework quality profiles modals

stas-vilchik:
replace backbone modals with react


server/sonar-web/src/main/js/apps/quality-profiles/components/ProfileActions.js
```diff
@@ -152,6 +182,30 @@ export default class ProfileActions extends React.PureComponent {
...
+          <DeleteProfileForm
+            onClose={this.closeDeleteForm}
+            onDelete={this.handleProfileDelete}
+            onRequestFail={this.props.onRequestFail}
+            profile={profile}
+          />}
+
+        {this.state.renameFormOpen &&
+          <RenameProfileForm
+            onClose={this.closeDeleteForm}
```
gregaubert: Copy-paste error probably, currently the rename form can't close when clicking on cancel

server/sonar-web/src/main/js/apps/quality-profiles/components/RenameProfileForm.js
```diff
@@ -0,0 +1,132 @@
...
+                required={true}
+                size="50"
+                type="text"
+                value={this.state.name != null ? this.state.name : profile.name}
+              />
+            </div>
+          </div>
+          <div className="modal-foot">
+            {this.state.loading && <i className="spinner spacer-right" />}
+            <button disabled={this.state.loading || !this.state.name} id="rename-profile-submit">
```
gregaubert: Like for the `copy form`, I think it would be good to not allow submit if the name didn't change

server/sonar-web/src/main/js/apps/quality-profiles/details/ChangeParentForm.js
```diff
@@ -0,0 +1,138 @@
...
+                    ? this.state.selected
+                    : this.props.profile.parentKey || ''
+                }
+              />
+            </div>
+          </div>
+          <div className="modal-foot">
+            {this.state.loading && <i className="spinner spacer-right" />}
+            <button
+              disabled={this.state.loading || this.state.selected == null}
```
gregaubert: Same as rename and copy, we can submit when selecting the same parent as before.
In the end I don't know if we should really prevent the user from submitting, maybe we should just not do the api call.

server/sonar-web/src/main/js/apps/quality-profiles/details/ChangeProjectsForm.js
```diff
@@ -64,10 +72,32 @@ export default ModalFormView.extend({
...
+        <div className="modal-head">
+          <h2>{header}</h2>
+        </div>
+
+        <div className="modal-body">
+          <div id="profile-projects" ref={node => (this.container = node)} />
+        </div>
+
+        <div className="modal-foot">
+          <a href="#" onClick={this.handleCloseClick}>{translate('close')}</a>
```
gregaubert: The other modals have the "cancel" text instead of "close"

stas-vilchik: Yes, because all other modals have two options: submit or cancel. Here there is no explicit submit action, when you select/deselect a project, it is submitted automatically.

PR #2126, opened 2017-05-30
MMF-324 Update UI for built-in quality profiles

stas-vilchik:


stas-vilchik:
very good point @gregaubert ! 👏

PR #2131, opened 2017-05-31
SONAR-9328 optimized delete settings and manual measures when deleting a project

sns-seb:
based on #2127 

includes SONAR-9351

PR #2132, opened 2017-05-31
SONAR-9350 delete components by project when deleting a project

sns-seb:
works with views also, of course

based on #2131


server/sonar-db-dao/src/main/resources/org/sonar/db/purge/PurgeMapper.xml
```diff
@@ -185,10 +181,8 @@
   <delete id="deleteComponents" parameterType="map">
     delete from projects
     where
-      uuid in
-      <foreach collection="componentUuids" open="(" close=")" item="componentUuid" separator=",">
-        #{componentUuid,jdbcType=VARCHAR}
-      </foreach>
+      project_uuid = #{rootUuid,jdbcType=VARCHAR}
+      or uuid = #{rootUuid,jdbcType=VARCHAR}
```
simonbrandhof: "uuid" is supposed to equal "project_uuid" on root components, but ok, this can be a safeguard.

PR #2143, opened 2017-06-02
SONAR-9028 add delete non root components for governance

sns-seb:


server/sonar-db-dao/src/main/java/org/sonar/db/purge/PurgeDao.java
```diff
@@ -171,6 +180,55 @@ private static void deleteProject(String rootUuid, PurgeMapper mapper, PurgeComm
...
+    purgeCommands.deleteByRootAndModulesOrSubviews(modulesOrSubviews);
+    List<String> nonRootComponentUuids = nonRootComponents.stream().map(ComponentDto::uuid).collect(MoreCollectors.toList(nonRootComponents.size()));
+    purgeCommands.deleteComponentMeasures(nonRootComponentUuids);
+    purgeCommands.deleteFileSources(nonRootComponents.stream().filter(PurgeDao::isFile).map(ComponentDto::uuid).collect(MoreCollectors.toList()));
+    purgeCommands.deleteIssues(nonRootComponentUuids);
+    purgeCommands.deleteComponents(nonRootComponentUuids);
+  }
+
+  private static boolean isNotRoot(ComponentDto dto) {
+    return !(SCOPE_PROJECT.equals(dto.scope()) && QUALIFIERS_PROJECT_VIEW.contains(dto.qualifier()));
```
simonbrandhof: `ComponentDto#isRootProject()` should be used. Formula is a bit different.

sns-seb: indeed the algorithm is different, it either relies on `module_uuid` in addition to scope (for method `isRootProject`) or solely on `uuid_path` (for method `isRoot`).

I'd rather keep to the column I already use in the methodfor other parts of the algorithm (eg. to identify modules/subview or files). It makes the whole thing simpler because more consistent.

PR #2144, opened 2017-06-05
SONAR-9356 Add column "show_onboarding" to db table "users"

bartfastiel:
* Boolean
* Not nullable
* Initialized with false (but no default value)
* true for all new users


bartfastiel:
This change will be part of a future PR, possibly #2146

PR #2150, opened 2017-06-06
apply feedback of built-in profiles

stas-vilchik:


server/sonar-web/src/main/js/apps/quality-profiles/utils.js
```diff
@@ -42,7 +42,11 @@ export function sortProfiles(profiles: Array<Profile>) {
     children.forEach(child => putProfile(child, depth + 1));
   }
 
-  putProfile();
+  sorted
+    .filter(
+      profile => profile.parentKey == null || sorted.find(p => p.key === profile.parentKey) == null
```
gregaubert: A quality profile inheriting from another one is not correctly displayed on the quality profiles page. It should be indented, but it's not. It seem's like it only happens for first level inheritance.

stas-vilchik: the problem was that a profile `depth` started from 0 instead of 1, which results in the negative padding of the first element

PR #2159, opened 2017-06-09
SONAR-9374 Update page titles for SonarCloud mode

gregaubert:


server/sonar-web/src/main/js/app/components/DefaultHelmetContainer.js
```diff
@@ -18,13 +18,21 @@
...
+import { connect } from 'react-redux';
 import Helmet from 'react-helmet';
+import { getSettingValue } from '../../store/rootReducer';
 
-export default function DefaultHelmetContainer({ children }) {
+function DefaultHelmetContainer({ children, isSonarCloud }) {
   return (
     <div>
-      <Helmet defaultTitle="SonarQube" />
+      <Helmet defaultTitle={isSonarCloud ? 'SonarCloud' : 'SonarQube'} />
```
stas-vilchik: `isSonarCloud` can be `null` or `{ value: 'true' }` or `{ value: 'false' }`. When it's `{ value: 'false' }`, the title is wrong.

gregaubert: :scream: Thanks!

gregaubert:
@stas-vilchik I fixed it as well as the `GlobalFooter`, and also renamed `isSonarCloud` back to `sonarqubeDotCom` so we will not miss it when doing the global renaming ticket.

PR #2160, opened 2017-06-09
Update flow to last version

gregaubert:


server/sonar-web/src/main/less/components/modals.less
```diff
@@ -31,6 +31,10 @@
   background-color: #fff;
   opacity: 0;
   transition: all 0.2s ease;
+
+  &:focus {
```
stas-vilchik: 🙀

gregaubert: oups added this by mistake!

gregaubert:
@stas-vilchik Did you check carefully the last commit ? Fix flow issue for the SourceViewerBase, I wasn't totally sure about the fix for this one.

PR #2163, opened 2017-06-12
SONAR-9357 Hide default login form if there are other ways to login

stas-vilchik:


server/sonar-web/src/main/js/apps/sessions/components/LoginForm.js
```diff
@@ -22,26 +22,49 @@ import React from 'react';
...
+  handleMoreOptionsClick = (event: Event) => {
+    event.preventDefault();
+    this.setState({ collapsed: false });
+  };
+
   render() {
     return (
       <div>
-        <h1 className="maintenance-title text-center">Log In to SonarQube</h1>
+        <h1 className="maintenance-title text-center">{translate('login.login_to_sonarqube')}</h1>
```
gregaubert: We need a `login.login_to_sonarclaude` too

server/sonar-web/src/main/js/apps/sessions/components/LoginForm.js
```diff
@@ -22,26 +22,49 @@ import React from 'react';
...
+    event.preventDefault();
+    this.setState({ collapsed: false });
+  };
+
   render() {
     return (
-      <div>
-        <h1 className="maintenance-title text-center">Log In to SonarQube</h1>
+      <div id="login_form">
+        <h1 className="maintenance-title text-center">{translate('login.login_to_sonarqube')}</h1>
```
gregaubert: We need a `login.login_to_sonarclaude` too

stas-vilchik: the request to settings api to known if the current instance is sonarcloud is not done for the login page (and it's not even possible if the forced authentication is enabled)

PR #2178, opened 2017-06-16
upgrade webpack

stas-vilchik:


stas-vilchik:
@gregaubert ping me when you start a review

PR #2182, opened 2017-06-19
update projects page layout

stas-vilchik:


server/sonar-web/src/main/js/apps/projects/components/ProjectsSortingSelect.js
```diff
@@ -88,7 +88,7 @@ export default class ProjectsSortingSelect extends React.PureComponent {
       <div className={this.props.className}>
         <label>{translate('projects.sort_by')}:</label>
         <Select
-          className="little-spacer-left input-large"
+          className="little-spacer-left input-medium"
```
gregaubert: This is a bit small when the last analysis is the default sorting

PR #2183, opened 2017-06-19
use new trusty image in group edge

tomverin:


tomverin:
No issue on travis side so I close this PR

PR #2184, opened 2017-06-19
remember last selected options on projects page

stas-vilchik:


it/it-tests/src/test/java/it/projectSearch/ProjectsPageTest.java
```diff
@@ -55,6 +57,11 @@ public static void setUp() {
     ORCHESTRATOR.executeBuild(SonarScanner.create(projectDir("duplications/file-duplications")).setProjectKey("key-bar"));
   }
 
+  @Before
+  public void before() {
+    clearBrowserLocalStorage();
```
stas-vilchik: @simonbrandhof I wonder if we could do it always

simonbrandhof: Yes we could, even my first tests failed. I'll do that in my branch feature/sb/it-refactoring.

server/sonar-web/src/main/js/apps/projects/components/AllProjects.js
```diff
@@ -80,24 +89,38 @@ export default class AllProjects extends React.PureComponent {
...
+          query.sort = (sort.sortDesc ? '-' : '') + utils.SORTING_SWITCH[sort.sortValue];
         }
       }
       this.props.router.push({ pathname: this.props.location.pathname, query });
     } else {
       this.updateLocationQuery(query);
     }
+
+    utils.saveView(view);
+    utils.saveVisualization(visualization);
```
gregaubert: Sort should be saved here too, otherwise if you : 
1. Choose leak perspective
2. Choose a leak sorting option
3. Choose the overall perspective
4. Go to issues, come back to project
5. You have the wrong sort selected and remembered in the localstorage

Or maybe all `saveX` actions should be done in `updateLocationQuery`

server/sonar-web/src/main/js/apps/projects/components/AllProjects.js
```diff
@@ -72,6 +75,12 @@ export default class AllProjects extends React.PureComponent {
 
   isFiltered = () => Object.keys(this.state.query).some(key => this.state.query[key] != null);
 
+  getSavedOptions = () => ({
+    sort: utils.getSort(),
+    view: utils.getView(),
+    visualization: utils.getVisualization()
```
gregaubert: We end up with some parameters set to nothing if we don't have values in the localstorage, maybe they should also not be set if the default values are selected like "overall", "name" or "risk".

gregaubert: I'm not sure if this logic should be done here where we return the saved options, or where we save them or where retrieve them from the localstorage.

PR #2187, opened 2017-06-20
replace require.ensure with dymamic import

stas-vilchik:


stas-vilchik:
this PR opens door for great [react-loadable](https://github.com/thejameskyle/react-loadable) project

server/sonar-web/src/main/js/apps/coding-rules/routes.js
```diff
@@ -21,9 +21,9 @@ const routes = [
   {
     indexRoute: {
       getComponent(_, callback) {
-        require.ensure([], require => {
-          callback(null, require('./components/CodingRulesAppContainer').default);
-        });
+        import('./components/CodingRulesAppContainer').then(i =>
+          callback(null, { component: i.default })
```
gregaubert: Are you sure we must return an object with the `component` field here ? Seems like we were returning the component directly with the `require.ensure`

gregaubert: Ok I just saw in the next file that the component should be returned directly in this case.

PR #2188, opened 2017-06-20
SONAR-9440 All analyses are returned in WS api/measures/search_history even if there is no measure

teryk:


server/sonar-server/src/test/java/org/sonar/server/measure/ws/SearchHistoryActionTest.java
```diff
@@ -249,7 +251,9 @@ public void do_not_return_developer_measures() {
     SearchHistoryResponse result = call();
 
     assertThat(result.getMeasuresCount()).isEqualTo(1);
-    assertThat(result.getMeasures(0).getHistoryCount()).isEqualTo(0);
+    assertThat(result.getMeasures(0).getHistoryCount()).isEqualTo(1);
+    assertThat(result.getMeasures(0).getHistory(0).hasDate()).isTrue();
+    assertThat(result.getMeasures(0).getHistory(0).hasValue()).isFalse();
   }
```
julienlancelot: I think this fix would require a dedicated test in order to perfectly understand what happen when there's no measure on a analysis

PR #2211, opened 2017-06-30
Hardening

teryk:
👨‍🏭 🚧 👷

### DIFF
--- a/server/sonar-db-dao/src/main/java/org/sonar/db/metric/MetricDtoValidator.java
+++ b/server/sonar-db-dao/src/main/java/org/sonar/db/metric/MetricDtoValidator.java
@@ -1,0 +1,63 @@
+/*
+ * SonarQube
+ * Copyright (C) 2009-2017 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.sonar.db.metric;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nullable;
+
+import static com.google.common.base.Preconditions.checkArgument;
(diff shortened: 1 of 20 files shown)


server/sonar-db-dao/src/main/java/org/sonar/db/metric/MetricDao.java
```diff
@@ -81,7 +82,7 @@ public int countEnabled(DbSession session, @Nullable Boolean isCustom) {
   }
 
   public MetricDto insert(DbSession session, MetricDto dto) {
-    mapper(session).insert(dto);
+    mapper(session).insert(validate(dto));
```
julienlancelot: AFAIK, It's not the responsability of DAO to validate dto. This should be in a higher layer, a XXXUpdater or the WS

julienlancelot: Or maybe in the dto setters

server/sonar-server/src/main/java/org/sonar/server/measure/index/ProjectMeasuresIndex.java
```diff
@@ -348,15 +348,16 @@ private static QueryBuilder toValueQuery(MetricCriterion criterion) {
     }
   }
 
-  public List<String> searchTags(@Nullable String textQuery, int pageSize) {
-    checkArgument(pageSize <= 100, "Page size must be lower than or equals to " + 100);
-    if (pageSize == 0) {
+  public List<String> searchTags(@Nullable String textQuery, int size) {
+    int maxPageSize = 500;
```
julienlancelot: Please don't forget to add this WS in the ticket

server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
```diff
@@ -177,6 +172,160 @@ public IssueIndex(EsClient client, System2 system, UserSession userSession, Auth
     this.sorting.addDefault(IssueIndexDefinition.FIELD_ISSUE_KEY);
   }
 
+  /**
```
julienlancelot: Not sure it was a good idea to reformat the class in this commit as it's now super hard to see what has changed...

server/sonar-db-dao/src/main/java/org/sonar/db/metric/MetricDtoValidator.java
```diff
@@ -0,0 +1,103 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.function.Consumer;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+public class MetricDtoValidator {
```
julienlancelot: The JIRA ticket should contains really what have you done, as the ticket is only about description, and if my understanding is correct you have added more checks than only on description

server/sonar-server/src/main/java/org/sonar/server/issue/index/IssueIndex.java
teryk: indeed 😭

PR #2214, opened 2017-07-03
Fix reactivating user as local

julienlancelot:


server/sonar-server/src/test/java/org/sonar/server/user/UserUpdaterCreateTest.java
```diff
@@ -0,0 +1,557 @@
+/*
```
teryk: remark: outside the scope of the PR, but I see several test classes for one production class as a sign to separate responsabilities

server/sonar-server/src/main/java/org/sonar/server/authentication/UserIdentityAuthenticator.java
```diff
@@ -124,8 +124,7 @@ private void registerExistingUser(DbSession dbSession, UserDto userDto, UserIden
     userUpdater.update(dbSession, UpdateUser.create(userDto.getLogin())
       .setEmail(user.getEmail())
       .setName(user.getName())
-      .setExternalIdentity(new ExternalIdentity(provider.getKey(), user.getProviderLogin()))
-      .setPassword(null));
```
teryk: why have you removed this line?

PR #2225, opened 2017-07-07
SONAR-9508 Display worker counter in background tasks page

stas-vilchik:


server/sonar-web/src/main/js/apps/background-tasks/components/WorkersForm.js
```diff
@@ -0,0 +1,124 @@
...
+      () => {
+        if (this.mounted) {
+          this.setState({ submitting: false });
+        }
+      }
+    );
+  };
+
+  render() {
+    const options = times(10).map((_, i) => ({ label: i + 1, value: i + 1 }));
```
gregaubert: Maybe use a constant here:
```
const MAX_WORKERS = 10;
times(MAX_WORKERS)...
```

PR #2226, opened 2017-07-07
SONAR-9403 Add custom graph to the project activity page

gregaubert:


gregaubert:
@stas-vilchik Tell me if you don't want to review this, since you couldn't review a big part of the sprint.

gregaubert:
@stas-vilchik Thanks! I'll fix the issues that you found, thanks.
Yes the last graph that you selected or created is remembered in the localstorage and we show this one too on the project dashboard.
Yes that's correct, for now since we have only one custom graph available, as soon as you select a metric the other types are filtered out. I'll add a message to explain that to the user.

gregaubert:
@stas-vilchik I applied some of the feedback and added two commits for the remaining two tickets that were not done yet.

server/sonar-web/src/main/js/apps/issues/sidebar/CreationDateFacet.js
```diff
@@ -178,10 +173,8 @@ export default class CreationDateFacet extends React.PureComponent {
       </div>
     );
   }
-
```
stas-vilchik: there are weird changes in this file (some lines are removed)

gregaubert:
@stas-vilchik Okay so both bugs you showed me are fixed and I also fixed a problem with missing data that were not displayed on the preview graph of the project dashboard.

stas-vilchik:
* there is an overflow problem on the project homepage
* there is a little problem when I move both handles to the very right

PR #2251, opened 2017-07-13
Update Jest test runner to version 20.0.4

gregaubert:
Trying to update the version of Jest to prevent a memory leak that is affecting our tests : https://github.com/facebook/jest/issues/1893

Running the tests with Jest version 19.x would use up to 1.6Gb Ram, making the tests fail on Travis.

Unfortunately updating to version 20.0.4 does not bring a big improvement, we only went down to 1.4Gb consumption, so the leak is still there.

Jest is using babel-polyfill automatically if the module is available, so trying to remove the dependency to babel-polyfill shows a little better results, down to max 900Mb Ram used.
But that would mean we need to drop babel-polyfill and it's still used for older browsers.


gregaubert:
@stas-vilchik We probably need to also merge this, since the other PR is using a `ts-jest` version that needs `jest v20.0.4`

PR #2254, opened 2017-07-14
upgrade prettier and remove prettier-eslint

stas-vilchik:


stas-vilchik:
@gregaubert yet another upgrade :)

don't forget to update your editor configuration, because it's no more inferred from the eslint config

server/sonar-web/.eslintrc
```diff
@@ -48,7 +48,7 @@
     "jsx-quotes": 2,
     "key-spacing": 2,
     "keyword-spacing": 2,
-    "max-len": [2, 100],
+    // "max-len": [2, 100],
```
gregaubert: Can we remove it ? Or is it intended ?

stas-vilchik: I kept it for the memory ;)

PR #2255, opened 2017-07-17
SONAR-9546 Two custom graph on project activity page

gregaubert:


gregaubert:
@stas-vilchik You might see a lot of change in tests and snapshot (like missing `style` parameter or change of orders in array), they are due on the fact that I changed the way the style is selected for a serie, now it's based on the order in the array instead of the parameter `style`.

server/sonar-web/src/main/js/apps/projectActivity/components/ProjectActivityGraphs.js
```diff
@@ -120,6 +133,18 @@ export default class ProjectActivityGraphs extends React.PureComponent {
     }
   };
 
+  getMetricsTypeFilter = (): ?Array<string> => {
+    if (this.state.graphs.length < MAX_GRAPH_NB) {
+      return null;
+    }
+    return this.state.graphs.reduce((acc, graph) => {
```
stas-vilchik: this is a matter of preference, but I like this more (it's more readable IMO):
```javascript
return this.state.graphs
  .filter(graph => graph.length < MAX_SERIES_PER_GRAPH)
  .map(graph => graph[0].type);
```

gregaubert: :+1:

gregaubert:
@stas-vilchik I rebased on the changes from 6.5 and had to update some part of it.

PR #2265, opened 2017-07-18
SONAR-9567 Divide by 2 the size of the Elasticsearch index for issues

simonbrandhof:


server/sonar-server/src/main/java/org/sonar/server/batch/IssuesAction.java
```diff
@@ -77,38 +78,53 @@ public void define(WebService.NewController controller) {
...
+      switch (component.scope()) {
+        case Scopes.PROJECT:
+          dbClient.issueDao().scrollNonClosedByModuleOrProject(dbSession, component, handler);
+          break;
+        case Scopes.FILE:
+          dbClient.issueDao().scrollNonClosedByComponentUuid(dbSession, component.uuid(), handler);
+          break;
+        default:
+          // only projects, modules and files are supported. Other types of components are not allowed.
+          throw new IllegalStateException(format("Component of scope '%s' is not allowed", component.scope()));
```
julienlancelot: I would propose to replace this ISA by a IAE, as it's a functional error

simonbrandhof: Correct. I will change (even if it's out-of-scope of this PR).
@henryju @dbmeneses @janos-ss is it a problem for SonarLint if WS batch/issues returns 400 instead of 500 when requested component is not a project, module or file ?

henryju: I don't think it is a problem.

dbmeneses: no problem AFAIK

PR #2273, opened 2017-07-20
MMF-941 Unified fixed-width layout

stas-vilchik:


server/sonar-web/pom.xml
```diff
@@ -230,7 +230,7 @@
         <dependency>
           <groupId>com.sonarsource</groupId>
           <artifactId>sonarsource-branding</artifactId>
-          <version>1.3.0.304</version>
+          <version>1.3.0.307</version>
```
gregaubert: What is this ?

server/sonar-web/src/main/js/components/ui/OrganizationIcon.js
```diff
@@ -20,10 +20,17 @@
 // @flow
 import React from 'react';
 
-export default function OrganizationIcon() {
+type Props = { className?: string, size?: number };
+
+export default function OrganizationIcon({ className, size = 16 }: Props) {
```
gregaubert: Could we move it to `js/components/icons-components/` folder ?

server/sonar-web/pom.xml
stas-vilchik: https://github.com/SonarSource/sonarsource-branding

It contains SonarQube logo and the "official" footer.

gregaubert: Ah ok, and it's because you updated the logo on sonarsource-branding. Got it, thanks!

PR #2275, opened 2017-07-20
SONAR-9576 Add analysis mode in the scanner report

henryju:


sonar-scanner-engine/src/main/java/org/sonar/scanner/task/ScanTask.java
```diff
@@ -44,6 +45,8 @@ public ScanTask(TaskContainer taskContainer, TaskProperties taskProps) {
   @Override
   public void execute() {
     AnalysisProperties props = new AnalysisProperties(taskProps.properties(), taskProps.property(CoreProperties.ENCRYPTION_SECRET_KEY_PATH));
-    new ProjectScanContainer(taskContainer, props).execute();
+    ProjectScanContainer scanContainer = new ProjectScanContainer(taskContainer, props);
+    scanContainer.add(DefaultAnalysisMode.class);
+    scanContainer.execute();
```
dbmeneses: Do we need it so early? It's already in the ProjectScanContainer.

henryju: I have removed it from the ProjectScanContainer: https://github.com/SonarSource/sonarqube/pull/2275/files#diff-413a4831d77c1e1fec55d5d3e191f7e1

dbmeneses: Is there a reason? I think it's better to have all classes added to the container in one place.

PR #2279, opened 2017-07-21
Feature/greg/6.5 hardening

gregaubert:


server/sonar-web/src/main/js/apps/overview/events/PreviewGraph.js
```diff
@@ -139,7 +143,7 @@ export default class PreviewGraph extends React.PureComponent {
         tabIndex={0}
         role="link">
         <AutoSizer disableHeight={true}>
-          {({ width }) => (
+          {({ width }) =>
```
stas-vilchik: I think you were using prettier from master

PR #2296, opened 2017-07-26
SONAR-9584 Add root project key to response of api/components/search

henryju:


server/sonar-db-dao/src/main/resources/org/sonar/db/component/ComponentMapper.xml
```diff
@@ -238,18 +238,21 @@
 
   <select id="selectByQuery" resultType="Component">
     select
-      <include refid="componentColumns"/>
+      <include refid="componentColumns"/>,
```
julienlancelot: Warning !!! We should not add a join for a single use case !!!

server/sonar-db-dao/src/main/java/org/sonar/db/component/ComponentDto.java
```diff
@@ -221,6 +226,18 @@ public ComponentDto setProjectUuid(String projectUuid) {
     return this;
   }
 
+  /**
+   * Return the root project key. On a root project, return itself
+   */
+  public String projectKey() {
```
julienlancelot: Description is wrong : this method only returns something when selectByQuery is used.......

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchAction.java
```diff
@@ -179,6 +179,7 @@ private static SearchWsResponse buildResponse(List<ComponentDto> components, Org
       .setOrganization(organization.getKey())
       .setId(dto.uuid())
       .setKey(dto.key())
+      .setProjectKey(dto.projectKey())
```
julienlancelot: The project key should NOT come from the DTO.
We should : 
1. Get all project uuids from searched components
2. Get all projects dto from project uuids => create a map of dto by project uuid 
3. Use this map to get the project key of a component

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchAction.java
```diff
@@ -120,11 +125,27 @@ private SearchWsResponse doHandle(SearchWsRequest request) {
...
+      return buildResponse(components, organization, projectKeysByUuids, paging);
     }
   }
 
+  private Map<String, String> searchProjectsKeysByUuids(DbSession dbSession, List<ComponentDto> components) {
+    // Some of the components are maybe projects themself
+    Map<String, String> projectsKeysByUuid = components.stream()
+      .filter(c -> c.uuid().equals(c.projectUuid()))
+      .collect(toMap(ComponentDto::uuid, ComponentDto::key));
+    Predicate<String> alreadyKnowProjectKey = projectsKeysByUuid::containsKey;
```
julienlancelot: This optimization is not needed

sonar-ws/src/main/protobuf/ws-components.proto
```diff
@@ -118,6 +118,8 @@ message Component {
   optional Tags tags = 14;
   optional string visibility = 15;
   optional string leakPeriodDate = 16;
+  // Root project key
+  optional string projectKey = 17;
```
julienlancelot: It should not be `projectKey` but `project`

PR #2298, opened 2017-07-27
implement an easy/non check https client on demand, secure per default

olamy:
Using system properties to enable an non secure mode.
So users are be able to use self signed certificates. 
see https://github.com/SonarSource/sonar-scanner-api/pull/45
NOTE same sys properties are used. 
Not sure if you guys have any plan to have a common shared library to handle http call.
I can see some duplicate code with org.sonarsource.scanner.api.internal.OkHttpClientFactory and org.sonarqube.ws.client.OkHttpClientBuilder just saying :P

### DIFF
--- a/sonar-scanner-engine/src/main/java/org/sonar/scanner/bootstrap/ScannerWsClientProvider.java
+++ b/sonar-scanner-engine/src/main/java/org/sonar/scanner/bootstrap/ScannerWsClientProvider.java
@@ -26,6 +26,16 @@ import org.sonar.batch.bootstrapper.EnvironmentInformation;
 import org.sonarqube.ws.client.HttpConnector;
 import org.sonarqube.ws.client.WsClientFactories;
 
+import javax.net.ssl.KeyManager;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+
 import static java.lang.Integer.parseInt;
 import static java.lang.String.valueOf;
 import static org.apache.commons.lang.StringUtils.defaultIfBlank;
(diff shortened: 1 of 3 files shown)


dbmeneses:
I wonder if this change is needed. When using self signed certificates, why not add the certificate to JVM's truststore?
We weren't checking server certificates in the past and we made a considerable effort to make all connections secure using JVM's default truststore and keystore.

About the duplication, from my experience in cases like this it's easier to maintain a duplicated class than to extract into a third component and maintain it.

sonar-scanner-engine/src/main/java/org/sonar/scanner/bootstrap/ScannerWsClientProvider.java
```diff
@@ -58,7 +68,44 @@ public synchronized ScannerWsClient provide(final GlobalProperties settings, fin
...
+          sslContext.init( new KeyManager[0], new TrustManager[]{easyTrust}, new java.security.SecureRandom());
+          // get an ssl socket factory with our all trusting manager
+          SSLSocketFactory sslSocketFactory = sslContext.getSocketFactory();
+          connectorBuilder.setSSLSocketFactory( sslSocketFactory ).setTrustManager( easyTrust );
+        } catch ( NoSuchAlgorithmException | KeyManagementException e ) {
+          // should not happen but just in case log it
+          throw new RuntimeException( e.getMessage(), e );
+        }
+      }
+      if(Boolean.valueOf( System.getProperty( "sonar.http.ssl.allowall", "false" ) )) {
```
janos-ss: Not the usual formatting we use, in many places in this file, and also in the others.

olamy:
@dbmeneses when you use Jenkins with dynamic nodes from a cloud provider it's not so easy to install certificates....
regarding code duplication well maybe.... (I guess people have different experience :-) ) Anyway I was only joking about maybe Sonarqube could complain about code duplication :P

sonar-scanner-engine/src/main/java/org/sonar/scanner/bootstrap/ScannerWsClientProvider.java
olamy: @janos-ss  do you have any configuration file for intellij?
should be better now.

janos-ss: @olamy You can find it here: https://github.com/SonarSource/sonar-developer-toolset

I was mainly picking on `if(Boolean.valueOf( System.getProperty( "sonar.http.ssl.allowall", "false" ) )) {` instead of the default formatting style in both IntelliJ and Eclipse which would be `if (Boolean.valueOf(System.getProperty("sonar.http.ssl.allowall", "false"))) {`. And I see several other cases like this, still in the updated code.

olamy: @janos-ss ok. Anyway I changed to Boolean.getBoolean as it's even shorter :-)

janos-ss: @olamy Anyway don't worry about the formatting yet, while we are not convinced if the change should be merged.

olamy:
ok forget it. I will try to use javax.net.ssl.keyStore configuration....

dbmeneses:
@olamy ok, please let us know if you find a specific case where it doesn't work.

PR #2305, opened 2017-07-31
SONAR-8550 Update function to group analyses by version

gregaubert:


gregaubert:
@stas-vilchik This fix is for 6.5, the version badges were misplaced on the activity list

server/sonar-web/src/main/js/apps/projectActivity/utils.js
```diff
@@ -141,10 +141,18 @@ export const getAnalysesByVersionByDay = (
   byDay: { [string]: Array<Analysis> }
 }> =>
   analyses.reduce((acc, analysis) => {
-    if (acc.length === 0) {
-      acc.push({ version: undefined, key: undefined, byDay: {} });
+    let currentVersion = acc[acc.length - 1];
+    const versionEvent = analysis.events.find(event => event.category === 'VERSION');
+    if (versionEvent && versionEvent.category === 'VERSION') {
```
stas-vilchik: why do you need the second check `versionEvent.category === 'VERSION'`? It's already so, because you're searching only for such events.

server/sonar-web/src/main/js/apps/projectActivity/__tests__/utils-test.js
```diff
@@ -21,13 +21,17 @@
 import * as utils from '../utils';
 
 const ANALYSES = [
-  { key: 'AVyMjlK1HjR_PLDzRbB9', date: new Date('2017-06-09T13:06:10+0200'), events: [] },
```
gregaubert: I updated it because the first analysis is supposed to always have a version event and it was not the case.

server/sonar-web/src/main/js/apps/projectActivity/utils.js
gregaubert: True, I don't need to, it was just a dumb copy past of the previous condition, I'm gonna remove it, thanks!

PR #2313, opened 2017-08-03
use POST instead GET in order to get settings values.

warlof:
This is still RESTFull since the api endpoint is using filters which are defined by the user in order to obtain specific information.

Allow usage of proxy without issue related to URI length limit.
(ie: specific IIS reverse SSL Proxy setup https://blogs.msdn.microsoft.com/visualstudioalmrangers/2016/06/04/running-sonarqube-behind-an-iis-reversed-proxy/)

### DIFF
--- a/server/sonar-web/src/main/js/api/settings.js
+++ b/server/sonar-web/src/main/js/api/settings.js
@@ -36,7 +36,7 @@ export function getValues(keys, componentKey) {
   if (componentKey) {
     data.component = componentKey;
   }
-  return getJSON(url, data).then(r => r.settings);
+  return postJSON(url, data).then(r => r.settings);
 }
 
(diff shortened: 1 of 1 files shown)


julienlancelot:
Thanks for this PR but we really want to always use GET method when no update is done.
A ticket has been created to decrease the URI size of this call : https://jira.sonarsource.com/browse/SONAR-9325, it should automatically fix this issue without having to do any "hack".

warlof:
Indead, it makes sense and looks suitable to the proxy issue due to the header size.

I'm not experienced enough with java and maven project, therefore it's still hard for me to debug such change.

julienlancelot:
@warlof No problem, thanks anyway to have created this PR (and closed it), it increases the need to fix SONAR-9325 !

PR #2314, opened 2017-08-03
Clarify description of CoreProperties.SERVER_BASE_URL

TheBigAleski:
Added use by plugins for clarity
-> https://github.com/SonarQubeCommunity/sonar-build-breaker/issues/12


henryju:
@simonbrandhof I find this change too technical and not valuable for end users. WDYT?

simonbrandhof:
Same feeling. It targets plugin developers but not end-users.
Thanks anyway for the contribution.

PR #2326, opened 2017-08-07
Small project activity page fixes

gregaubert:


server/sonar-web/src/main/js/apps/overview/events/PreviewGraph.js
```diff
@@ -139,44 +141,49 @@ export default class PreviewGraph extends React.PureComponent {
...
-                  graph={graph}
-                  graphWidth={width}
-                  metrics={this.props.metrics}
-                  selectedDate={selectedDate}
-                  series={series}
-                  tooltipIdx={tooltipIdx}
-                  tooltipPos={tooltipXPos}
-                />}
+        {hasHistoryDataValue(series)
+          ? <AutoSizer disableHeight={true}>
```
stas-vilchik: Could you extract it to a dedicated function please (like `renderTimeline`)? It'd increase the readability.

gregaubert: You are right, I'll do it

PR #2340, opened 2017-08-10
MFF-871 Apply feedbacks

gregaubert:


gregaubert:
@stas-vilchik OK I'm done with the feedback, I don't know if you reviewed everything yet or not, there is some commit at the end that might have changed a bit.

PR #2358, opened 2017-08-15
Configure for use with Typescript

gregaubert:


gregaubert:
@stas-vilchik I had to fix the transform section for jest. `babel-jest` was not run anymore since I changed it. So all other tests were failing :scream:

PR #2361, opened 2017-08-15
Feature/dm/branches

dbmeneses:


server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/issue/IssueTrackingDelegator.java
```diff
@@ -19,28 +19,57 @@
...
+
+  /**
+   * Special case where we want to do the issue tracking with the merge branch, and copy matched issue to the current branch.
+   */
+  private boolean isFirstAnalysisSecondaryLongLivingBranch() {
+    if (analysisMetadataHolder.isFirstAnalysis()) {
+      Optional<Branch> branch = analysisMetadataHolder.getBranch();
+      if (branch.isPresent()) {
+        return !branch.get().isMain() && branch.get().getType() == BranchType.LONG;
+      }
```
janos-ss: Not exactly the same, but probably good enough: `return branch.map(b -> !b.isMain() && b.getType() == BranchType.LONG).orElse(false);`

PR #2362, opened 2017-08-15
SONAR-9708 apply feedback

stas-vilchik:


server/sonar-web/src/main/js/apps/background-tasks/components/Tasks.js
```diff
@@ -86,7 +85,6 @@ export default class Tasks extends React.PureComponent {
               index={index}
               tasks={tasks}
               component={component}
-              types={types}
```
gregaubert: `types` is style required in the prop types of `Task` component

PR #2372, opened 2017-08-17
make quality gate ITs more reliable, by not using default qg (2)

bartfastiel:
This tries to fix false positives on QualityGateTest::does_not_fail_when_condition_is_on_removed_metric


tests/src/test/java/org/sonarqube/tests/qualityGate/QualityGateTest.java
```diff
@@ -257,23 +260,37 @@ public void ad_hoc_build_break_strategy() throws IOException {
 
   @Test
   public void does_not_fail_when_condition_is_on_removed_metric() throws Exception {
+
+    // create project
+    Project project = tester.projects().generate(null);
```
julienlancelot: why this `null` ?

bartfastiel: Organizations are not enabled in `Category1`, so we have to use the default organization. :(

julienlancelot: Ok !

PR #2373, opened 2017-08-17
SONAR-9628 Branches must not appear in administration of projects

dbmeneses:


server/sonar-db-dao/src/main/java/org/sonar/db/component/ComponentDao.java
```diff
@@ -104,19 +104,31 @@ public ComponentDto selectOrFailByUuid(DbSession session, String uuid) {
...
 
+  /**
+   * Branches are filtered out. 
+   */
   public List<ComponentDto> selectByQuery(DbSession session, ComponentQuery query, int offset, int limit) {
     return selectByQueryImpl(session, null, query, offset, limit);
   }
 
+  /**
+   * Branches are filtered out.
```
julienlancelot: No need for such javadoc, branches should only be returned by selectByKeyAndBranch and selectByKeysAndBranch

PR #2391, opened 2017-08-21
SONAR-9616 Write main branch to DB even if plugin is not installed

dbmeneses:


server/sonar-server/src/main/java/org/sonar/server/component/ComponentUpdater.java
```diff
@@ -100,6 +103,17 @@ private ComponentDto createRootComponent(DbSession session, NewComponent newComp
       .setPrivate(newComponent.isPrivate())
       .setCreatedAt(new Date(system2.now()));
     dbClient.componentDao().insert(session, component);
+
+    BranchDto branch = new BranchDto()
```
julienlancelot: Warning, this class is used by Governance to create View.
And we should not create main branch on views.

I think the creation of the main branch should only be done during the analysis, not during the provisioning.

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/component/BranchPersisterDelegate.java
```diff
@@ -19,12 +19,96 @@
...
+import org.sonar.server.computation.task.projectanalysis.analysis.AnalysisMetadataHolder;
+import org.sonar.server.computation.task.projectanalysis.analysis.Branch;
 
-@ComputeEngineSide
-public interface BranchPersisterDelegate {
+import static org.sonar.db.component.ComponentDto.UUID_PATH_OF_ROOT;
+import static org.sonar.db.component.ComponentDto.UUID_PATH_SEPARATOR;
 
-  void persist(DbSession dbSession);
+public class BranchPersisterDelegate {
```
julienlancelot: The name of the class should be changed, as it's no more a delegate

julienlancelot: => BranchPersister

dbmeneses: It is still a delegate, as much as it was before

server/sonar-server/src/main/java/org/sonar/server/component/ComponentUpdater.java
dbmeneses: it needs to be created as soon as the project is created if we want it to always be consistent. I will change it to only be done when it's a regular project.

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/component/BranchPersisterDelegate.java
dbmeneses: That being said, I agree with the name simplification :)

PR #2401, opened 2017-08-22
SONAR-7493 fix length of issue message for Oracle DB

simonbrandhof:


sns-seb:
commit message is too long

> SONAR-7493 do not fail on Oracle when inserting issues with too long message

PR #2404, opened 2017-08-22
SONAR-9745 Drop support of Microsoft SQLServer 2012

simonbrandhof:


server/sonar-db-core/src/test/java/org/sonar/db/DatabaseCheckerTest.java
```diff
@@ -108,6 +109,23 @@ public void test_mysql() throws Exception {
   }
 
   @Test
+  public void mssql_2012_is_not_supported() throws Exception {
+    expectedException.expect(MessageException.class);
+    expectedException.expectMessage("Unsupported mssql version: 11.0. Minimal supported version is 12.0.");
+
+    Database db = mockDb(new MsSql(), 11, 0, "6.1");
```
sns-seb: "6.1" is a bad copy/paste, right?

server/sonar-db-core/src/test/java/org/sonar/db/DatabaseCheckerTest.java
```diff
@@ -108,6 +109,23 @@ public void test_mysql() throws Exception {
...
+    new DatabaseChecker(db).start();
+    // no error
+  }
+
+  @Test
+  public void mssql_2014_is_supported() throws Exception {
+    Database db = mockDb(new MsSql(), 12, 0, "6.1");
```
sns-seb: "6.1" is a bad copy/paste, right?

sns-seb: a test that an intermediary version such as "12.0.4100.1" would be nice

server/sonar-db-core/src/test/java/org/sonar/db/DatabaseCheckerTest.java
simonbrandhof: nope, it's version of MS driver

server/sonar-db-core/src/test/java/org/sonar/db/DatabaseCheckerTest.java
simonbrandhof: in fact only the major and minor versions are loaded from JDBC metadata. Complete version is not used.

PR #2459, opened 2017-08-31
SONAR-8640 WS api/projects/search query on key does partial match and is case insensitive

teryk:


server/sonar-db-dao/src/test/java/org/sonar/db/component/ComponentDaoTest.java
```diff
@@ -965,6 +965,7 @@ public void selectByQuery_name_with_special_characters() {
   @Test
   public void selectByQuery_key_with_special_characters() {
     db.components().insertProjectAndSnapshot(ComponentTesting.newPrivateProjectDto(db.organizations().insert()).setDbKey("project-_%-key"));
+    db.components().insertProjectAndSnapshot(ComponentTesting.newPrivateProjectDto(db.organizations().insert()).setDbKey("project-key-that-does-not-match"));
```
julienlancelot: I would do a static import of ComponentTesting to improve readability

PR #2462, opened 2017-09-01
SONAR-9616 Fix notification on branch

julienlancelot:


server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/step/SendIssueNotificationsStep.java
```diff
@@ -55,7 +63,8 @@
   private final TreeRootHolder treeRootHolder;
   private final NotificationService service;
   private final AnalysisMetadataHolder analysisMetadataHolder;
-  private NewIssuesNotificationFactory newIssuesNotificationFactory;
+  private final NewIssuesNotificationFactory newIssuesNotificationFactory;
+  private Map<String, Component> componentsByDbKey = new HashMap<>();
```
sns-seb: this uselessly allocates a HashMap in memory.
In addition, if there is no leaf (unlikely, but possible), the map will be computed for each call to `sendIssueChangeNotification`.

That's 2 reasons you should have `componentsByDbKey` assigned null when declared and test on `null` rather than empty in `getComponentKey`

sns-seb: since @julienlancelot is off this week, I'm pinging you @dbmeneses  :)

dbmeneses: @sns-seb fixed in dm/quality

PR #2469, opened 2017-09-01
Silence logs from Hazelcast and Logback

ehartmann:


server/sonar-process-monitor/src/main/java/org/sonar/application/cluster/AppStateClusterImpl.java
```diff
@@ -87,7 +88,12 @@ public void reset() {
 
   @Override
   public void close() {
-    hazelcastCluster.close();
+    try {
+      hazelcastCluster.close();
```
sns-seb: I believe it would be more appropriate to catch this error in class `HazelcastCluster` for two reasons:
1. would be able to put try/catch around `hzInstance.shutdown()` (which I believe it the method actually raising the exception in SQ's code)
2. design and import-wize AppStateClusterImpl does not deal with the hazelcast framework. It should stay that way

PR #2476, opened 2017-09-04
Do not trigger twice the notification thread

simonbrandhof:


julienlancelot:
Sorry, but this change doesn't work : notification is no more sent at startup : see failing test in QA BuiltInQualityProfilesNotificationTest.send_mail_if_quality_profile_is_updated

simonbrandhof:
Yes I saw. There's something strange. I will have a look.

PR #2484, opened 2017-09-06
SONAR-9701 Don't publish coverage, scm and source of unchanged files

henryju:


sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/filesystem/InputComponentStore.java
```diff
@@ -80,7 +83,7 @@ public InputComponentStore(DefaultInputModule root, AnalysisMode mode) {
     return inputFileCache.values().stream()
       .map(f -> (DefaultInputFile) f)
       .filter(DefaultInputFile::isPublished)
-      .filter(f -> !mode.isIncremental() || f.status() != Status.SAME)::iterator;
+      .filter(f -> (!mode.isIncremental() && !branchConfiguration.isShortLivingBranch()) || f.status() != Status.SAME)::iterator;
```
dbmeneses: if status is SAME, shouldn't we filter it regardless of the branch type (currently it should only happen for short living branches, right?)

henryju: * "Normal analysis": write all components
* incremental / short living branches: write only changed components

PR #2490, opened 2017-09-08
SONAR-9181 WS bulk_apply_template accepts the same parameters as projects/search

teryk:


server/sonar-server/src/main/java/org/sonar/server/ws/WsParameterBuilder.java
```diff
@@ -45,6 +45,13 @@ private WsParameterBuilder() {
       .setPossibleValues(getRootQualifiers(context.getResourceTypes()));
   }
 
+  public static WebService.NewParam createRootQualifiersParameter(WebService.NewAction action, QualifierParameterContext context) {
```
julienlancelot: As discussed, it would have been best to refactor WsParameterBuilder.java to use injection for I18n and ResourceTypes resourceTypes instead of using this strange context object.

server/sonar-server/src/main/java/org/sonar/server/ws/WsParameterBuilder.java
```diff
@@ -45,6 +45,13 @@ private WsParameterBuilder() {
...
+    return action.createParam(PARAM_QUALIFIERS)
+      .setDescription("Comma-separated list of component qualifiers. Filter the results with the specified qualifiers. " +
+        "Possible values are:" + buildRootQualifiersDescription(context))
+      .setPossibleValues(getRootQualifiers(context.getResourceTypes()));
```
julienlancelot: As api/projects/search is using a default value 'TRK' for qualifiers, why not using the same here ?
For consistency I think it would be better

server/sonar-server/src/test/java/org/sonar/server/permission/ws/template/BulkApplyTemplateActionTest.java
```diff
@@ -104,7 +106,10 @@ public void bulk_apply_template_by_template_uuid() throws Exception {
     ComponentDto view = db.components().insertView(organization);
     loginAsAdmin(organization);
 
-    newRequest().setParam(PARAM_TEMPLATE_ID, template1.getUuid()).execute();
+    newRequest()
+      .setParam(PARAM_TEMPLATE_ID, template1.getUuid())
+      .setParam(PARAM_QUALIFIERS, String.join(",", Qualifiers.PROJECT, Qualifiers.VIEW))
```
julienlancelot: If default value is added to the qualifiers parameter, this line becomes useless

server/sonar-server/src/main/java/org/sonar/server/permission/ws/template/BulkApplyTemplateAction.java
```diff
@@ -91,6 +104,40 @@ public void define(WebService.NewController context) {
...
+
+    action
+      .createParam(PARAM_PROJECT_IDS)
+      .setDescription("Comma-separated list of project ids")
+      // to map parameters of api/projects/search
+      .setSince("6.6")
+      .setDeprecatedSince("6.6")
+      .setExampleValue(String.join(",", UUID_EXAMPLE_01, UUID_EXAMPLE_02));
+
+    action.createParam(PARAM_VISIBILITY)
```
julienlancelot: Is it correct that this parameter is not used ?

server/sonar-server/src/main/java/org/sonar/server/permission/ws/template/BulkApplyTemplateAction.java
```diff
@@ -91,6 +104,40 @@ public void define(WebService.NewController context) {
...
     createTemplateParameters(action);
+
+    action
+      .createParam(PARAM_PROJECTS)
+      .setDescription("Comma-separated list of project keys")
+      .setSince("6.6")
+      .setExampleValue(String.join(",", KEY_PROJECT_EXAMPLE_001, KEY_PROJECT_EXAMPLE_002));
+
+    action
+      .createParam(PARAM_PROJECT_IDS)
```
julienlancelot: This parameter should not be added, it didn't exist before

server/sonar-server/src/main/java/org/sonar/server/permission/ws/template/BulkApplyTemplateAction.java
julienlancelot: If it's not used, I wouldn't add it

server/sonar-server/src/main/java/org/sonar/server/ws/WsParameterBuilder.java
teryk: indeed, I haven't taken the time to do it

server/sonar-server/src/main/java/org/sonar/server/ws/WsParameterBuilder.java
teryk: it's done in the BulkApplyTemplate action, and this class is used along the whole permission domain, so I don't want to change that

server/sonar-server/src/test/java/org/sonar/server/permission/ws/template/BulkApplyTemplateActionTest.java
teryk: I don't understand. The default value is set if there is no value provided

server/sonar-server/src/main/java/org/sonar/server/permission/ws/template/BulkApplyTemplateAction.java
teryk: it existed in the `api/projects/search` WS

server/sonar-server/src/main/java/org/sonar/server/permission/ws/template/BulkApplyTemplateAction.java
teryk: I haven't checked yet, but to be consistent either it's available in api/projects/search and this WS or not at all

server/sonar-server/src/main/java/org/sonar/server/ws/WsParameterBuilder.java
julienlancelot: No pb

server/sonar-server/src/main/java/org/sonar/server/ws/WsParameterBuilder.java
julienlancelot: Then I would propose to add the default value in BulkApplyTemplateAction

server/sonar-server/src/test/java/org/sonar/server/permission/ws/template/BulkApplyTemplateActionTest.java
julienlancelot: So this line can be removed

server/sonar-server/src/main/java/org/sonar/server/permission/ws/template/BulkApplyTemplateAction.java
julienlancelot: Let's discuss this

PR #2491, opened 2017-09-08
SONAR−9772 fail when changing mandatory JVM options through properties

sns-seb:


server/sonar-process/src/main/java/org/sonar/process/jmvoptions/JvmOptions.java
```diff
@@ -21,31 +21,90 @@
...
+      List<String> jvmOptions = Arrays.stream(value.split(" (?=-)")).map(String::trim).collect(Collectors.toList());
+      checkOptionFormat(propertyName, jvmOptions);
+      checkMandatoryOptionOverwrite(propertyName, jvmOptions);
+      options.addAll(jvmOptions);
     }
 
     return castThis();
   }
 
+  private void checkOptionFormat(String propertyName, List<String> jvmOptionsFromProperty) {
```
SonarTech: MAJOR Make "checkOptionFormat" a "static" method. [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS2325)

SonarTech:
* ![MAJOR][MAJOR] 1 major

[MAJOR]: https://sonarsource.github.io/sonar-github/severity-major.png 'Severity: MAJOR'

server/sonar-process/src/main/java/org/sonar/process/jmvoptions/JvmOptions.java
```diff
@@ -21,31 +21,90 @@
...
+        return true;
+      }).forEach(e -> {
+        String key = e.getKey().trim();
+        String value = e.getValue().trim();
+        mandatoryOptions.put(key, value);
+        add(key + value);
+      });
   }
 
   public T addFromMandatoryProperty(Props props, String propertyName) {
```
ehartmann: The term `mandatory` looks strange to me because by looking at the callers we use it with non mandatory property such as `.addFromMandatoryProperty(props, ProcessProperties.WEB_JAVA_ADDITIONAL_OPTS)`.
I would prefer to rename this method with addFromProperty.

server/sonar-process/src/main/java/org/sonar/process/jmvoptions/JvmOptions.java
```diff
@@ -55,14 +114,37 @@ public T addFromMandatoryProperty(Props props, String propertyName) {
...
+        "a JVM option can't overwrite mandatory JVM options. %s overwrites %s",
+        value,
+        overriddenMandatoryOption.getKey() + overriddenMandatoryOption.getValue()));
+    }
+  }
+
+  @CheckForNull
+  private Map.Entry<String, String> mandatoryOptionFor(String jvmOption) {
+    return mandatoryOptions.entrySet().stream()
+      .filter(s -> jvmOption.startsWith(s.getKey()) && !jvmOption.equals(s.getKey() + s.getValue()))
```
ehartmann: From my point of view, we should display a message when a user is setting the property in both place even with same value.
And I think we may have corner case here, let's image that the JVM allows two options :
- -a
- -ab

We are not allowing to add `-ab` to optional parameters if `-a` is a mandatory parameter.
So maybe the solution is to extract the key from the configuration property and compare the keys without the value.

server/sonar-process/src/main/java/org/sonar/process/jmvoptions/JvmOptions.java
sns-seb: it's mandatory but not necessarily specified by the user thanks to a default value added at runtime.
This default value is not known to this class.

server/sonar-process/src/main/java/org/sonar/process/jmvoptions/JvmOptions.java
sns-seb: This is theoretical. I considered it, looked a bit into it and dismissed the case.

As per the documentation provided by Oracle, there is no such case that I could find.
In addition, as for the option we actually define in our code (which is a restricted set of the available options), there is no such case for sure.

PR #2492, opened 2017-09-08
.level label localazed size css fix

asosnoviy:
Fixing problem with localazed string in level label.  

Before:  

image  

After:   

image

In En local size is not changed:

image


stas-vilchik:
Looks good! 👍

PR #2515, opened 2017-09-15
SONAR-9621 Always load duplication blocks from report when mode != in…

henryju:
…cremental


henryju:
Before merging, I'm waiting to know if this is the correct ticket number.

PR #2516, opened 2017-09-15
SONAR-9813 allow to change portfolio visibility

stas-vilchik:


stas-vilchik:
Good suggestions! And they all will be done https://jira.sonarsource.com/browse/GOV-235 😉

PR #2518, opened 2017-09-18
SONAR-9806 fail restarting attempts on cluster nodes

bartfastiel:


tests/src/test/java/org/sonarqube/tests/cluster/ClusterTest.java
```diff
@@ -245,6 +246,25 @@ public void restarting_all_application_nodes_elects_a_new_startup_leader() throw
   }
 
   @Test
+  public void restart_action_is_not_allowed_for_cluster_nodes() throws Exception {
```
bartfastiel: Probably this feature is not worth an IT, I'll probably better remove this before merging.

simonbrandhof: nope, please keep it !

PR #2521, opened 2017-09-18
SONAR-9807 Redesign standalone system page

gregaubert:


stas-vilchik:
Two small details:

- the chevron icon is slightly misaligned (macOS, Chrome):
image

- could be nice to make these kind of indicators consistent:
image

image

PR #2528, opened 2017-09-19
SONAR-9841 display tooltip on language distribution chart

stas-vilchik:


stas-vilchik:
@gregaubert the tooltip is a bit misplaced, but I plan to fix it in https://jira.sonarsource.com/browse/SONAR-9697

PR #2530, opened 2017-09-19
Governance hardening 2

stas-vilchik:


stas-vilchik:
@gregaubert about SONAR-9841: the tooltip is a bit misplaced, but I plan to fix it in https://jira.sonarsource.com/browse/SONAR-9697

server/sonar-web/src/main/js/apps/projectsManagement/Search.tsx
```diff
@@ -223,17 +223,19 @@ export default class Search extends React.PureComponent<Props, State> {
...
                   {translate('permission_templates.bulk_apply_permission_template')}
                 </button>
-                <button
-                  className="js-delete button-red"
-                  disabled={this.props.total === 0}
-                  onClick={this.handleDeleteClick}>
-                  {translate('delete')}
-                </button>
+                {this.props.qualifiers === 'TRK' && (
+                  <button
```
gregaubert: Adding an IT to check that the deletion is not there for portfolios could be nice

server/sonar-web/package.json
```diff
@@ -46,6 +46,8 @@
   },
   "devDependencies": {
     "@types/classnames": "2.2.0",
+    "@types/d3-array": "1.2.1",
+    "@types/d3-scale": "1.0.10",
```
gregaubert: We already have different versions of those in the dependencies, is it on purpose ?

server/sonar-web/src/main/js/components/charts/LanguageDistribution.tsx
```diff
@@ -28,34 +28,41 @@ interface Props {
...
+      yTooltips={yTooltips}
       yValues={yValues}
-      barsWidth={10}
-      height={data.length * 25}
-      padding={[0, 60, 0, 80]}
+      width={props.width}
     />
   );
 
   function getLanguageName(langKey: string) {
```
gregaubert: Didn't you want to put those two functions after the component ? It's currently inside.

stas-vilchik: `getLanguageName` needs access to the component props, so it's better to keep it inside, but `cutLanguageName` can be moved outside

server/sonar-web/src/main/js/apps/projectsManagement/Search.tsx
stas-vilchik: ok, I will add it, but in sonar-governance

PR #2531, opened 2017-09-20
fix quality flaws

bartfastiel:
This PR tries to fix the last two remaining issues, that are currently making the SonarQube quality gate red.


bartfastiel:
These issues have been fixed as part of other commits.

PR #2540, opened 2017-09-22
some hardening

stas-vilchik:


server/sonar-web/src/main/js/api/components.ts
```diff
@@ -65,6 +65,10 @@ export function deleteProject(project: string): Promise<void> {
   return post('/api/projects/delete', { project });
 }
 
+export function deletePortfolio(portfolio: string): Promise<void> {
+  return post('/api/views/delete', { key: portfolio });
```
gregaubert: I'd add `throwGlobalError` here (same for delete project)

PR #2544, opened 2017-09-22
SONAR-9811 change es data dir to data/es5, delete data/es during startup

bartfastiel:


server/sonar-main/src/main/java/org/sonar/application/process/ProcessLauncherImpl.java
```diff
@@ -88,6 +90,22 @@ public ProcessMonitor launch(EsCommand esCommand) {
...
+    esFileSystem.getOutdatedDataDirectories().forEach(outdatedDir -> {
+      if (outdatedDir.exists()) {
+        LOG.info("Deleting outdated search index data directory {}", outdatedDir.getAbsolutePath());
+        try {
+          FileUtils.deleteDirectory(outdatedDir);
+        } catch (IOException e) {
+          LOG.info("Failed to delete outdated search index data directory " + outdatedDir.getAbsolutePath(), e);
+        }
+      } else {
+        LOG.trace("No outdated search index data directory found at {}", outdatedDir.getAbsolutePath());
```
bartfastiel: This logging is actually challengeable. Should I drop it?

simonbrandhof: Yes I think that it should be dropped. It's the regular case.

server/sonar-main/src/main/java/org/sonar/application/process/ProcessLauncherImpl.java
```diff
@@ -88,6 +90,20 @@ public ProcessMonitor launch(EsCommand esCommand) {
     }
   }
 
+  private static void cleanupOutdatedEsData(EsCommand esCommand) {
+    EsFileSystem esFileSystem = esCommand.getFileSystem();
+    esFileSystem.getOutdatedDataDirectories().forEach(outdatedDir -> {
+      if (outdatedDir.exists()) {
+        LOG.info("Deleting outdated search index data directory {}", outdatedDir.getAbsolutePath());
```
sns-seb: @bartfastiel there is an inconsistency here.

you added support for multiple outdated data dir (which is overkill since there is only one such directory) but the log will work only for a single such directory, always the search directory.

I'd remove the useless collection and avoid the problem

bartfastiel: I created PR #2597 for to avoid misunderstandings with the variable names.

PR #2575, opened 2017-09-27
LICENSE-64 add internal CE extension point CeTaskInitialization

simonbrandhof:


server/sonar-ce/src/main/java/org/sonar/ce/taskprocessor/CeWorkerImpl.java
```diff
@@ -128,6 +130,7 @@ private void executeTask(CeTask task) {
       // TODO delegate the message to the related task processor, according to task type
       Optional<CeTaskProcessor> taskProcessor = taskProcessorRepository.getForCeTask(task);
       if (taskProcessor.isPresent()) {
+        taskInitializations.onInit(task);
```
sns-seb: @simonbrandhof if task processing breaks (for any reason, license or whatever) at this stage of processing, webhooks won't be triggered

simonbrandhof:
As suggested a different approach will be used. Thanks @sns-seb for the good feedback.

PR #2576, opened 2017-09-27
SONAR-9886 Load issues not closed or open for a component

janos-ss:


server/sonar-db-dao/src/main/resources/org/sonar/db/issue/IssueMapper.xml
```diff
@@ -213,6 +213,18 @@
     </foreach>
   </select>
 
+  <select id="selectResolvedOrConfirmedByComponentUuid" parameterType="String" resultType="Issue">
+    select
+    <include refid="issueColumns"/>
+    from issues i
+      inner join rules r on r.id=i.rule_id
+      inner join projects p on p.uuid=i.component_uuid
+      inner join projects root on root.uuid=i.project_uuid
```
dbmeneses: We probably don't need all this, so perhaps we can create a lighter project and get rid of all the joins

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/issue/ResolvedShortBranchIssuesFactory.java
```diff
@@ -0,0 +1,31 @@
...
+  public ResolvedShortBranchIssuesFactory(ShortBranchComponentsWithIssues shortBranchComponentsWithIssues, DbClient dbClient) {
+    this.shortBranchComponentsWithIssues = shortBranchComponentsWithIssues;
+    this.dbClient = dbClient;
+  }
+
+  public Collection<DefaultIssue> create(Component component) {
+    try (DbSession session = dbClient.openSession(false)) {
+      return shortBranchComponentsWithIssues.getUuids(component.getKey())
+        .stream()
+        .flatMap(uuid -> dbClient.issueDao().selectResolvedOrConfirmedByComponentUuid(session, uuid).stream())
```
dbmeneses: Would it be a good idea to pass all uuids into the dao at once?

server/sonar-db-dao/src/main/resources/org/sonar/db/issue/IssueMapper.xml
janos-ss: Yeah, this was lazy of me to try to reuse the existing mapper. I'll create a new light-weight object to map to, with only the fields we need. Should perform better.

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/issue/ResolvedShortBranchIssuesFactory.java
janos-ss: There will be multiple uuids only when issues appear on multiple branches. I would have liked to assume that this will be rare in practice, but maybe that's too naive. I'll change this.

simonbrandhof:
@janos-ss @dbmeneses https://jira.sonarsource.com/browse/SONAR-9886 is closed, so this PR can be closed too ?

PR #2579, opened 2017-09-27
Add Compute Engine extension point for task initialization

simonbrandhof:


server/sonar-server/src/main/java/org/sonar/server/plugins/PluginDownloader.java
```diff
@@ -117,7 +115,10 @@ public boolean hasDownloads() {
    * @return the list of download plugins as {@link PluginInfo} instances
    */
   public Collection<PluginInfo> getDownloadedPlugins() {
-    return newArrayList(transform(listPlugins(this.downloadDir), jarToPluginInfo()));
+    return listPlugins(this.downloadDir)
+      .stream()
+      .map(PluginInfo::create)
+      .collect(Collectors.toList());
```
sns-seb: Using MoreCollectors will (use Guava under the hood) return an immutable list

server/sonar-server/src/main/java/org/sonar/server/plugins/ServerPluginRepository.java
```diff
@@ -346,7 +344,10 @@ private void appendDependentPluginKeys(String pluginKey, Set<String> appendTo) {
    * @return the list of plugins to be uninstalled as {@link PluginInfo} instances
    */
   public Collection<PluginInfo> getUninstalledPlugins() {
-    return newArrayList(transform(listJarFiles(uninstalledPluginsDir()), jarToPluginInfo()));
+    return listJarFiles(uninstalledPluginsDir())
+      .stream()
+      .map(PluginInfo::create)
+      .collect(Collectors.toList());
```
sns-seb: same suggestion to use MoreCollectors here

server/sonar-server/src/main/java/org/sonar/server/computation/task/step/ExecuteTaskInitExtensionsStep.java
```diff
@@ -0,0 +1,48 @@
...
+@ComputeEngineSide
+public class ExecuteTaskInitExtensionsStep implements ComputationStep {
+
+  private final TaskInitExtension[] extensions;
+
+  public ExecuteTaskInitExtensionsStep(TaskInitExtension[] extensions) {
+    this.extensions = extensions;
+  }
+
+  public ExecuteTaskInitExtensionsStep() {
```
sns-seb: could be useful for new comers to add comment that this constructor is required by Pico when there is no TaskInitExtension

server/sonar-server/src/main/java/org/sonar/server/computation/task/step/TaskInitExtension.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Extension point that is called during processing of a task
+ * by {@link ExecuteTaskInitExtensionsStep}.
+ */
+@ComputeEngineSide
+@ExtensionPoint
+public interface TaskInitExtension {
+
+  void onInit();
```
sns-seb: could be wise documenting whether exception are supported or not

server/sonar-server/src/test/java/org/sonar/server/computation/task/step/ExecuteTaskInitExtensionsStepTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Test
+  public void execute_extensions() {
+    TaskInitExtension ext1 = mock(TaskInitExtension.class);
+    TaskInitExtension ext2 = mock(TaskInitExtension.class);
+
+    ExecuteTaskInitExtensionsStep underTest = new ExecuteTaskInitExtensionsStep(new TaskInitExtension[] {ext1, ext2});
+    underTest.execute();
+
+    verify(ext1).onInit();
```
sns-seb: I'd would use Mockito.inOrder(ext1, ext2) here, to verify that TaskInitExtensions are executed in order they are passed to the constructor

server/sonar-server/src/test/java/org/sonar/server/computation/task/step/ExecuteTaskInitExtensionsStepTest.java
```diff
@@ -0,0 +1,77 @@
...
+    TaskInitExtension ext1 = mock(TaskInitExtension.class);
+    doThrow(new IllegalStateException("BOOM")).when(ext1).onInit();
+    TaskInitExtension ext2 = mock(TaskInitExtension.class);
+
+    ExecuteTaskInitExtensionsStep underTest = new ExecuteTaskInitExtensionsStep(new TaskInitExtension[] {ext1, ext2});
+
+    expectedException.expect(IllegalStateException.class);
+    expectedException.expectMessage("BOOM");
+
+    underTest.execute();
```
sns-seb: I would:
1. use 3 TaskInitExtension rather than 2
2. make the second one in the list throw the exception
3. try and catch the exception rather than use `ExpectedException` here
4. to be able to verify that first and second TaskInitExtension are called but not the third one

server/sonar-server/src/main/java/org/sonar/server/computation/task/step/TaskInitExtension.java
```diff
@@ -0,0 +1,35 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.computation.task.step;
+
+import org.sonar.api.ExtensionPoint;
+import org.sonar.api.ce.ComputeEngineSide;
+
+/**
+ * Extension point that is called during processing of a task
+ * by {@link ExecuteTaskInitExtensionsStep}.
```
sns-seb: I would document that `TaskInitExtension` instances are executed in no specific order but one after the other and that failure of one of them will make the others not to be executed.

server/sonar-server/src/main/java/org/sonar/server/computation/task/step/TaskInitExtension.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+@ComputeEngineSide
+@ExtensionPoint
```
sns-seb: I'm not sure if the `@ExtensionPoint` annotation is correct here.

implementation of this class should not be returned as extension of the plugin but as classes to be instantiated in the container of the report processing task using a `ReportAnalysisComponentProvider`

This way, instances of implementations of `TaskInitExtension` can be stateful and get injected with components during report processing (such as `CeTask` or whatever)

server/sonar-server/src/main/java/org/sonar/server/computation/task/step/TypedException.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.computation.task.step;
+
+import static java.util.Objects.requireNonNull;
+
+public class TypedException extends RuntimeException {
```
sns-seb: javadoc to explain why and how to use this exception?

server/sonar-server/src/test/java/org/sonar/server/computation/task/step/ExecuteTaskInitExtensionsStepTest.java
simonbrandhof: I don't think it's worth to test that because order is not assured at time when injecting plugin components into pico.

sns-seb: well, this UT is not testing pico, is it?

If you consider just this class, which ever the order of the extension provided to it:
1. it executes the extension in order
2. it fails at the first one failing

As far as your tests go, I can't tell that this is how it works.
The class could as well implement a mechanism to call all extensions in isolation or with an internal ordering and your tests would succeed all the same while the contract of the class has changed quite a lot.

Point 2 is a relevant information from the point of view of the dev writing a `TaskInitExtension`.

PR #2584, opened 2017-09-28
SONAR-9756 update branch management page

stas-vilchik:

PR #2588, opened 2017-09-28
Background task notification

gregaubert:
@stas-vilchik If you want to test the licencing errors you can use the same procedure as the last PR about the license plugin. Just know that if you want the analysis to fail you have to install the commercial plugin on your instance, for that you must use the fake cobol plugin generated from the IT's of the license plugin.


server/sonar-web/src/main/js/api/ce.ts
```diff
@@ -20,6 +20,31 @@
 import { getJSON, post, RequestData } from '../helpers/request';
 import throwGlobalError from '../app/utils/throwGlobalError';
 
+export interface PendingTask {
+  componentId: string;
```
stas-vilchik: I suggest to pretend that we don't have `componentId` field 😄 It should be deprecated & removed soon

server/sonar-web/src/main/js/app/components/nav/component/ComponentNavBgTaskNotif.tsx
```diff
@@ -0,0 +1,106 @@
...
+      component.configuration != undefined && component.configuration.showBackgroundTasks;
+    const url = getComponentBackgroundTaskUrl(component.key);
+    if (isInProgress) {
+      return (
+        <NavBarNotif className="alert alert-info">
+          <i className="spinner spacer-right text-bottom" />
+          <span
+            dangerouslySetInnerHTML={{
+              __html: canSeeBackgroundTasks
+                ? translateWithParameters('component_navigation.status.in_progress.admin', url)
```
stas-vilchik: I suggest using `FormattedMessage` from `react-intl` to put react components (`<Link />` in this case) inside the messages.

For example see https://github.com/SonarSource/sonarqube/pull/2584/files#diff-6a888b67f80f7a034a299b5351bfbdceR87

server/sonar-web/src/main/js/app/components/nav/component/ComponentNavBgTaskNotif.tsx
```diff
@@ -0,0 +1,106 @@
...
+            )}
+          </NavBarNotif>
+        );
+      } else {
+        return (
+          <NavBarNotif className="alert alert-danger">
+            <span
+              dangerouslySetInnerHTML={{
+                __html: canSeeBackgroundTasks
+                  ? translateWithParameters('component_navigation.status.failed.admin', url)
```
stas-vilchik: blue link in red alert is super ugly

server/sonar-web/src/main/js/api/ce.ts
gregaubert: Ah yeah good idea!

server/sonar-web/src/main/js/app/components/nav/component/ComponentNavBgTaskNotif.tsx
gregaubert: Common it's not "super ugly", maybe a "bit ugly", but that's a link... Trust me, it's way better than a blue button! And we have the same colors on issues page.
I don't think we should change the color of links, do you suggest to maybe change a bit the red background ?

stas-vilchik: I'd suggest to use red font for link

gregaubert: Ok let me talk to Laura about this

gregaubert: Ok let me talk to Laura about this

gregaubert: The problem is that red links are for dangerous/irreversible actions.

gregaubert:
@stas-vilchik I added the fix for the sidebar and fixed your feedback expect the color link.

PR #2590, opened 2017-09-29
Hardening phase SonarQube 6 - Daniel 2017-09-29

bartfastiel:
…escription


bartfastiel:
As discussed with @julienlancelot and @teryk , we are looking for a more powerful way of providing the information about API permission later (after the LTS).

PR #2600, opened 2017-10-02
SONAR-8828 Make sure all messages can be localized

stas-vilchik:
💣 💣 💣

### DIFF
--- a/server/sonar-web/src/main/js/app/components/nav/settings/SettingsNav.js
+++ b/server/sonar-web/src/main/js/app/components/nav/settings/SettingsNav.js
@@ -116,7 +116,7 @@ class SettingsNav extends React.PureComponent {
               </li>
               <li>
                 <IndexLink to="/admin/custom_metrics" activeClassName="active">
-                  Custom Metrics
+                  {translate('custom_metrics.page')}
                 </IndexLink>
               </li>
               {this.props.extensions.map(this.renderExtension)}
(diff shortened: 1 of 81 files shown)


nixel2007:
Hello.
We can build this PR, try to translate new keys localy (to russian) and give you feedback, if you give us a few days for that.

stas-vilchik:
@nixel2007 would be great!

However I'd like to merge this PR as soon as possible to avoid future merge conflicts. So I'll open a new PR if you find missing places.

nixel2007:
@stas-vilchik okey, sure.

nixel2007:
@stas-vilchik we've made a draft of translation and it looks very good so far! even "Show measures" page (from file) is fully translated.

There are some css problems cause of short english terms (like "bugs" term).
image

I'll ask our web-programmer to take a look at it, but it would be great if SonarSource will keep in mind the problem of the size of words in other languages.

The last part of untranslated strings was found in admin area - configuration of project or server itself.
image

All keys aren't localizable. May I ask you to take a look?

stas-vilchik:
@nixel2007 thanks for your investigation!

> There are some css problems cause of short english terms (like "bugs" term).

Can I ask you to report those problems case by case, since it's quite hard to anticipate everything :) And I'll see what we can do for the problem you've reported (with "Bugs").

> The last part of untranslated strings was found in admin area - configuration of project or server itself.

We don't put localization messages for every possible setting, because we use the default name and description. However plugins can override them by providing two messages: `property.<setting key>.name` and `property.<setting key>.description`. So even if there are not such strings in the default localization bundle, you can still add them in your own.

nixel2007:
> We don't put localization messages for every possible setting, because we use the default name and description. However plugins can override them by providing two messages: property.<setting key>.name and property.<setting key>.description. So even if there are not such strings in the default localization bundle, you can still add them in your own.

Oh, thanks a lot! I'll try it.

PR #2608, opened 2017-10-02
Hardening 02.10.2017

gregaubert:


sonar-core/src/main/resources/org/sonar/l10n/core.properties
```diff
@@ -2460,7 +2461,7 @@ footer.web_api=Web API
 #------------------------------------------------------------------------------
 onboarding.header=Welcome to SonarQube!
 onboarding.header.sonarcloud=Welcome to SonarCloud!
-onboarding.header.description=Let's analyze a new project.
+onboarding.header.description=Want to quickly analyze a first project? Follow these 3 easy steps.
```
stas-vilchik: why 3 steps? For SonarQube on-premise, there are only two: token and project.

server/sonar-web/src/main/js/apps/system/utils.ts
```diff
@@ -110,9 +110,17 @@ export function getNodeName(nodeInfo: NodeInfo): string {
   return nodeInfo['Name'];
 }
 
+function getSystemData(sysInfoData: SysInfo): SysValueObject {
+  const statData: SysValueObject = {};
+  if (sysInfoData['Statistics']) {
+    statData['Lines of Code'] = (sysInfoData['Statistics'] as SysValueObject)['ncloc'];
```
stas-vilchik: I wonder if it's easy to format this number.

gregaubert: It is, at first I didn't want to because all other numbers won't be formatted

PR #2614, opened 2017-10-03
Teryk's hardening – 10-03

teryk:


gregaubert:
@bartfastiel I fixed the categories of the new coverage metrics

PR #2619, opened 2017-10-03
Web Hardening Oct 3

stas-vilchik:


server/sonar-web/src/main/js/apps/projectQualityGate/App.tsx
```diff
@@ -82,7 +82,7 @@ export default class App extends React.PureComponent<Props> {
     );
   }
 
-  handleChangeGate = (oldId: string | undefined, newId: string | undefined) => {
+  handleChangeGate = (oldId: number | undefined, newId: number | undefined) => {
```
gregaubert: I think you can use `(oldId?: number, newId?: number)`

server/sonar-web/src/main/js/apps/account/organizations/OrganizationCard.js
```diff
@@ -52,11 +53,22 @@ export default function OrganizationCard(props /*: Props */) {
...
         <div className="account-project-description">{organization.description}</div>
       )}
+
+      <div className="account-project-key">
+        <span className="little-spacer-right">
+          {translate('key')}
+          {':'}
+        </span>
+        <input
+          onClick={event => event.currentTarget.select()}
```
gregaubert: Maybe extract that function, but I don't think it's really important on that page

server/sonar-web/src/main/js/apps/projectQualityGate/App.tsx
stas-vilchik: this is tricky: `oldId?: number` means the parameter can be omitted, `oldId: number | undefined` means the parameter must be provided, but can have value `undefined`

server/sonar-web/src/main/js/apps/account/organizations/OrganizationCard.js
stas-vilchik: since `input` is not a pure component, it's fine to pass new function every time :) I hope :)

PR #2623, opened 2017-10-04
Allow to navigate back to the issue from the last issue locations

gregaubert:


server/sonar-web/src/main/js/apps/issues/actions.js
```diff
@@ -21,12 +21,13 @@
...
 export const enableLocationsNavigator = (state /*: State */) => {
-  const { openIssue } = state;
+  const { openIssue, selectedLocationIndex } = state;
   if (openIssue && (openIssue.secondaryLocations.length > 0 || openIssue.flows.length > 0)) {
     return {
       locationsNavigator: true,
       selectedFlowIndex: state.selectedFlowIndex || (openIssue.flows.length > 0 ? 0 : null),
-      selectedLocationIndex: state.selectedLocationIndex || 0
+      selectedLocationIndex:
+        !selectedLocationIndex || selectedLocationIndex < 0 ? 0 : selectedLocationIndex
```
stas-vilchik: could you put comment explaining when `selectedLocationIndex < 0`, otherwise it'll be very hard to remember :)

gregaubert: Yeah true, it's tricky. But I still I think it was the simplest way to do it. I'll add a few more comments to make it more easy to understand :)

PR #2628, opened 2017-10-04
SONAR-9445 Permission 'Administer Quality Gates' is mandatory and sufficient to delete a QG condition

teryk:


server/sonar-server/src/main/java/org/sonar/server/qualitygate/ws/DeleteConditionAction.java
```diff
@@ -22,36 +22,52 @@
...
       .setDescription("Condition ID")
       .setExampleValue("2");
   }
 
   @Override
   public void handle(Request request, Response response) {
-    qualityGates.deleteCondition(QualityGatesWs.parseId(request, QualityGatesWsParameters.PARAM_ID));
-    response.noContent();
+    try (DbSession dbSession = dbClient.openSession(false)) {
+      long conditionId = request.mandatoryParamAsLong(PARAM_ID);
```
bartfastiel: Should be done before opening a db session - for to fail fast.

PR #2630, opened 2017-10-05
SONAR-9263 log warning once when OAuth URL is http but not https

simonbrandhof:


server/sonar-server/src/main/java/org/sonar/server/authentication/LogOAuthWarning.java
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Used by default by picocontainer when no OAuth2IdentityProvider are present
+   */
+  public LogOAuthWarning(Server server) {
+    this(server, new OAuth2IdentityProvider[0]);
+  }
+
+  @Override
+  public void start() {
+    if (providers.length > 0) {
```
julienlancelot: To improve readability, you can inverse the condition : 
```
if (providers.length == 0) {
  return;
}
```

server/sonar-server/src/main/java/org/sonar/server/authentication/LogOAuthWarning.java
```diff
@@ -0,0 +1,60 @@
...
+      String publicRootUrl = server.getPublicRootUrl();
+      if (StringUtils.startsWithIgnoreCase(publicRootUrl, "http:")) {
```
julienlancelot: Same here

server/sonar-server/src/main/java/org/sonar/server/authentication/LogOAuthWarning.java
simonbrandhof: ok

PR #2635, opened 2017-10-09
Hardening for 6.7

gregaubert:


server/sonar-web/src/main/js/apps/sessions/components/LoginFormContainer.tsx
```diff
@@ -0,0 +1,99 @@
...
+  componentDidMount() {
+    this.mounted = true;
+    Promise.all([
+      getIdentityProviders(),
+      this.props.fetchAppState()
+    ]).then(([identityProvidersResponse, appState]) => {
+      if (this.mounted) {
+        this.setState({
+          onSonarCloud:
+            appState.settings != undefined &&
```
SonarTech: MAJOR != should be !== [rule](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=typescript%3AS1440)

SonarTech:
* ![MAJOR][MAJOR] 1 major

[MAJOR]: https://sonarsource.github.io/sonar-github/severity-major.png 'Severity: MAJOR'

server/sonar-web/src/main/js/app/components/SimpleContainer.tsx
```diff
@@ -17,21 +17,16 @@
...
-type Props = {
-  children?: React.Element<*> | Array<React.Element<*>>,
-  hideLoggedInInfo?: boolean
-};
-*/
-
-export default class SimpleContainer extends React.PureComponent {
-  /*:: props: Props; */
+interface Props {
+  children?: React.ReactElement<any>;
```
stas-vilchik: maybe `React.ReactNode` ?

server/sonar-web/src/main/js/apps/sessions/components/LoginFormContainer.tsx
```diff
@@ -51,14 +59,12 @@ class LoginFormContainer extends React.PureComponent {
   handleSuccessfulLogin = () => {
     const { location } = this.props;
     const queryReturnTo = location.query['return_to'];
-    const returnTo = queryReturnTo ? `${queryReturnTo}${location.hash}` : `${window.baseUrl}/`;
-    window.location = returnTo;
+    const returnTo = queryReturnTo ? `${queryReturnTo}${location.hash}` : `${getBaseUrl()}/`;
+    this.context.router.push(returnTo);
```
stas-vilchik: two problems here:

- first, as you're redirecting with router and both `queryReturnTo` and `getBaseUrl()` are paths with `baseUrl` included, this will lead to the double web context problem (`/sonarqube/sonarqube`)

- second, it's better to fully reload the page after login to re-fetch all data

server/sonar-web/src/main/js/apps/sessions/components/Logout.tsx
```diff
@@ -17,25 +17,27 @@
...
-      .then(() => {
+    this.props.doLogout().then(
+      () => {
         RecentHistory.clear();
-        window.location = window.baseUrl + '/';
-      })
-      .catch(() => {
-        /* do nothing */
-      });
+        (window as any).location = getBaseUrl() + '/';
```
stas-vilchik: according to the answer [on SO](https://stackoverflow.com/questions/13106950/set-window-location-with-typescript), you can write `window.location.href = ...`

server/sonar-web/src/main/js/apps/sessions/components/LoginFormContainer.tsx
```diff
@@ -22,32 +22,41 @@ import * as PropTypes from 'prop-types';
...
   static contextTypes = {
     router: PropTypes.object
   };
 
   componentDidMount() {
     this.mounted = true;
-    getIdentityProviders().then(r => {
+    Promise.all([
+      getIdentityProviders(),
+      this.props.fetchAppState()
```
stas-vilchik: When forced authentication is enabled, this WS fails with 401. In this case you need to avoid calling it, and assume that it's usual SonarQube on-premise.

gregaubert: :thinking: I didn't think about that... I'm not sure how we could avoid calling it. Because unless it fails we don't know if there is forced authentication...
But I could still call it and prevent it from raising an error and consider that we are on-premise as usual.

server/sonar-web/src/main/js/api/auth.ts
```diff
@@ -34,6 +35,21 @@ export function logout(): Promise<Response> {
     .then(basicCheckStatus);
 }
 
+export function tryGetGlobalNavigation(): Promise<any> {
```
stas-vilchik: I'd put it in `api/nav.ts`

PR #2638, opened 2017-10-09
Stas Hardening Oct 9

stas-vilchik:
To test SONAR-8371 you can search for java rule with key `S1820 `

### DIFF
--- a/server/sonar-web/src/main/js/apps/coding-rules/templates/coding-rules-workspace-list-item.hbs
+++ b/server/sonar-web/src/main/js/apps/coding-rules/templates/coding-rules-workspace-list-item.hbs
@@ -16,7 +16,7 @@
 
     <td>
       <div class="coding-rule-title">
-        <a class="js-rule link-no-underline">{{name}}</a>
+        <a class="js-rule link-no-underline" href="{{permalink}}">{{name}}</a>
       </div>
     </td>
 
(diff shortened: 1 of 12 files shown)


server/sonar-web/src/main/js/apps/projectsManagement/Search.tsx
```diff
@@ -185,6 +185,7 @@ export default class Search extends React.PureComponent<Props, State> {
       <td className="thin nowrap text-middle">
         <DateInput
           inputClassName="input-medium"
+          maxDate="+0"
```
gregaubert: I guess it could be nice do this for every usage of component `DateInput` ?

server/sonar-web/src/main/js/apps/coding-rules/templates/coding-rules-workspace-list-item.hbs
```diff
@@ -17,6 +17,8 @@
     <td>
       <div class="coding-rule-title">
         <a class="js-rule link-no-underline" href="{{permalink}}">{{name}}</a>
+        <span class="outline-badge spacer-left" title="{{t 'coding_rules.rule_template.title'}}"
```
gregaubert: I think you forgot a check on `isTemplate`, currently all rules have the badge

stas-vilchik: 🙀

PR #2639, opened 2017-10-09
SLF4J-related fixes

dtrebbien:


sns-seb:
Hello @dtrebbien 

Thanks very much for this contribution. I merged it as part of #2645.

FYI, the changes you provided in commit "Incorporate the constant property names into the logger format string" will lead to a rule improvement (see https://jira.sonarsource.com/browse/SONARJAVA-2518)

PR #2640, opened 2017-10-09
Correct the thread safety of logOnce()

dtrebbien:
[logOnce()](https://github.com/SonarSource/sonarqube/blob/6810d4b43783fe4dcbfd7d2361fe5584f261dc89/sonar-scanner-engine/src/main/java/org/sonar/scanner/sensor/DefaultSensorStorage.java#L211) is documented as thread safe; however, it calls Set\<E>.add(E) on the `alreadyLogged` member. This mutates the set and therefore is not thread safe.

To fix, wrap `alreadyLogged` as a synchronized set.


sns-seb:
Hello @dtrebbien,

Thank you for this PR.

Code is indeed not thread safe.
But here, there is really no need for the code to be thread safe.
Actually, the problem is the comment on method `logOnce()` which is misleading and should be removed. (I checked with the author of the comment, @dbmeneses, and the intent of the comment was to "flag" code obviously already thread safe and in this case, it was a mistake).

I'll remove the comment as I merge another PR.

PR #2644, opened 2017-10-10
fix CE startup log

sns-seb:


server/sonar-ce/src/main/java/org/sonar/ce/app/CeServer.java
```diff
@@ -138,6 +139,7 @@ public CeMainThread() {
     @Override
     public void run() {
       boolean startupSuccessful = attemptStartup();
+      this.operational = startupSuccessful;
```
henryju: Can't you directly assign `this.operational = attemptStartup();` and remove startupSuccessful local variable?

sns-seb: I could, but having the variable `startupSuccessful` strongly improved readability of this piece of code, doesn't it?

Since, eg., you did have trouble of this code, I believe anything helping understanding it is welcome :)

henryju: As you want. On my side it was confusing, since I was looking for the use of this local variable.

PR #2649, opened 2017-10-10
SONAR-9924 Analyze all files even in short branches

janos-ss:


sonar-scanner-engine/src/main/java/org/sonar/scanner/sensor/DefaultSensorStorage.java
```diff
@@ -200,7 +203,11 @@ public DefaultSensorStorage(MetricFinder metricFinder, ModuleIssues moduleIssues
   @Override
   public void store(Measure newMeasure) {
     if (newMeasure.inputComponent() instanceof DefaultInputFile) {
-      ((DefaultInputFile) newMeasure.inputComponent()).setPublished(true);
+      DefaultInputFile defaultInputFile = (DefaultInputFile) newMeasure.inputComponent();
+      if (shouldSkipStorage(defaultInputFile)) {
```
henryju: I have a bad feeling about this duplicate pattern. I fear we might forget it if some day we add a new "domain" to store. But I don't have a better proposal :) BTW, looking at DefaultSensorContext, I see we are returning a NO_OP builder for some domains (coverage, cpd, ...) when we are on a short branch so in theory it is useless to filter later in the SensorStorage.

janos-ss: I saw the NO_OP builders, but decided to add these checks anyway, to make them explicit, rather than implicitly depending on the implementation details of `DefaultSensorContext`.

henryju: Ok, fine for me. At least it make all methods in SensorStorage consistant.

PR #2653, opened 2017-10-10
Fix security vulnerabilities

sjoshi10:
Fix security vulnerabilities for xstream and logback package. CVE-2017-7957 and CVE-2017-5929


simonbrandhof:
Note that SonarQube does not suffer from https://nvd.nist.gov/vuln/detail/CVE-2017-5929 because logback socket server is not enabled.

simonbrandhof:
Replaced by #2672

PR #2669, opened 2017-10-12
SONAR-8798 fix error when sonar.log.maxFiles=0

simonbrandhof:


server/sonar-process/src/test/java/org/sonar/process/logging/Log4JPropertiesBuilderTest.java
```diff
@@ -140,6 +140,31 @@ public void configureGlobalFileLog_sets_properties_for_daily_time_rolling_policy
...
+    Log4JPropertiesBuilder underTest = newLog4JPropertiesBuilder(
+      ROLLING_POLICY_PROPERTY, "time:yyyy-MM-dd",
+      PROPERTY_MAX_FILES, "0");
+    underTest.configureGlobalFileLog(esRootLoggerConfig, logDir, logPattern);
+
+    verifyTimeRollingPolicy(underTest, logDir, logPattern, "yyyy-MM-dd", 100_000);
+  }
+
+  @Test
+  public void size_rolling_policy_has_large_max_files_if_property_is_zero() throws IOException {
```
sns-seb: missing test for when property is less than zero. Code would also put large max file

sns-seb: one option is to use "- new Random().nextInt(10)" instead of hardcoded zero :)

PR #2671, opened 2017-10-13
Feature/dm/marketplace

dbmeneses:


server/sonar-server/src/main/java/org/sonar/server/plugins/AbstractPluginUninstaller.java
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.commons.io.FileUtils;
+import org.picocontainer.Startable;
+import org.sonar.core.platform.PluginInfo;
+import org.sonar.core.util.stream.MoreCollectors;
+
+import static java.lang.String.format;
+import static org.apache.commons.io.FileUtils.forceMkdir;
+
+public abstract class AbstractPluginUninstaller implements Startable {
+  protected static final String PLUGIN_EXTENSION = "jar";
```
sns-seb: should be private

server/sonar-server/src/main/java/org/sonar/server/plugins/AbstractPluginUninstaller.java
```diff
@@ -0,0 +1,96 @@
...
+
+  protected final ServerPluginRepository serverPluginRepository;
```
sns-seb: should be private

server/sonar-server/src/main/java/org/sonar/server/plugins/AbstractPluginUninstaller.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  private static Collection<File> listJarFiles(File dir) {
+    if (dir.exists()) {
+      return FileUtils.listFiles(dir, new String[] {PLUGIN_EXTENSION}, false);
+    }
+    return Collections.emptyList();
+  }
+
+  protected void ensurePluginIsInstalled(String key) {
```
sns-seb: should be private

server/sonar-server/src/main/java/org/sonar/server/plugins/edition/EditionInstaller.java
```diff
@@ -0,0 +1,92 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.plugins.edition;
```
sns-seb: hard to find strong arguments, but I would have used "org.sonar.server.editions.plugins"

sns-seb: actually, juste noticed that error message refer to installation of edition rather than to installation of plugin

I'd say one point in favor of using "edition.plugins" rather that "plugins.edition" :)

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginDownloaderTest.java
```diff
@@ -0,0 +1,93 @@
...
+  private ServerFileSystem fs;
+  @Mock
+  private HttpDownloader httpDownloader;
+
+  private File downloadDir;
+  private EditionPluginDownloader downloader;
+
+  @Before
+  public void setUp() throws IOException {
+    MockitoAnnotations.initMocks(this);
```
sns-seb: long time since I've seen usage of MockitoAnnotations. I'd say it's not the trend in SQ's code.

I personally think it's useless compared to just writing mock(Foo.class) and not using MockitoAnnotation saves from writing a setup method when there is no other need for it (not the case here)

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginDownloaderTest.java
```diff
@@ -0,0 +1,93 @@
...
+import org.sonar.updatecenter.common.UpdateCenter;
+import org.sonar.updatecenter.common.Version;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Matchers.anyBoolean;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class EditionPluginDownloaderTest {
```
sns-seb: when `updateCenterMatrixFactory.getUpdateCenter(anyBoolean())` returns `Optional.empty()` is not tested

As this related to manual installation. I think a FIXME would be sane to not forget about it.

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginDownloaderTest.java
```diff
@@ -0,0 +1,93 @@
...
+  public void setUp() throws IOException {
+    MockitoAnnotations.initMocks(this);
+    downloadDir = temp.newFolder();
+    when(updateCenterMatrixFactory.getUpdateCenter(anyBoolean())).thenReturn(Optional.of(updateCenter));
+    when(fs.getDownloadedEditionPluginsDir()).thenReturn(downloadDir);
+    downloader = new EditionPluginDownloader(updateCenterMatrixFactory, httpDownloader, fs);
+  }
+
+  @Test
+  public void use_correct_dir() throws IOException, URISyntaxException {
```
sns-seb: I read this test thoroughly several times.
Still, I don't understand it. (I intentionally did not read the production code)

What is this tmp file? why is it supposed to be deleted when call "installEdition"?

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginUninstallerTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+
+public class EditionPluginUninstallerTest {
+  @Rule
+  public TemporaryFolder temp = new TemporaryFolder();
+
+  @Test
+  public void use_correct_directory() {
```
sns-seb: what about "start_creates_uninstall_directory"?

I'm suggesting this because I had to read the test to understand the name of the method (which I then found quite vague). I think it should be the other way around.

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginUninstallerTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.sonar.server.platform.ServerFileSystem;
+import org.sonar.server.plugins.ServerPluginRepository;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+import static org.mockito.Mockito.when;
+
+public class EditionPluginUninstallerTest {
```
sns-seb: none of the public methods of `EditionPluginUninstaller` but `start` are tested

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginDownloaderTest.java
sns-seb: none of the public methods of `EditionPluginDownloader` but `installEdition` are tested

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginDownloaderTest.java
sns-seb: a test with several plugin keys is missing

(as far as we tested, the code could fail when multiple keys are provided)

What happens if input Set contains null?

server/sonar-server/src/main/java/org/sonar/server/plugins/edition/EditionPluginDownloader.java
```diff
@@ -0,0 +1,52 @@
...
+  private final UpdateCenterMatrixFactory updateCenterMatrixFactory;
+
+  public EditionPluginDownloader(UpdateCenterMatrixFactory updateCenterMatrixFactory, HttpDownloader downloader, ServerFileSystem fileSystem) {
+    super(fileSystem.getDownloadedEditionPluginsDir(), downloader);
+    this.updateCenterMatrixFactory = updateCenterMatrixFactory;
+  }
+
+  public void installEdition(Set<String> pluginKeys) {
+    Optional<UpdateCenter> updateCenter = updateCenterMatrixFactory.getUpdateCenter(true);
+    Set<Release> pluginsToInstall = new HashSet<>();
```
sns-seb: no need to create this Set when updateCenter is not present

server/sonar-server/src/main/java/org/sonar/server/plugins/AbstractPluginUninstaller.java
```diff
@@ -0,0 +1,96 @@
...
+  public void uninstall(String pluginKey) {
+    ensurePluginIsInstalled(pluginKey);
+    serverPluginRepository.uninstall(pluginKey, uninstallDir);
+  }
+
+  public void cancelUninstalls() {
+    serverPluginRepository.cancelUninstalls(uninstallDir);
+  }
+
+  public List<String> getUninstalledPluginFilenames() {
```
sns-seb: never used

server/sonar-server/src/main/java/org/sonar/server/platform/ServerFileSystem.java
```diff
@@ -84,4 +91,8 @@
    * @return an existing file
    */
   File getPluginIndex();
+  
+  File uninstalledPluginsDir();
```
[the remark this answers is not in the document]
dbmeneses: agree

server/sonar-server/src/main/java/org/sonar/server/platform/ServerFileSystem.java
```diff
@@ -84,4 +91,8 @@
...
+  
+  File uninstalledEditionPluginsDir();
```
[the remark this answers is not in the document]
dbmeneses: ok

server/sonar-server/src/main/java/org/sonar/server/platform/ServerFileSystemImpl.java
```diff
@@ -84,6 +88,11 @@ public File getDownloadedPluginsDir() {
   }
 
   @Override
+  public File getDownloadedEditionPluginsDir() {
```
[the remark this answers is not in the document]
dbmeneses: yep

server/sonar-server/src/main/java/org/sonar/server/platform/ServerFileSystemImpl.java
```diff
@@ -84,6 +88,11 @@ public File getDownloadedPluginsDir() {
...
+    return new File(getHomeDir(), "extensions/new-edition");
```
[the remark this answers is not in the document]
dbmeneses: Other methods include that reference as well. It relies on other class (downloader) to create the directory.
I will remove it for all the methods.

server/sonar-server/src/main/java/org/sonar/server/plugins/AbstractPluginDownloader.java
```diff
@@ -0,0 +1,131 @@
...
+    try {
+      if (downloadDir.exists()) {
+        org.sonar.core.util.FileUtils.cleanDirectory(downloadDir);
+      }
+    } catch (IOException e) {
+      throw new IllegalStateException("Fail to clean the plugin downloads directory: " + downloadDir, e);
+    }
+  }
+
+  public boolean hasDownloads() {
```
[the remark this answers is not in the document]
dbmeneses: All such methods from Downloader were inherited from the old implementation, so I didn't change anything.
I will review accessibility for all of them.

server/sonar-server/src/main/java/org/sonar/server/plugins/edition/EditionInstaller.java
dbmeneses: one point in favor of 'plugins.edition' is to have these classes next of the ones in 'plugins'. They share hierarchy, it's kind of an specialization of plugins.

server/sonar-server/src/main/java/org/sonar/server/plugins/edition/EditionInstallerExecutor.java
```diff
@@ -0,0 +1,41 @@
...
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonar.server.plugins.edition;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+
+public class EditionInstallerExecutor {
```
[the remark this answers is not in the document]
dbmeneses: well spotted, specially since the threads are not daemon! (they should be daemon, btw)

server/sonar-server/src/main/java/org/sonar/server/plugins/edition/EditionPluginDownloader.java
dbmeneses: ok

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginDownloaderTest.java
dbmeneses: if you don't write a set up method, it's a different behaviour, and you risk interference between tests by reusing the mock.
I think when you have a lot of mocks like in this case this looks cleaner, but it really doesn't make much difference.
Also using the annotation can make mock of generics type-safe.

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginUninstallerTest.java
dbmeneses: All methods are inherited by AbstractPluginuninstaller, which is tested by PluginUninstaller. The point here is just to test from where comes the uninstall directory

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginDownloaderTest.java
dbmeneses: All other methods are inherited by AbstractPluginDownloader, which is tested by PluginDownloader.

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginDownloaderTest.java
dbmeneses: Ok, I will rewrite it

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginUninstallerTest.java
dbmeneses: ok, will improve it

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginDownloaderTest.java
sns-seb: @dbmeneses there is not risk of interference between tests method as test methods all use a different test object with JUnit (concern is valid with TestNG, though)

sns-seb: > Also using the annotation can make mock of generics type-safe.

almost true. It only saves from an unchecked cast :)

server/sonar-server/src/test/java/org/sonar/server/plugins/edition/EditionPluginDownloaderTest.java
dbmeneses: I will work on this in the next ticket

server/sonar-server/src/main/java/org/sonar/server/platform/db/migration/DatabaseMigrationImpl.java
```diff
@@ -69,38 +60,24 @@ public DatabaseMigrationImpl(DatabaseMigrationExecutorService executorService, M
...
-
-    running.set(true);
-    executorService.execute(this::doDatabaseMigration);
-  }
-
   private void doDatabaseMigration() {
-    migrationState.setStatus(Status.RUNNING);
-    migrationState.setStartedAt(new Date());
-    migrationState.setError(null);
     Profiler profiler = Profiler.create(LOGGER);
```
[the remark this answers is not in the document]
dbmeneses: it's used in the catch

server/sonar-server/src/main/java/org/sonar/server/plugins/AbstractPluginDownloader.java
```diff
@@ -0,0 +1,131 @@
...
+import static org.apache.commons.io.FileUtils.copyFileToDirectory;
+import static org.apache.commons.io.FileUtils.forceMkdir;
+import static org.apache.commons.io.FileUtils.toFile;
+import static org.apache.commons.lang.StringUtils.substringAfterLast;
+import static org.sonar.core.util.FileUtils.deleteQuietly;
+
+public class AbstractPluginDownloader implements Startable {
+  private static final Logger LOG = Loggers.get(AbstractPluginDownloader.class);
+
+  protected static final String TMP_SUFFIX = "tmp";
```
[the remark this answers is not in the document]
dbmeneses: fixed

dbmeneses:
Yep, I would certainly not merge into master :)

PR #2683, opened 2017-10-16
Fix some logger calls

dtrebbien:
The `org.sonar.api.utils.log.Logger` interface supports logging instances of exception/throwable directly at the warn and error levels, but not at the trace, debug, or info levels. Instead, at the trace, debug, or info levels, the exception/throwable must be formatted as a regular format arg (Object).

### DIFF
--- a/server/sonar-ce/src/main/java/org/sonar/ce/taskprocessor/CeProcessingSchedulerImpl.java
+++ b/server/sonar-ce/src/main/java/org/sonar/ce/taskprocessor/CeProcessingSchedulerImpl.java
@@ -87,7 +87,7 @@ public class CeProcessingSchedulerImpl implements CeProcessingScheduler {
       try {
         Thread.sleep(200L);
       } catch (InterruptedException e) {
-        LOG.debug("Graceful stop period has been interrupted", e);
+        LOG.debug("Graceful stop period has been interrupted: {}", e);
         Thread.currentThread().interrupt();
         break;
       }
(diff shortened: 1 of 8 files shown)


server/sonar-ce/src/main/java/org/sonar/ce/taskprocessor/CeProcessingSchedulerImpl.java
```diff
@@ -87,7 +87,7 @@ public void stopScheduling() {
       try {
         Thread.sleep(200L);
       } catch (InterruptedException e) {
-        LOG.debug("Graceful stop period has been interrupted", e);
+        LOG.debug("Graceful stop period has been interrupted: {}", e);
```
simonbrandhof: good catch !

henryju: Are you sure? Here the point is to attache the stacktrace to the log. Not to log is as part of the message.

sns-seb: I agree, that's not a good catch at all. The stacktrace of the error is lost.

dtrebbien: `LOG` here is not an SLF4J/Logback logger, but a `org.sonar.api.utils.log.Logger`:

https://github.com/SonarSource/sonarqube/blob/6bf0c30c664b7eb6e817b74d0a5e2948627c0caa/server/sonar-ce/src/main/java/org/sonar/ce/taskprocessor/CeProcessingSchedulerImpl.java#L38

If you take a look at the [`org.sonar.api.utils.log.Logger` interface](https://github.com/SonarSource/sonarqube/blob/6bf0c30c664b7eb6e817b74d0a5e2948627c0caa/sonar-plugin-api/src/main/java/org/sonar/api/utils/log/Logger.java), only at the warn and error levels is there the ability to log an exception/throwable with full stack trace.

With regard to this specific change, the absence of a debug(String, Throwable) method means that debug(String, @Nullable Object) is called instead. And because there is no formatting anchor `{}` in the format string, no information about the exception would be logged; the only thing that would be logged is "Graceful stop period has been interrupted".

If preserving the stack trace is desired, then the log should be at the warn or error level.

PR #2691, opened 2017-10-16
SONAR-9335 Remove some deprecated code from 4.X

julienlancelot:


julienlancelot:
Not enough value is brought by this PR, we'll wait for 7.X to remove more deprecated code (since 5.X)

PR #2694, opened 2017-10-16
SONAR-9527 Clean up of deactivated users

ehartmann:


server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v67/CleanupDisabledUsers.java
```diff
@@ -0,0 +1,67 @@
...
+      "  external_identity_provider is not null ) ");
+    massUpdate.update("update users set " +
+      "  email = null, " +
+      "  crypted_password = null, " +
+      "  salt = null, " +
+      "  external_identity = null, " +
+      "  external_identity_provider = null, " +
+      "  updated_at = ? " +
+      " where id = ?");
+    massUpdate.rowPluralName("deactivated users with remaining data");
```
julienlancelot: IMO "deactivated users" is enough

server/sonar-db-migration/src/test/java/org/sonar/server/platform/db/migration/version/v67/CleanupDisabledUsersTest.java
```diff
@@ -0,0 +1,203 @@
...
+  private final static String SELECT_USERS = "select name, scm_accounts, user_local, login, crypted_password, salt, email, external_identity, external_identity_provider, active, is_root, onboarded, created_at, updated_at from users";
+
+  private System2 system2 = new TestSystem2().setNow(NOW);
+
+  @Rule
+  public CoreDbTester db = CoreDbTester.createForSchema(CleanupDisabledUsersTest.class, "users.sql");
+
+  private CleanupDisabledUsers underTest = new CleanupDisabledUsers(db.database(), system2);
+
+  private final List<User> users = Arrays.asList(
```
julienlancelot: Such shared data in UT is misleading => it's really hard to understand why test case is using it.
I prefer that each test is creating its own objects (using factory if necessary to have small and clear test)

server/sonar-db-migration/src/test/java/org/sonar/server/platform/db/migration/version/v67/CleanupDisabledUsersTest.java
```diff
@@ -0,0 +1,203 @@
...
+      this.name = randomAlphanumeric(200);
+      this.updatedAt = PAST;
+      this.createdAt = PAST;
+    }
+
+    private void insert() {
+      db.executeInsert("USERS", toMap());
+    }
+
+    private User cleanedupUser() {
```
julienlancelot: there's too much logic in this

server/sonar-db-migration/src/test/java/org/sonar/server/platform/db/migration/version/v67/CleanupDisabledUsersTest.java
```diff
@@ -0,0 +1,203 @@
...
+    underTest.execute();
+  }
+
+  @Test
+  public void execute_must_update_database() throws SQLException {
+    users.forEach(User::insert);
+
+    underTest.execute();
+
+    assertDatabaseContainsExpectedValues();
```
julienlancelot: Same comment as for the shared usage of list of users : it's better to clearly display what is the output when expect instead of using a shared method

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v67/CleanupDisabledUsers.java
ehartmann: Ok, I've added this because we are not updating deactivated users who have already been cleaned up.

PR #2700, opened 2017-10-17
Add marketplace form to apply a license

gregaubert:


server/sonar-web/src/main/js/apps/marketplace/App.tsx
```diff
@@ -148,6 +150,19 @@ export default class App extends React.PureComponent<Props, State> {
       () => {}
     );
 
+  fetchEditionStatus = () =>
```
stas-vilchik: there is something wrong with loading here: the flag is not always set, and it's actually not used in render

gregaubert: Ah yeah I forgot to use it in render, I'll fix it.
As for not every `fetchSomething` function setting it, I think some of them should not, like `fetchEditionStatus` that wont use for sure because I'm gonna change things to refresh the status every 5 seconds.
And we could discuss about `fetchPendingPlugins`, I'm not sur this one really deserve a loading flag. At page loading we will display the pending plugins as an alert if there are some, no need to notify the user that we are fetching that... And when clicking an install button there is already the spinner of the current install before the pending actions are displayed...

gregaubert:
Sorry I forgot to provide the needed information to test, you need to put that in you `sonar.properties` : 
```
sonar.updatecenter.url=http://update.sonarsource.org/update-center-dev.properties
sonar.editions.jsonUrl=https://gist.githubusercontent.com/gregaubert/e34535494f8a94bec7cbc4d750ae7d06/raw/74b242285f634b85ec257e03676570234d5e8494/editions.json
```

server/sonar-web/src/main/js/apps/marketplace/components/LicenseEditionForm.tsx
```diff
@@ -0,0 +1,116 @@
...
+          className="modal-body"
+          edition={edition}
+          updateLicense={this.handleLicenseChange}
+        />
+
+        <footer className="modal-foot">
+          {this.state.loading && <i className="spinner spacer-right" />}
+          {status &&
+          ['NO_INSTALL', 'AUTOMATIC_INSTALL'].includes(status) && (
+            <button className="js-confirm" onClick={this.handleConfirmClick}>
```
stas-vilchik: disable the button while loading

PR #2701, opened 2017-10-17
SONAR-7992 Restrict length of version names

bartfastiel:
to 100 characters

### DIFF
--- a/server/sonar-db-dao/src/main/java/org/sonar/db/component/SnapshotDto.java
+++ b/server/sonar-db-dao/src/main/java/org/sonar/db/component/SnapshotDto.java
@@ -22,6 +22,8 @@ package org.sonar.db.component;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nullable;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 public final class SnapshotDto {
 
   /**
@@ -29,6 +31,7 @@ public final class SnapshotDto {
    */
   public static final String STATUS_UNPROCESSED = "U";
   public static final String STATUS_PROCESSED = "P";
+  public static final int MAX_VERSION_LENGTH = 100;
 
   private Long id;
   private String uuid;
@@ -116,6 +119,10 @@ public final class SnapshotDto {
   }
 
   public SnapshotDto setVersion(@Nullable String version) {
+    if (version != null) {
+      checkArgument(version.length() <= MAX_VERSION_LENGTH,
+        "Event name length (%s) is longer than the maximum authorized (%s). '%s' was provided.", version.length(), MAX_VERSION_LENGTH, version);
+    }
     this.version = version;
     return this;
   }
--- a/server/sonar-db-dao/src/test/java/org/sonar/db/component/SnapshotDtoTest.java
+++ b/server/sonar-db-dao/src/test/java/org/sonar/db/component/SnapshotDtoTest.java
@@ -19,12 +19,18 @@
  */
 package org.sonar.db.component;
 
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.ExpectedException;
 
+import static org.apache.commons.lang.StringUtils.repeat;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.sonar.api.utils.DateUtils.parseDate;
 
 public class SnapshotDtoTest {
+
+  @Rule
+  public ExpectedException expectedException = ExpectedException.none();
 
   @Test
   public void test_getter_and_setter() throws Exception {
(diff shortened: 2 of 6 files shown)


sonar-scanner-engine/src/main/java/org/sonar/scanner/report/ComponentsPublisher.java
```diff
@@ -167,6 +169,7 @@ private boolean shouldSkipComponent(DefaultInputComponent component, Collection<
   private void writeVersion(DefaultInputModule module, ScannerReport.Component.Builder builder) {
     String version = getVersion(module);
     if (version != null) {
+      checkArgument(version.length() <= 100, "The version name must not be longer then 100 characters");
```
bartfastiel: @henryju Could you please have a look at this? Is this the way to solve this? Or would it be better here: https://github.com/SonarSource/sonarqube/blob/master/sonar-plugin-api/src/main/java/org/sonar/api/batch/bootstrap/ProjectDefinition.java#L194 ?

henryju: @bartfastiel we usually try to fail as soon as possible. I suggest to do the check in ProjectReactorValidator

henryju: Also note that it would be safer to also validate on CE side, when reading the field from the report, since on SonarCloud we may receive corrupted report (using a modified scanner for example).

server/sonar-server/src/main/java/org/sonar/server/projectanalysis/ws/UpdateEventAction.java
```diff
@@ -95,12 +95,14 @@ public void handle(Request httpRequest, Response httpResponse) throws Exception
   }
 
   private UpdateEventResponse doHandle(UpdateEventRequest request) {
+    String name = request.getName();
+    checkArgument(name == null || name.length() <= 100, "The name must not be longer then 100 characters.");
```
teryk: The form below is used elsewhere:
```
Event name length (%s) is longer than the maximum authorized (%s). '%s' was provided.
```

teryk: I wonder if we really want to restrict event length to 100 characters for events that are not Version events

server/sonar-db-dao/src/main/java/org/sonar/db/component/SnapshotDto.java
```diff
@@ -116,6 +118,8 @@ public String getVersion() {
   }
 
   public SnapshotDto setVersion(@Nullable String version) {
+    checkArgument(version == null || version.length() <= 100, "\"%s\" is not a valid version name. " +
```
teryk: Please use the form given before

server/sonar-db-dao/src/test/java/org/sonar/db/component/SnapshotDtoTest.java
```diff
@@ -48,4 +55,17 @@ public void test_getter_and_setter() throws Exception {
     assertThat(snapshotDto.getPeriodDate()).isEqualTo(parseDate("2014-06-01").getTime());
   }
 
+  @Test
+  public void fail_if_version_name_is_longer_then_100_characters() throws Exception {
+    SnapshotDto snapshotDto = new SnapshotDto();
+    snapshotDto.setVersion(null);
+    snapshotDto.setVersion("1.0");
+    snapshotDto.setVersion(IntStream.range(0, 100).mapToObj(i -> "a").collect(Collectors.joining()));
```
teryk: Apache provides utility methods, `StringUtils.repeat(…)`. I think it's overkill to use stream here to build a string

server/sonar-db-dao/src/main/java/org/sonar/db/component/SnapshotDto.java
bartfastiel: Done.

server/sonar-db-dao/src/test/java/org/sonar/db/component/SnapshotDtoTest.java
bartfastiel: Done.

server/sonar-server/src/main/java/org/sonar/server/projectanalysis/ws/UpdateEventAction.java
bartfastiel: Done.

bartfastiel:
I had to rebase, and I'm planning to merge once the QA is green.

PR #2708, opened 2017-10-18
SONAR-10002 support failure of automatic install

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/edition/StandaloneEditionManagementStateImpl.java
```diff
@@ -42,12 +45,10 @@
...
   private static final String PENDING_EDITION_KEY = "pendingEditionKey";
   private static final String PENDING_LICENSE = "pendingLicense";
+  private static final String AUTOMATIC_INSTALL_ERROR_MESSAGE = "autoInstallError";
 
   private final DbClient dbClient;
-  private String currentEditionKey;
-  private PendingStatus pendingInstallationStatus;
-  private String pendingEditionKey;
-  private String pendingLicense;
+  private final AtomicReference<State> state = new AtomicReference<>();
```
dbmeneses: I see a couple of problems in terms of thread safety in this class.
1. There is one getter for each field in the state. When another class uses these getters, there is no guarantee that the values will be consistent with each other, since they might be changed between each request. It would be better to return a copy of the state.
2. What's the point of having an atomic reference to the state? Unless atomic operations are used, getting and settings separately has the same problems as doing it in a regular member var. It's fine to just use a var since we have the sync blocks.
3. newEditionWithoutInstall needs to be synchronized as well.

PR #2715, opened 2017-10-18
SONAR-10010 Add WS api/edition/form_data

dbmeneses:


server/sonar-server/src/main/java/org/sonar/server/edition/ws/FormDataAction.java
```diff
@@ -0,0 +1,68 @@
...
+    this.server = server;
+    this.measuresIndex = measuresIndex;
+  }
+
+  @Override
+  public void define(WebService.NewController controller) {
+    controller.createAction("form_data")
+      .setSince("6.7")
+      .setPost(true)
+      .setDescription("Provide data for fields: the server ID and the total number of lines of code.")
```
sns-seb: what about "Provide data to prefill license request forms: ..." ?

server/sonar-server/src/main/java/org/sonar/server/edition/ws/FormDataAction.java
```diff
@@ -0,0 +1,68 @@
...
+    this.userSession = userSession;
+    this.server = server;
+    this.measuresIndex = measuresIndex;
+  }
+
+  @Override
+  public void define(WebService.NewController controller) {
+    controller.createAction("form_data")
+      .setSince("6.7")
+      .setPost(true)
```
sns-seb: should be GET

PR #2722, opened 2017-10-19
SONAR-9142 Remove previous_analysis option

ehartmann:


server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v67/MigratePreviousAnalysisLeakPeriod.java
```diff
@@ -0,0 +1,57 @@
...
+
+package org.sonar.server.platform.db.migration.version.v67;
+
+import java.sql.SQLException;
+import org.sonar.db.Database;
+import org.sonar.db.dialect.Oracle;
+import org.sonar.server.platform.db.migration.step.DataChange;
+import org.sonar.server.platform.db.migration.step.MassUpdate;
+
+public class MigratePreviousAnalysisLeakPeriod extends DataChange {
```
julienlancelot: I would rename it to `RemovePreviousAnalysisLeakPeriod` or `MigratePreviousAnalysisToPreviousVersion`

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v67/MigratePreviousAnalysisLeakPeriod.java
```diff
@@ -0,0 +1,57 @@
...
+
+  private static final String ORACLE_SQL = "select id from properties " +
+    " where prop_key = 'sonar.leak.period' and (text_value='previous_analysis' or dbms_lob.compare(clob_value, to_clob('previous_analysis')) = 0)";
```
julienlancelot: I really don't like to have specific query => As the number of data related to leak in properties table is not too big, you can remove `and (text_value='previous_analysis'` and filter on "previous_analysis" value in the massUpdate.execute

julienlancelot: Forget what I've said before => clob_value only contains value greater than 4000 characters => "previous_analysis" is less than 400 characters so you can remove filter on clob_value

server/sonar-db-migration/src/test/java/org/sonar/server/platform/db/migration/version/v67/MigratePreviousAnalysisLeakPeriodTest.java
```diff
@@ -0,0 +1,188 @@
...
+  private final static String SELECT_PROPERTIES = "SELECT prop_key, is_empty, text_value, clob_value FROM properties";
+
+  @Rule
+  public CoreDbTester db = CoreDbTester.createForSchema(MigratePreviousAnalysisLeakPeriodTest.class, "properties.sql");
+
+  private MigratePreviousAnalysisLeakPeriod underTest = new MigratePreviousAnalysisLeakPeriod(db.database());
+
+  @Test
+  public void migration_must_update_the_database() throws SQLException {
+    List<Property> properties = generateAndInsertProperties();
```
julienlancelot: Not blocker, but I really don't like this pattern (one method at the begin that create users, and one method at the end that do the check) => I prefer to explicitly have the list of properties, as I've no idea which properties you've created.
See for instance [FixEmptyIdentityProviderInUsersTest](https://github.com/SonarSource/sonarqube/blob/master/server/sonar-db-migration/src/test/java/org/sonar/server/platform/db/migration/version/v66/FixEmptyIdentityProviderInUsersTest.java) => you know quickly which users are created, and which data do you expect at the end of the test without having to dig into other method

server/sonar-db-migration/src/test/java/org/sonar/server/platform/db/migration/version/v67/MigratePreviousAnalysisLeakPeriodTest.java
```diff
@@ -0,0 +1,188 @@
...
+    assertThat(db.select(SELECT_PROPERTIES)).isEqualTo(
+      expectedProperties.stream().map(Property::toMap).collect(Collectors.toList())
+    );
+  }
+
+  private List<Property> generateAndInsertProperties() {
+    List<Property> properties = ImmutableList.of(
+      new Property("sonar.leak.period", randomAlphanumeric(1000), null, false),
+      new Property("sonar.leak.period", "previous_version", null, false),
+      new Property("sonar.leak.period", null, "previous_analysis", false),
```
julienlancelot: This use case does not exist => clob_value will never contains "previous_analysis"

server/sonar-db-migration/src/test/java/org/sonar/server/platform/db/migration/version/v67/MigratePreviousAnalysisLeakPeriodTest.java
```diff
@@ -0,0 +1,188 @@
...
+        if ("previous_analysis".equals(result.textValue) || "previous_analysis".equals(result.clobValue)) {
+          result.clobValue = null;
+          result.textValue = "previous_version";
+        }
+
+        return result;
+      }).collect(Collectors.toList());
+  }
+
+  private class Property {
```
julienlancelot: Lot of logic here that could be avoid by using stream + assertj tuples ()

server/sonar-server/src/test/java/org/sonar/server/computation/task/projectanalysis/step/LoadPeriodsStepTest.java
```diff
@@ -232,26 +234,6 @@ public void no_period_by_days() {
...
-    assertThat(period.getModeParameter()).isNotNull();
-    assertThat(period.getSnapshotDate()).isEqualTo(1227934800000L);
-    assertThat(period.getAnalysisUuid()).isEqualTo("u1004");
-
-    assertThat(logTester.logs(LoggerLevel.DEBUG)).hasSize(1);
-    assertThat(logTester.logs(LoggerLevel.DEBUG).get(0)).startsWith("Compare to previous analysis (");
-  }
-
-  @Test
   public void no_period_by_previous_analysis() {
```
julienlancelot: This test should also have been removed

server/sonar-db-migration/src/test/java/org/sonar/server/platform/db/migration/version/v67/MigratePreviousAnalysisLeakPeriodTest.java
ehartmann: It does not exist currently but I was not sure that it was not the case before. I'll remove it.

server/sonar-server/src/test/java/org/sonar/server/computation/task/projectanalysis/step/LoadPeriodsStepTest.java
ehartmann: Oh thanks I've missed this one !

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v67/MigratePreviousAnalysisLeakPeriod.java
ehartmann: Yes I know, I wondered if I should or not read the value because functionally if the value is in clob we do not migrate it. I'm fine removing this.

julienlancelot: I understand, but if my understanding is correct, it could have never happen => Good thing to remove it !

PR #2728, opened 2017-10-20
SONAR-9333 Web service api/plugins/uninstall fails when plugin is alr…

dbmeneses:
…eady being uninstalling

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/plugins/ServerPluginRepository.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/plugins/ServerPluginRepository.java
@@ -316,10 +316,16 @@ public class ServerPluginRepository implements PluginRepository, Startable {
 
     for (String uninstallKey : uninstallKeys) {
       PluginInfo info = getPluginInfo(uninstallKey);
+
       try {
+        if (!getPluginFile(info).exists()) {
+          LOG.info("Plugin already uninstalled: {} [{}]", info.getName(), info.getKey());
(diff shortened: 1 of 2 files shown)


server/sonar-server/src/main/java/org/sonar/server/plugins/ServerPluginRepository.java
```diff
@@ -316,10 +316,16 @@ public void uninstall(String pluginKey, File uninstallDir) {
 
     for (String uninstallKey : uninstallKeys) {
       PluginInfo info = getPluginInfo(uninstallKey);
+
       try {
+        if (!getPluginFile(info).exists()) {
```
sns-seb: you should call `getPluginFile(info)` only once (rather than here and at line 328). Use a variable

sns-seb: If I call api/plugins/uninstall with plugin key foo, I'll get message "Plugin already uninstalled" while it's not true, there is no such plugin, at all.

To be more reliable, if `getPluginFile(info)` we should test file exists in uninstallDir. If so, display this message, if not, throw an error.

dbmeneses: The plugin info is loaded from memory. So the plugin existed at startup and was removed. 
I will use a var for the plugin file (even though this does not involve I/O)

PR #2737, opened 2017-10-20
Fix quality flaws

ehartmann:


server/sonar-server/src/test/java/org/sonar/server/settings/ChildSettingsTest.java
```diff
@@ -0,0 +1,124 @@
...
+    underTest.set(randomAlphanumeric(10), null);
+  }
+
+  @Test
+  public void childSettings_override_parent() {
+    String key = randomAlphanumeric(10);
+    parent.setProperty(key, randomAlphanumeric(20));
+    underTest.setProperty(key, randomAlphanumeric(10));
+
+    assertThat(underTest.get(key)).isNotEqualTo(parent.getString(key));
```
bartfastiel: You could add check, that it actually *is*  equal to the string that you set to `underTest`.

PR #2746, opened 2017-10-23
improve css management

stas-vilchik:
* stop using less
* move styles to app
* use theme


server/sonar-web/src/main/js/app/styles/select2.css
```diff
@@ -415,7 +415,7 @@ Version: 3.2 Timestamp: Mon Sep 10 10:38:04 PDT 2012
   right: 3px;
   top: 4px;
   width: 12px;
-  height: 13px;
+  height: var(--baseFontSize);
```
gregaubert: should probably not use fontsize var

PR #2757, opened 2017-10-25
Stas Hardening Oct 25

stas-vilchik:


server/sonar-web/src/main/js/apps/projectActivity/components/ProjectActivityAnalysesList.js
```diff
@@ -187,7 +188,9 @@ export default class ProjectActivityAnalysesList extends React.PureComponent {
             <li key={version.key || 'noversion'}>
               {version.version && (
                 <div className={classNames('project-activity-version-badge', { first: idx === 0 })}>
-                  <span className="badge">{version.version}</span>
+                  <Tooltip overlay={version.version} mouseEnterDelay={0.5}>
```
gregaubert: I think it could be nice to use a "Version {0}" text in the tooltips. I remember a comment that was saying that it might not be obvious that those badges are versions...

PR #2758, opened 2017-10-26
Hardening oct 25

gregaubert:


sonar-core/src/main/resources/org/sonar/l10n/core.properties
```diff
@@ -2184,7 +2184,7 @@ background_tasks.failing_count=Count of projects where processing of most recent
 #------------------------------------------------------------------------------
 system.application_nodes_title=Application Nodes
 system.are_you_sure_to_restart=Are you sure you want to restart the server?
-system.cluster_log_level.info=Changes apply to all Application nodes but not to Search nodes.
+system.cluster_log_level.info=Your selection affect all Application nodes but not the Search nodes.
```
stas-vilchik: "Search Engine nodes" maybe?

gregaubert: We are referring to "Search nodes" when in cluster mode because it matches the type of node. We also display "Search Nodes" in the list of node in the content of the page.
But I'll confirm it with Christophe just to be sure.

gregaubert: Confirmed by Christophe

PR #2759, opened 2017-10-26
wrap react-modal and react-select in own components

stas-vilchik:
just look on the amount of removed lines 😉


server/sonar-web/src/main/js/components/controls/Modal.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+import * as React from 'react';
+import * as ReactModal from 'react-modal';
+import * as classNames from 'classnames';
+
+// white-list only some props
+type AllowedProps = 'contentLabel' | 'onAfterOpen' | 'onRequestClose';
+type Props = Pick<ReactModal.Props, AllowedProps> & { medium?: boolean; large?: boolean };
```
gregaubert: Maybe we could add `& Partial<ReactModal.Props>`, to still allow other props of the original modal

server/sonar-web/src/main/js/components/controls/Modal.tsx
```diff
@@ -22,8 +22,14 @@ import * as ReactModal from 'react-modal';
 import * as classNames from 'classnames';
 
 // white-list only some props
-type AllowedProps = 'contentLabel' | 'onAfterOpen' | 'onRequestClose';
-type Props = Pick<ReactModal.Props, AllowedProps> & { medium?: boolean; large?: boolean };
+type AllowedProps = Pick<ReactModal.Props, 'contentLabel'>;
```
gregaubert: I'd say more like `MendatoryProps`

PR #2766, opened 2017-10-26
Skip compiling Test classes

kondareddyr:
-DskipTests just skips the test execution: the tests are still compiled. Where as-Dmaven.test.skip will skip both compilation and execution of the tests.

### DIFF
--- a/quick-build.sh
+++ b/quick-build.sh
@@ -41,4 +41,4 @@ $$$$"""$$$$$$$$$$uuu   uu$$$$$$$$$"""$$$"
 EOF
 
 ./stop.sh
-mvn clean install -B -e -V -DskipTests=true -Dmaven.source.skip=true -Pdev,skipSanityChecks $*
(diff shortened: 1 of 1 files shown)


simonbrandhof:
Thanks

quick-build.sh
```diff
@@ -41,4 +41,4 @@ $$$$"""$$$$$$$$$$uuu   uu$$$$$$$$$"""$$$"
 EOF
 
 ./stop.sh
-mvn clean install -B -e -V -DskipTests=true -Dmaven.source.skip=true -Pdev,skipSanityChecks $*
+mvn clean install -B -e -V -Dmaven.test.skip -Dmaven.source.skip=true -Pdev,skipSanityChecks $*
```
sns-seb: I was relying on the fact this script compiled the tests and just did not execute them.
It's convenient to refresh all local artifacts quickly in the morning.

I'll just stop using this script.

sns-seb: Also, it's convenient to just get compiler check on all code before just pushing to CI to get full test and QA feeback.

PR #2782, opened 2017-11-01
pass fast branch build

stas-vilchik:


simonbrandhof:
Instead of duplicating travis-utils/cancel_branch_build_with_pr, I suggest to improve the parent/child process call:

* travis-utils/cancel_branch_build_with_pr should not call Travis API to cancel the task but should return 0 or 1:
   ```
   # replace https://github.com/SonarSource/travis-utils/blob/master/bin/cancel_branch_build_with_pr#L28-L33 by:
   exit 1
   ```
* sonarqube/travis.sh must check the return code and exit if on branch with PR:
    ```
    cancel_branch_build_with_pr
    returned_code=$?
    if [ "${returned_code}" != "0" ]
    then
      # should cancel the job without marking it as failed
      exit 0
    fi
    ```

simonbrandhof:
The travis.sh condition can maybe be simplified: https://stackoverflow.com/a/16215397/229031
I didn't test, but it would be something like:
```
[[ $(cancel_branch_build_with_pr ]] || exit 1
```

simonbrandhof:
See https://github.com/SonarSource/travis-utils/pull/16

PR #2790, opened 2017-11-02
Refactoring for specialising Optional

ameyaKetkar:
Hello,
Like the specialisation of Functional interfaces in the previous pull request, these are the results I obtained while testing specialisation of java.util.Optional.
There have been a few autoboxings introduced by the refactorings, but they are in the test files. That was the price to pay from eliminating autoboxing from prod files.

### DIFF
--- a/server/sonar-db-dao/src/main/java/org/sonar/db/measure/ProjectMeasuresIndexerIterator.java
+++ b/server/sonar-db-dao/src/main/java/org/sonar/db/measure/ProjectMeasuresIndexerIterator.java
@@ -30,7 +30,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
+import java.util.OptionalDouble;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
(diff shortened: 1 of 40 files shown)


sns-seb:
Hello @ameyaKetkar ,

What was the scope of the analysis you ran? (my understanding is you used a tool)

I'm surprised the PR does not include fixes for https://github.com/SonarSource/sonarqube/blob/master/sonar-plugin-api/src/main/java/org/sonar/api/config/Configuration.java#L136 or https://github.com/SonarSource/sonarqube/blob/master/sonar-plugin-api/src/main/java/org/sonar/api/config/Configuration.java#L149, for example.

ameyaKetkar:
Hello,
Yes, those changes are pending along with MigrationHistory.java and DatabaseVersion.java. The reason I didn't check in earlier is because the refactorings required are a bit too complex for the current state of my tool :( . Especially the ones in ConfigurationTest.java and PurgeConfiguration.java. I have made those changes manually for now and committed them to my branch. 
Thank you for your interest.

ameyaKetkar:
Hello,
Yes, those changes are pending along with MigrationHistory.java and databaseVersion.java. The reason I didn't check in them earlier is because the refactorings required are a bit too complex for the current state of my tool. Especially the ones in ConfigurationTest.java and PurgeConfiguration.java. I have made those changes manually for now and committed them to my branch. 
Thank you.

ameyaKetkar:
My apologies. After adding the new changes, the build fails because there is an exception which i am unable to understand. I can just close this PR for now.

ameyaKetkar:
Hello @sns-seb ,
I have created another pull request for refactoring. 
https://github.com/SonarSource/sonarqube/pull/2838
(getDouble() in Configuration.java is not refactored since it has no calls with are autoboxed.)
Thank you.

PR #2800, opened 2017-11-09
SONAR-10057 upgrade PostgreSQL driver

simonbrandhof:
to fix compatibility with Postgres 10


sonar-application/src/main/assembly/conf/sonar.properties
```diff
@@ -34,7 +34,7 @@
 #sonar.jdbc.url=jdbc:oracle:thin:@localhost:1521/XE
 
 
-#----- PostgreSQL 8.x/9.x
+#----- PostgreSQL 9.x or greater
```
julienlancelot: Please either remove this line, or upgrade requirement guides

PR #2810, opened 2017-11-17
upgrade xstream version

sjoshi10:
Without specifying the version of xstream in the pom.xml, by default it installs xstream version 1.3.1. There is security vulnerability with this version and the CVS score is 7.5 [https://nvd.nist.gov/vuln/detail/CVE-2017-7957](url). To remove this vulnerability, xstream package needs to be upgraded to 1.4.10. I have updated the pom.xml with the fix.

PR #2811, opened 2017-11-17
ban xstream less than 1.4.9

sjoshi10:
Without specifying the version of xstream in the pom.xml, by default it installs xstream version 1.3.1. There is security vulnerability with this version and the CVS score is 7.5 https://nvd.nist.gov/vuln/detail/CVE-2017-7957. To remove this vulnerability, xstream package needs to be upgraded to 1.4.10. I have updated the pom.xml with the fix.


simonbrandhof:
Duplicate #2810

PR #2815, opened 2017-11-21
Add USER_PATCHES_HOME support and minor quality improvements

janos-ss:


janos-ss:
Good point about the README. Updated. That shebang works fine on Mac (and the recommended way anyway), and I'm confident about the other changes too, so going to merge without testing on Mac.

PR #2824, opened 2017-11-22
SONAR-10111 Move administration to default organization context

stas-vilchik:


server/sonar-web/src/main/js/app/components/nav/settings/SettingsNav.tsx
```diff
@@ -1,240 +0,0 @@
...
-        </ul>
-      </li>
-    );
-  }
-
-  render() {
-    const { editionStatus, extensions } = this.props;
-    const hasSupportExtension = extensions.find(extension => extension.key === 'license/support');
-
-    let notifComponent;
```
gregaubert: By removing that settings navbar we loose the notifications system of the marketplace.
So these notifications should probably be moved to the default organization navbar and be visible only for users with admin (root) rights.

server/sonar-web/src/main/js/app/components/DefaultOrganizationNavigation.tsx
```diff
@@ -19,121 +19,228 @@
...
-    <div className="navbar-context-header">
-      <h1 className="display-inline-block">
-        <OrganizationIcon className="little-spacer-right" />
-        <Link to="/projects" className="link-base-color link-no-underline">
-          <strong>{organization.name}</strong>
+  render() {
+    const { appState, organization } = this.props;
+    const portfoliosExtension = appState.globalPages.find(({ key }) => key === PORTFOLIOS);
+    const portfoliosLink = portfoliosExtension && (
+      <li key="portfolios">
```
gregaubert: I think you can use `this.renderLink` here

PR #2829, opened 2017-11-23
SONAR-9000 My organizations are the ones the user is a member of

teryk:


julienlancelot:
It seems that whatever the value of member, as long as a value is set, the results are the same.
Using `api/organizations/search?member=false` is returning the exact same thing as `api/organizations/search?member=true`

PR #2844, opened 2017-11-28
SONAR-10131 Purge of db table project_measures is incomplete

simonbrandhof:


server/sonar-db-dao/src/test/java/org/sonar/db/purge/PurgeConfigurationTest.java
```diff
@@ -57,39 +57,40 @@ public void should_delete_only_old_closed_issues() {
...
-    assertThat(underTest.scopesWithoutHistoricalData()).contains(Scopes.FILE)
-      .doesNotContain(Scopes.DIRECTORY);
+    assertThat(underTest.getQualifiersWithoutHistoricalData())
+      .containsExactlyInAnyOrder(Qualifiers.FILE, Qualifiers.UNIT_TEST_FILE);
     assertThat(underTest.maxLiveDateOfClosedIssues(now)).isEqualTo(DateUtils.addDays(now, -5));
   }
 
   @Test
-  public void delete_directory_if_in_settings() {
+  public void elete_directory_if_enabled_in_settings() {
```
bartfastiel: There is a typo here.

simonbrandhof:
Fed up with MySQL, the DELETE SQL request does not work:
```
You can't specify target table 'project_measures' for update in FROM clause
```

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v671/DeleteHistoryOfTestFileMeasures.java
```diff
@@ -0,0 +1,74 @@
...
+import org.sonar.db.dialect.H2;
+import org.sonar.db.dialect.MySql;
+import org.sonar.server.platform.db.migration.step.DataChange;
+
+public class DeleteHistoryOfTestFileMeasures extends DataChange {
+
+  private static final String MYSQL_SQL = "delete pm from project_measures pm " +
+    "inner join snapshots s on s.uuid = pm.analysis_uuid " +
+    "inner join projects c on c.uuid = pm.component_uuid " +
+    "where s.islast=? and c.qualifier in ('UTS', 'FIL')";
```
bartfastiel: We should not delete `FIL` components here, right?

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v671/DeleteHistoryOfTestFileMeasures.java
```diff
@@ -0,0 +1,74 @@
...
+
+  private static final String OTHER_SQL = "delete from project_measures " +
+    "where exists ( " +
+    "  select 1 from project_measures pm " +
+    "  inner join snapshots s on s.uuid = pm.analysis_uuid " +
+    "  inner join projects c on c.uuid = pm.component_uuid " +
+    "  where s.islast=? and c.qualifier in ('UTS', 'FIL')" +
```
bartfastiel: (occurance 2) We should not delete `FIL` components here, right?

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v671/DeleteHistoryOfTestFileMeasures.java
```diff
@@ -0,0 +1,74 @@
...
+    ")";
+
+  private static final String H2_SQL = "delete from project_measures " +
+    "where id in ( " +
+    "  select pm.id from project_measures pm " +
+    "  inner join snapshots s on s.uuid = pm.analysis_uuid " +
+    "  inner join projects c on c.uuid = pm.component_uuid " +
+    "  where s.islast=? and c.qualifier in ('UTS', 'FIL') " +
```
bartfastiel: (occurance 3) We should not delete `FIL` components here, right?

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v671/DeleteHistoryOfTestFileMeasures.java
bartfastiel: (occurance 2) 💣   Add a `where` clause, referring to the measure to delete. Currently the inner select is not correlated and this will **delete all unit test and file measures**!

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v671/DeleteHistoryOfTestFileMeasures.java
bartfastiel: (occurance 2) 💣   Add a `where` clause, referring to the measure to delete. Currently the inner select is not correlated and this will **delete all unit test and file measures**!

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v671/DeleteHistoryOfTestFileMeasures.java
simonbrandhof: Exact, but as described on the ticket there may have some file orphans. It's the opportunity to drop them, for exactly the same reason as test files: they must exist only if `islast=true`.

server/sonar-db-dao/src/test/java/org/sonar/db/purge/PurgeDaoTest.java
```diff
@@ -132,7 +134,7 @@ public void should_purge_inactive_short_living_branches() {
   @Test
   public void shouldDeleteHistoricalDataOfDirectoriesAndFiles() {
     dbTester.prepareDbUnit(getClass(), "shouldDeleteHistoricalDataOfDirectoriesAndFiles.xml");
-    PurgeConfiguration conf = new PurgeConfiguration(new IdUuidPair(THE_PROJECT_ID, "ABCD"), new String[] {Scopes.DIRECTORY, Scopes.FILE},
+    PurgeConfiguration conf = new PurgeConfiguration(new IdUuidPair(THE_PROJECT_ID, "ABCD"), asList(Qualifiers.DIRECTORY, Scopes.FILE),
```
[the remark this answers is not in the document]
simonbrandhof: Bad usage of Scopes instead of Qualifiers. However this test does not verify that configuration is correct. The test provides its own list of qualifiers to be purged.

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v671/DeleteHistoryOfTestFileMeasures.java
simonbrandhof: Not sure, see the inner join on `snapshots`. Moreover tests check that only the past file measures are deleted.

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v671/DeleteHistoryOfTestFileMeasures.java
bartfastiel: I would suggest to rename the step then. (However, since it is just naming, this is low priority)

server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v671/DeleteHistoryOfTestFileMeasures.java
bartfastiel: Actually I am quite sure, that this would delete all measures.

I interpret it as "delete all measures, if there is at least one test or unit test measure on an old analysis".

server/sonar-db-dao/src/main/java/org/sonar/db/purge/PurgeConfiguration.java
```diff
@@ -50,19 +52,19 @@ public PurgeConfiguration(IdUuidPair rootProjectId, String[] scopesWithoutHistor
   }
 
   public static PurgeConfiguration newDefaultPurgeConfiguration(Configuration config, IdUuidPair idUuidPair, Collection<String> disabledComponentUuids) {
-    String[] scopes = new String[] {Scopes.FILE};
+    Collection<String> qualifiersWithoutHistoricalData = asList(Qualifiers.FILE, Qualifiers.UNIT_TEST_FILE);
```
sns-seb: changes to this method must be rollbacked to define scopes which are actually scopes and not qualifiers in the default PurgeConfiguration

server/sonar-db-dao/src/test/java/org/sonar/db/component/ComponentDaoTest.java
```diff
@@ -1281,18 +1282,18 @@ public void selectAncestors() {
...
     ComponentDto module = newModuleDto(MODULE_UUID, project);
     db.components().insertComponent(module);
-    ComponentDto file1 = newFileDto(project, null, FILE_1_UUID).setDbKey("file-key-1").setName("File One");
-    db.components().insertComponent(file1);
-    ComponentDto file2 = newFileDto(module, null, FILE_2_UUID).setDbKey("file-key-2").setName("File Two");
-    db.components().insertComponent(file2);
-    ComponentDto file3 = newFileDto(module, null, FILE_3_UUID).setDbKey("file-key-3").setName("File Three");
-    db.components().insertComponent(file3);
+    ComponentDto fileInProject = newFileDto(project, null, FILE_1_UUID).setDbKey("file-key-1").setName("File One");
+    db.components().insertComponent(fileInProject);
```
julienlancelot: This 2 lines could have been inlined : 
`ComponentDto fileInProject = db.components().insertComponent(newFileDto(project, null, FILE_1_UUID).setDbKey("file-key-1").setName("File One"));`
Same for 2 following variables

server/sonar-db-dao/src/test/java/org/sonar/db/purge/PurgeDaoTest.java
```diff
@@ -54,6 +55,7 @@
 import org.sonar.db.rule.RuleDefinitionDto;
 
 import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
```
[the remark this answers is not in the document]
julienlancelot: Good point from Daniel, if I'm not wrong I don't see it ?

server/sonar-db-dao/src/main/java/org/sonar/db/purge/PurgeConfiguration.java
```diff
@@ -30,16 +30,19 @@
...
 import org.sonar.core.config.PurgeConstants;
 
+import static java.util.Arrays.asList;
+import static java.util.Collections.singletonList;
+
 public class PurgeConfiguration {
 
   private final IdUuidPair rootProjectIdUuid;
-  private final String[] scopesWithoutHistoricalData;
+  private final Collection<String> scopesWithoutHistoricalData;
```
sns-seb: a Set would be more appropriate than collection. We do care that there is no duplicate values (it doesn't make sense) but indeed order doesn't matter

simonbrandhof: Exact. Anyway this code is going to be dropped soon: https://jira.sonarsource.com/browse/SONAR-10180

PR #2859, opened 2017-11-30
SONAR-10067 an organization admin can manage project permissions

teryk:


teryk:
functional solution changed

PR #2867, opened 2017-12-04
SONAR-10037 Allow creation of Quality Gate by organization

ehartmann:


server/sonar-server/src/test/java/org/sonar/server/qualitygate/ws/CreateActionTest.java
```diff
@@ -59,26 +62,55 @@
...
     userSession.logIn();
 
     expectedException.expect(ForbiddenException.class);
     expectedException.expectMessage("Insufficient privileges");
 
-    executeRequest("Default");
+    executeRequest(Optional.empty(), "Default");
   }
 
   @Test
```
julienlancelot: This test should be renamed to "throw_ForbiddenException_if_not_gate_administrator" => default organization should not be mentioned as there's nothing specific to default org

server/sonar-server/src/test/java/org/sonar/server/qualitygate/ws/CreateActionTest.java
```diff
@@ -91,7 +123,7 @@ public void throw_ForbiddenException_if_not_gate_administrator_of_default_organi
     expectedException.expect(ForbiddenException.class);
     expectedException.expectMessage("Insufficient privileges");
 
-    executeRequest("Default");
+    executeRequest(Optional.empty(), "Default");
   }
```
julienlancelot: A test is missing when a query is executed on an unknown organization

server/sonar-server/src/test/java/org/sonar/server/qualitygate/ws/CreateActionTest.java
```diff
@@ -104,14 +136,20 @@ public void test_ws_definition() {
     assertThat(action.params()).hasSize(2);
   }
```
julienlancelot: A test is missing when trying to create an quality gate having the same name as another one from the same organization => ko
And another one when trying to use a name alredy used, but in another org => ok

julienlancelot:
Please add missing UTs, otherwise LGTM

PR #2879, opened 2017-12-05
SONAR-10123 UserSession#keepkeepAuthorizedComponents handles branches properly

teryk:


server/sonar-server/src/main/java/org/sonar/server/user/ServerUserSession.java
```diff
@@ -61,7 +61,7 @@
   private Map<String, Set<String>> permissionsByProjectUuid;
 
   ServerUserSession(DbClient dbClient, OrganizationFlags organizationFlags,
-    DefaultOrganizationProvider defaultOrganizationProvider, @Nullable UserDto userDto) {
+                    DefaultOrganizationProvider defaultOrganizationProvider, @Nullable UserDto userDto) {
```
julienlancelot: Wrong indent

PR #2905, opened 2017-12-13
SONAR-10184 Add stub for homepage fields to api/users/current

gjambet:


sonar-ws/src/main/protobuf/ws-users.proto
```diff
@@ -107,10 +107,16 @@ message CurrentWsResponse {
...
   optional bool showOnboardingTutorial = 11;
   optional string avatar = 12;
+  optional Homepage homepage = 13;
 
   message Permissions {
     repeated string global = 1;
   }
+
+  message Homepage {
+    optional string type = 1;
```
teryk: suggestion: you could use an enum for the Homepage type

gjambet: done

server/sonar-server/src/main/resources/org/sonar/server/user/ws/current-example.json
```diff
@@ -12,5 +12,9 @@
   "groups": ["Jedi", "Rebel"],
   "permissions": {
     "global": ["profileadmin", "scan"]
+  },
+  "homepage": {
+    "type": "PROJECT",
```
teryk: to remember: the value in parameters and in response should be the same (lowercase or uppercase)

PR #2913, opened 2017-12-15
update typescript and prettier, use shortcut fragment syntax

stas-vilchik:


server/sonar-web/tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "noUnusedParameters": true,
     "noImplicitAny": true,
     "strict": true,
+    "strictFunctionTypes": false,
```
stas-vilchik: This is a new option in TS 2.6. If I leave it `true`, it raises quite some issues around `react-select`. I don't want to waste time on this currently, but it would be nice to fix it one day.

PR #2919, opened 2017-12-19
SONAR-8216 System info page fails when a setting is defined both in sonar.properties and in DB

simonbrandhof:


server/sonar-server/src/main/java/org/sonar/server/setting/DatabaseSettingLoader.java
```diff
@@ -44,10 +46,12 @@ public String load(String key) {
...
 
   @Override
-  public void loadAll(ImmutableMap.Builder<String, String> appendTo) {
+  public Map<String,String> loadAll() {
     try (DbSession dbSession = dbClient.openSession(false)) {
+      Map<String,String> result = new HashMap<>();
       dbClient.propertiesDao().selectGlobalProperties(dbSession)
-        .forEach(p -> appendTo.put(p.getKey(), defaultString(p.getValue())));
+        .forEach(p -> result.put(p.getKey(), defaultString(p.getValue())));
+      return Collections.unmodifiableMap(result);
```
julienlancelot: You simplify this code like this (it will also return an immutable map) : 
```
      return dbClient.propertiesDao().selectGlobalProperties(dbSession)
        .stream()
        .collect(MoreCollectors.uniqueIndex(PropertyDto::getKey, p -> defaultString(p.getValue())));
```

PR #2932, opened 2017-12-28
SONAR-10217 Fix 500 error when branch is incorrect

ehartmann:


server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionTest.java
```diff
@@ -151,6 +151,19 @@ public void test_definition() {
       "Portfolios are not supported. If this parameter is set, 'projects' must not be set.");
   }
 
+  //SONAR-10217
```
teryk: remark: the reference to the Jira ticket is not necessary

server/sonar-server/src/test/java/org/sonar/server/issue/ws/SearchActionTest.java
```diff
@@ -151,6 +151,19 @@ public void test_definition() {
...
+  @Test
+  public void test_with_incorrect_branch() throws Exception {
```
teryk: remark: prefix with `test` is not needed. I'd name the test: `empty_search_with_unknown_branch`

PR #2934, opened 2017-12-29
SONAR-10154 No streaming of web services response by default

teryk:


server/sonar-server/src/main/java/org/sonar/server/platform/ws/InfoAction.java
```diff
@@ -52,19 +47,16 @@ public void define(WebService.NewController controller) {
       .setInternal(true)
       .setResponseExample(getClass().getResource("/org/sonar/server/platform/ws/info-example.json"))
```
julienlancelot: Please update the url by removing "/org/sonar/server/platform/ws/"

PR #2943, opened 2018-01-05
SONAR-10224 Revert backward compatible version of QGChangeEventListener

janos-ss:


server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/analysis/Organization.java
```diff
@@ -96,6 +96,10 @@ public OrganizationDto toDto() {
       .setDefaultQualityGateUuid(defaultQualityGateUuid);
   }
 
+  public static Organization from(OrganizationDto organizationDto) {
```
sns-seb: @janos-ss  I suggest to make this pull request being based on your work to remove the `organizationsEnabled` flag from the `Organization` class so that this change is not required. WDYT?

janos-ss: Yeah I removed this change from this PR so it can be about SONAR-10224 and ready to merge. I'll open a new PR for for the org-enabled fix. Thanks for the review.

PR #2945, opened 2018-01-08
Relocate organizations enabled flag

janos-ss:


server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/step/LoadReportAnalysisMetadataHolderStep.java
```diff
@@ -86,12 +87,19 @@ public void execute() {
...
     loadMetadata(reportMetadata);
+    loadOrganizationsEnabled();
     Organization organization = loadOrganization(reportMetadata);
     loadProject(reportMetadata, organization);
     loadQualityProfiles(reportMetadata, organization);
     branchLoader.load(reportMetadata);
   }
 
+  private void loadOrganizationsEnabled() {
+    try (DbSession dbSession = dbClient.openSession(false)) {
```
sns-seb: there is no reason to open a DB connection twice in a row. Even with a connection pool, that will uselessly consume extra resources.
you should refactor and use the same DB session in `loadOrganizationsEnabled` and `loadMetadata`

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/step/LoadReportAnalysisMetadataHolderStep.java
```diff
@@ -112,10 +113,13 @@ private void loadProject(ScannerReport.Metadata reportMetadata, Organization org
   }
 
   private Organization loadOrganization(ScannerReport.Metadata reportMetadata) {
-    Organization organization = toOrganization(ceTask.getOrganizationUuid());
-    checkOrganizationKeyConsistency(reportMetadata, organization);
-    analysisMetadata.setOrganization(organization);
-    return organization;
+    try (DbSession dbSession = dbClient.openSession(false)) {
+      analysisMetadata.setOrganizationsEnabled(organizationFlags.isEnabled(dbSession));
```
sns-seb: minor: I would not call `organizationFlags.isEnabled` before `checkOrganizationKeyConsistency`, if the check fails, we uselessly made a call to DB

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/step/LoadReportAnalysisMetadataHolderStep.java
```diff
@@ -204,12 +208,11 @@ private static boolean reportBelongsToDefaultOrganization(@Nullable String organ
...
   }
 
-  private Organization toOrganization(String organizationUuid) {
-    try (DbSession dbSession = dbClient.openSession(false)) {
-      Optional<OrganizationDto> organizationDto = dbClient.organizationDao().selectByUuid(dbSession, organizationUuid);
-      checkState(organizationDto.isPresent(), "Organization with uuid '%s' can't be found", organizationUuid);
-      return Organization.from(organizationDto.get(), organizationFlags.isEnabled(dbSession));
-    }
+  private Organization toOrganization(DbSession dbSession, String organizationUuid) {
+    analysisMetadata.setOrganizationsEnabled(organizationFlags.isEnabled(dbSession));
```
sns-seb: `analysisMetadata.setOrganizationsEnabled` is already called on line 117 and it really does not belong to a method called `toOrganization` which really should be static

PR #2948, opened 2018-01-09
Hardening Jan 9

stas-vilchik:


server/sonar-web/src/main/js/apps/component-measures/style.css
```diff
@@ -120,6 +120,18 @@
   margin-right: 0;
 }
 
+.measure-details-viewer .issue-list {
+  /* no math, just a good guess */
+  min-width: 640px;
```
gregaubert: I think 620px would be better, so we always see the full box.

PR #2961, opened 2018-01-15
Hardening jan 12

gregaubert:


server/sonar-web/src/main/js/apps/system/components/PageActions.tsx
```diff
@@ -139,7 +145,9 @@ export default class PageActions extends React.PureComponent<Props, State> {
           id="download-link"
           className="button spacer-left"
           onClick={this.removeElementFocus}
-          download="sonarqube_system_info.json"
+          download={`sonarqube-support-info-${
+            this.props.serverId ? this.props.serverId + '-' : ''
+          }${date}.json`}
```
stas-vilchik: this is not formatted well by prettier, could you try to extract the value to a variable maybe?

PR #2974, opened 2018-01-18
SONAR-10266 Add ITs for project badges ws

julienlancelot:


tests/src/test/java/org/sonarqube/tests/project/ProjectBadgesTest.java
```diff
@@ -30,13 +30,21 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
 import static util.ItUtils.projectDir;
 
 public class ProjectBadgesTest {
 
+  private static final String PROJECT_KEY = "sample";
+  private static final String WS_MEASURE_BADGES_ON_QUALITY_GATE = "api/project_badges/measure?project=" + PROJECT_KEY + "&metric=alert_status";
+  private static final String WS_MEASURE_BADGES_ON_BUGS = "api/project_badges/measure?project=" + PROJECT_KEY + "&metric=bugs";
+  private static final String WS_QUALITY_GATE_BADGE = "api/project_badges/quality_gate?project=" + PROJECT_KEY;
+  private static final String SONAR_CLOUD_WARN_BADGE = "images/project_badges/sonarcloud-orange.svg";
```
gregaubert: I would just call it ORANGE instead of WARN, because it's not a warning badge, just a different background color to match any targeted website.

julienlancelot: Thanks, good catch, I've fixed it !

PR #2980, opened 2018-01-24
SONAR-10117 minimize risk of deadlock on MySQL

simonbrandhof:


simonbrandhof:
Exact, the root problem is still unknown for the time being. Investigating how MySQL indices work would help. 
For sure PersistLiveMeasuresStep should not use a single transaction to insert thousands of rows. This PR is just a quick win.

PR #2990, opened 2018-01-26
Clean-up in sonar-server tests

simonbrandhof:
Removing transitive dependency on securemock allows to fix usage of Mockito

### DIFF
--- a/server/sonar-server/pom.xml
+++ b/server/sonar-server/pom.xml
@@ -176,6 +176,10 @@
           <groupId>com.carrotsearch.randomizedtesting</groupId>
           <artifactId>randomizedtesting-runner</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.elasticsearch</groupId>
+          <artifactId>securemock</artifactId>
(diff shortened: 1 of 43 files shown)


server/sonar-server/pom.xml
```diff
@@ -176,6 +176,10 @@
           <groupId>com.carrotsearch.randomizedtesting</groupId>
           <artifactId>randomizedtesting-runner</artifactId>
         </exclusion>
+        <exclusion>
```
sns-seb: given the title of the first commit, I would expect to see only the change on `server/sonar-server/pom.xml` in it
It seems the other changes belong to the last commit

simonbrandhof: it's on purpose. I agree that the commit title can be misleading. securemock overrode mockito in some cases like:
- signature of ArgumentMatcher implementations (Object -> T). That's the reason why they have been fixed and replaced by lambda
- any(Class) supported null values in securemock but not in latest version of mockito. Replacing calls by any() fixes the issue.

PR #2993, opened 2018-01-29
drop getCurrentUserFromStore

stas-vilchik:


server/sonar-web/src/main/js/components/issue/popups/SetAssigneePopup.js
```diff
@@ -155,3 +154,10 @@ export default class SetAssigneePopup extends React.PureComponent {
     );
   }
 }
+
+const mapStateToProps = state => ({
+  currentUser: getCurrentUser(state),
+  organizationEnabled: areThereCustomOrganizations(state)
```
gregaubert: We can get rid of this and use the `organizationsEnabled` property that is in the context.

PR #2995, opened 2018-01-29
SONAR-10319 Display Analysis Scope on rules detail page

gregaubert:


gregaubert:
@stas-vilchik Thanks for the fix! :+1:

PR #3019, opened 2018-02-05
Support pull request analysis

janos-ss:


sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/ProjectReactorValidator.java
```diff
@@ -74,6 +74,7 @@ public void validate(ProjectReactor reactor) {
       branchParamsValidator.validate(validationMessages, deprecatedBranchName);
     } else {
       validateBranchParamsWhenPluginAbsent(validationMessages);
+      validatePullRequestParamsWhenPluginAbsent(validationMessages);
```
teryk: please add a test similar to `org.sonar.scanner.scan.ProjectReactorValidatorTest#fail_when_branch_name_is_specified_but_branch_plugin_not_present` and `org.sonar.scanner.scan.ProjectReactorValidatorTest#fail_when_branch_target_is_specified_but_branch_plugin_not_present`

janos-ss: Good catch, thanks! (+1)

PR #3030, opened 2018-02-07
SONAR-10410 Rule#getSystemTags is populated with definition tags

sns-seb:


simonbrandhof:
The ticket is associated to 6.7.2, so the PR should be based with branch-6.7.

server/sonar-server/src/main/java/org/sonar/server/rule/CachingRuleFinder.java
```diff
@@ -174,7 +174,8 @@ private static Rule toRule(RuleDefinitionDto ruleDefinition, List<RuleParamDto>
       .setRepositoryKey(ruleDefinition.getRepositoryKey())
       .setSeverity(severity != null ? RulePriority.valueOf(severity) : null)
       .setStatus(ruleDefinition.getStatus().name())
-      .setTags(ruleDefinition.getSystemTags().toArray(new String[ruleDefinition.getSystemTags().size()]))
+      .setSystemTags(ruleDefinition.getSystemTags().toArray(new String[ruleDefinition.getSystemTags().size()]))
+      .setTags(new String[0])
```
simonbrandhof: I think that this is a change of behavior. To keep backward-compatibility what about referencing system tags in both `tags` and `systemTags` ?

sns-seb: I'll keep the backward compatibilty in LTS but fix it in 7.1

PR #3031, opened 2018-02-07
SONAR-10310 Fix issue when moving a deprecated key from one rule to another

ehartmann:


server/sonar-server/src/main/java/org/sonar/server/rule/RegisterRules.java
```diff
@@ -203,15 +203,16 @@ private static boolean noTemplateRuleWithOrganizationsEnabled(RegisterRulesConte
...
-      this.dbDeprecatedKeysById = dbDeprecatedKeysById;
-      this.dbRulesByDbDeprecatedKey = buildDbRulesByDbDeprecatedKey(dbDeprecatedKeysById, dbRules);
+      this.dbDeprecatedKeys = dbDeprecatedKeys;
+      this.dbRulesByDbDeprecatedKey = buildDbRulesByDbDeprecatedKey(dbDeprecatedKeys, dbRules);
     }
 
-    private static Map<RuleKey, RuleDefinitionDto> buildDbRulesByDbDeprecatedKey(Map<Integer, Set<SingleDeprecatedRuleKey>> dbDeprecatedKeysById,
-      Map<RuleKey, RuleDefinitionDto> dbRules) {
+    private static Map<RuleKey, RuleDefinitionDto> buildDbRulesByDbDeprecatedKey(Set<SingleDeprecatedRuleKey> dbDeprecatedKeys, Map<RuleKey, RuleDefinitionDto> dbRules) {
+      Map<Integer, Set<SingleDeprecatedRuleKey>> dbDeprecatedKeysById = dbDeprecatedKeys.stream()
```
sns-seb: this map is not required to build rulesByDbDeprecatedKey. 
Loop at line 220 can easily be changed to work directly with the `Set<SingleDeprecatedRuleKey>`

server/sonar-server/src/test/java/org/sonar/server/rule/RegisterRulesTest.java
```diff
@@ -842,6 +842,50 @@ public void declaring_a_rule_with_a_deprecated_RuleKey_still_used_should_throw_I
     });
   }
 
+  @Test
+  public void updating_the_parent_of_a_deprecated_rule_key_should_work() {
+    execute(context -> {
+      NewRepository repo = context.createRepository("javascript", "javascript");
+      repo.createRule("linelength")
+        .setName("s103")
```
sns-seb: "linelength" ?

server/sonar-server/src/test/java/org/sonar/server/rule/RegisterRulesTest.java
```diff
@@ -842,6 +842,50 @@ public void declaring_a_rule_with_a_deprecated_RuleKey_still_used_should_throw_I
     });
   }
 
+  @Test
+  public void updating_the_parent_of_a_deprecated_rule_key_should_work() {
```
sns-seb: it would help to identify what's changing from one execute to another if, for example, you used some methods to hide the repetitive/noise code from the one that matters and some variables/constants

PR #3035, opened 2018-02-08
SONAR-10395 Improve time compressing plugins in SonarCloud

dbmeneses:


server/sonar-server/src/main/java/org/sonar/server/plugins/PluginCompression.java
```diff
@@ -50,12 +50,26 @@ public PluginCompression(Configuration configuration) {
...
+      Path sourcePack200Path = sourceDir.resolve(targetPack200Path.getFileName());
+      
+      // check if packed file was deployed alongside the jar. If that's the case, use it instead of generating it (SONAR-10395).
+      if (Files.isRegularFile(sourcePack200Path)) {
+        try {
+          LOG.debug("Found pack200: " + sourcePack200Path);
+          Files.copy(sourcePack200Path, targetPack200Path);
+        } catch (IOException e) {
+          LOG.error("Failed to copy pack200 file from " + sourcePack200Path + " to " + targetPack200Path);
+          return;
```
ehartmann: I think it would be better either : 
* to pack200 the original here, instead of returning,
* to throw an ISE (as pack200 method is doing)

Because the file will be missing from compressedPlugins map, so I expect some weird issues in this case.
So throwing an exception to prevent SQ from starting maybe the best solution.

server/sonar-server/src/main/java/org/sonar/server/plugins/PluginCompression.java
```diff
@@ -50,12 +50,26 @@ public PluginCompression(Configuration configuration) {
...
+      Path targetPack200Path = FileUtils.getPack200FilePath(targetJarFile);
+      Path sourcePack200Path = sourceDir.resolve(targetPack200Path.getFileName());
+      
+      // check if packed file was deployed alongside the jar. If that's the case, use it instead of generating it (SONAR-10395).
+      if (Files.isRegularFile(sourcePack200Path)) {
+        try {
+          LOG.debug("Found pack200: " + sourcePack200Path);
+          Files.copy(sourcePack200Path, targetPack200Path);
+        } catch (IOException e) {
+          LOG.error("Failed to copy pack200 file from " + sourcePack200Path + " to " + targetPack200Path);
```
ehartmann: I do think we should log the exception to be able to know the root cause.

server/sonar-server/src/main/java/org/sonar/server/plugins/PluginCompression.java
```diff
@@ -50,12 +50,25 @@ public PluginCompression(Configuration configuration) {
...
     if (configuration.getBoolean(PROPERTY_PLUGIN_COMPRESSION_ENABLE).orElse(false)) {
-      Path pack200Path = FileUtils.getPack200FilePath(jarFile);
-      pack200(jarFile, pack200Path, pluginKey);
-      String hash = calculateMd5(pack200Path);
-      RemotePluginFile compressedPlugin = new RemotePluginFile(pack200Path.getFileName().toString(), hash);
+      Path targetPack200Path = FileUtils.getPack200FilePath(targetJarFile);
+      Path sourcePack200Path = sourceDir.resolve(targetPack200Path.getFileName());
+
+      // check if packed file was deployed alongside the jar. If that's the case, use it instead of generating it (SONAR-10395).
+      if (Files.isRegularFile(sourcePack200Path)) {
```
SonarTech: Code Smell Code Smell: Replace this with a call to the "toFile().isFile()" method ([squid:S3725](https://next.sonarqube.com/sonarqube/coding_rules#rule_key=squid%3AS3725))

[See it in SonarQube](https://next.sonarqube.com/sonarqube/project/issues?id=org.sonarsource.sonarqube%3Asonarqube&branch=feature%2Fcompress_plugins&issues=AWF1-O5v0orRW2HVTxyO&open=AWF1-O5v0orRW2HVTxyO)

[ISSUE_KEY]: AWF1-O5v0orRW2HVTxyO

PR #3044, opened 2018-02-09
6.7 SONAR-10301 delete from CE child tables when deleting project

sns-seb:


sns-seb:
Closed without merging as this fix won't be included in 6.7.2
see #3047

PR #3056, opened 2018-02-13
SONAR-10313 fix error when displaying rule facet

sns-seb:


simonbrandhof:
An integration test could be added to verify that the rules facet is correctly loaded :-)

server/sonar-server/src/main/java/org/sonar/server/issue/ws/SearchAction.java
```diff
@@ -384,35 +384,59 @@ private SearchWsResponse doHandle(SearchRequest request, Request wsRequest) {
...
-    if (facets == null || facets.get(PARAM_RULES) == null) {
+    if (facets == null) {
+      return;
+    }
+    LinkedHashMap<String, Long> rulesFacet = facets.get(PARAM_RULES);
+    if (rulesFacet == null) {
       return;
     }
 
     // The facet for PARAM_RULES contains the id of the rule as the key
```
ehartmann: Should be before 390

PR #3057, opened 2018-02-13
SONAR-10409 Update too long keys of permission templates

ehartmann:


server/sonar-db-migration/src/main/java/org/sonar/server/platform/db/migration/version/v71/UpdatePermissionTooLongTemplateKeys.java
```diff
@@ -0,0 +1,67 @@
...
+      lengthFunction = "len";
+    } else {
+      lengthFunction = "length";
+    }
+  }
+
+  @Override
+  protected void execute(Context context) throws SQLException {
+    MassUpdate massUpdate = context.prepareMassUpdate();
+    massUpdate.select("select kee from permission_templates where " + lengthFunction + "(kee) > " + VarcharColumnDef.UUID_SIZE);
```
sns-seb: select return a `SqlStatement` on which you can call `setLong` to safely set the size to compare to

server/sonar-db-migration/src/test/java/org/sonar/server/platform/db/migration/version/v71/UpdatePermissionTooLongTemplateKeysTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+import static java.util.stream.Collectors.toList;
+import static org.apache.commons.lang.RandomStringUtils.randomAlphanumeric;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class UpdatePermissionTooLongTemplateKeysTest {
+  @Rule
+  public CoreDbTester dbTester = CoreDbTester.createForSchema(UpdatePermissionTooLongTemplateKeysTest.class, "templates.sql");
+
+  private static final RecordingUuidFactory UUID_FACTORY = new RecordingUuidFactory();
```
sns-seb: its not safe to have a stateful UUIDFactory static

Junit create a instance of the test class for each method annotated with `@Test` and could run them in parallel

PR #3070, opened 2018-02-15
rewrite parts of permission templates app in react

stas-vilchik:


server/sonar-web/src/main/js/apps/projectsManagement/Projects.tsx
```diff
@@ -35,7 +34,13 @@ interface Props {
   selection: string[];
 }
 
-export default class Projects extends React.PureComponent<Props> {
+interface State {
+  applyTemplate: boolean;
+}
+
+export default class Projects extends React.PureComponent<Props, State> {
+  state: State = { applyTemplate: false };
```
gregaubert: If I'm not mistaken `applyTemplate` is not used right ?

server/sonar-web/src/main/js/apps/permissions/project/components/PageHeader.tsx
```diff
@@ -17,43 +17,48 @@
...
-    e.preventDefault();
-    e.target.blur();
-    const { component, loadHolders } = this.props;
-    const organization = component.organization ? { key: component.organization } : null;
-    new ApplyTemplateView({ project: component, organization })
-      .on('done', () => loadHolders())
-      .render();
+  handleApplyTemplate = (event: React.SyntheticEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+    event.currentTarget.blur();
```
gregaubert: I think those 2 lines can be removed

server/sonar-web/src/main/js/apps/permission-templates/components/ActionsCell.tsx
```diff
@@ -17,58 +17,81 @@
...
+      this.setState({ updateModal: false });
+    }
+  };
+
+  handleSubmitUpdateModal = (data: {
+    description: string;
+    name: string;
+    projectKeyPattern: string;
+  }) => {
+    return updatePermissionTemplate({ id: this.props.permissionTemplate.id, ...data }).then(() => {
```
gregaubert: I think you can put directly `.then(this.props.refresh)`

server/sonar-web/src/main/js/apps/permissions/project/components/PageHeader.tsx
stas-vilchik: I try to always prevent the default behavior (just in case) and remove the focus from a link/button

gregaubert: mmh I think we are not very consistant on this then. We are for links but not for buttons. We can find a lot of buttons handler with and without the default behavior.
Let's say we always prevent it from now on.
Then maybe we should have a button component that always does it ?

stas-vilchik: good idea

PR #3071, opened 2018-02-15
SONAR-10420 Fail at startup when cluster enabled on MySQL

ehartmann:


server/sonar-server/src/main/java/org/sonar/server/startup/ClusterConfigurationCheck.java
```diff
@@ -0,0 +1,50 @@
...
+  private final Database database;
+
+  public ClusterConfigurationCheck(Database database) {
+    this.database = database;
+  }
+
+  @Override
+  public void start() {
+    if (MySql.ID.equals(database.getDialect().getId())) {
+      throw new IllegalStateException("When cluster is activater, the MySQL database is not supported");
```
sns-seb: "activer" => "activated"

I'd rather go with an error message "MySQL is not supported as database to run SonarQube DataCenter Edition"

Please check with PMs (@ganncamp ?)

PR #3077, opened 2018-02-15
SONAR-10323 Fix permission on api/project_branches/list

ehartmann:


server/sonar-server/src/main/java/org/sonar/server/branch/ws/ListAction.java
```diff
@@ -80,7 +82,7 @@ public void define(WebService.NewController context) {
     WebService.NewAction action = context.createAction(ACTION_LIST)
       .setSince("6.6")
       .setDescription("List the branches of a project.<br/>" +
-        "Requires 'Administer' rights on the specified project.")
+        "Requires 'Browse on the project' or 'Execute analysis' rights on the specified project.")
```
simonbrandhof: Same question as in other PR.

PR #3081, opened 2018-02-18
Put ES javaOpts arguments on one line

M1dgard:
Make it consistent with the two other javaOpts settings, avoiding errors due to only uncommenting one of the three lines.

Without exaggeration I spent three hours troubleshooting due to having uncommented only the top line.


simonbrandhof:
Thanks for the contribution.

PR #3082, opened 2018-02-19
SONAR-10433 Store pull request in projects table

teryk:


server/sonar-db-dao/src/test/java/org/sonar/db/component/ComponentTesting.java
```diff
@@ -231,6 +240,7 @@ public static ComponentDto newProjectBranch(ComponentDto project, BranchDto bran
     checkArgument(project.qualifier().equals(Qualifiers.PROJECT));
     checkArgument(project.getMainBranchProjectUuid() == null);
     String branchName = branchDto.getKey();
+    String branchSeparator = branchDto.getBranchType() == PULL_REQUEST ? ":PULL_REQUEST:" : ":BRANCH:";
```
janos-ss: Why not use the existing constants instead of hard-coding the separator strings?

teryk:
because we keep encapsulation of this implementation detail.
ComponentTesting is used only for tests

Teryk Bellahsene | SonarSource
Software Developer
http://sonarsource.com

server/sonar-db-dao/src/test/java/org/sonar/db/component/ComponentTesting.java
teryk: because we keep encapsulation of this implementation detail. ComponentTesting is used only for tests

PR #3087, opened 2018-02-20
create and use Button component

stas-vilchik:
As a follow-up of our conversation about creating a `Button` component to encapsulate `onClick` logic, I've created this PR. I replaced all usages of a regular `<button />` element, as well as "cancel" links in modals windows.


server/sonar-web/src/main/js/components/ui/buttons.css
```diff
@@ -29,28 +32,28 @@
   color: inherit;
 }
 
-.button-icon.button-small {
+.button-icon.button-icon.button-small {
```
gregaubert: I'm not sure this is intended ? The double `.button-icon.button-icon`

PR #3090, opened 2018-02-21
MMF-1185 SONAR-10426 add homepage portfolios portfolio and application on sonarqube

gjambet:


gjambet:
Only last commit is relevant, do not merge

PR #3094, opened 2018-02-21
SONAR-10365 Do not store pr id in analysis properties

janos-ss:


teryk:
if my understanding is correct this PR has already been merged in another PR.

PR #3100, opened 2018-02-22
Hardening 22.02.2018

gregaubert:


server/sonar-web/src/main/js/components/common/MultiSelect.tsx
```diff
@@ -23,21 +23,22 @@ import MultiSelectOption from './MultiSelectOption';
 import SearchBox from '../controls/SearchBox';
 
 interface Props {
-  selectedElements: Array<string>;
-  elements: Array<string>;
+  elements: string[];
   listSize?: number;
+  loading?: boolean;
```
stas-vilchik: I wonder if it'd better to have `loading` inside the component's state. If we change `onSearch` prop to return a Promise `onSearch: (query: string) => Promise<void>`, we could use this promise to change the `loading` flag. WDYT?

server/sonar-web/src/main/js/apps/marketplace/components/LicenseEditionSet.tsx
```diff
@@ -114,42 +117,55 @@ export default class LicenseEditionSet extends React.PureComponent<Props, State>
...
-    this.setState({ acceptTerms: checked }, () =>
-      this.updateLicense(this.state.license, this.state.licenseEdition, this.state.previewStatus)
-    );
+  handleTermsCheck = (checked: boolean) => {
+    this.setLicense({ acceptTerms: checked });
+  };
 
-  updateLicense = (license: string, licenseEdition?: Edition, previewStatus?: string) => {
-    this.setState({ license, licenseEdition, loading: false, previewStatus });
+  resetLicense = (state = {}) => {
```
stas-vilchik: When you use a default value `{}`, you lose the type-checking, i.e. you can pass basically anything to this function. Instead always type such patterns with `state: Partial<State> = {}`.

server/sonar-web/src/main/js/apps/coding-rules/components/RuleDetailsMeta.tsx
```diff
@@ -125,6 +125,7 @@ export default class RuleDetailsMeta extends React.PureComponent<Props, State> {
         ) : (
           <TagsList
             allowUpdate={canWrite}
+            className="display-flex-center"
```
stas-vilchik: could you also fix the display in the workspace as well please?

image

server/sonar-web/src/main/js/components/controls/SearchBox.tsx
```diff
@@ -145,7 +147,9 @@ export default class SearchBox extends React.PureComponent<Props, State> {
           value={value}
         />
 
-        <SearchIcon className="search-box-magnifier" />
+        <DeferredSpinner loading={loading !== undefined ? loading : false}>
```
stas-vilchik: the "x" icon is no more aligned on (at least) Chrome:

image

server/sonar-web/src/main/js/components/common/MultiSelect.tsx
gregaubert: Yes I think it would be, the only thing is the initial value of `loading`, since we receive the elements from outside, the initial `onSearch` is not called from inside that component currently

server/sonar-web/src/main/js/apps/marketplace/components/LicenseEditionSet.tsx
gregaubert: We can't use the spread operator but at least this works :
```TS
resetLicense<K extends keyof State>(state: Pick<State, K>) {
    this.setLicense(
      Object.assign(
        {
          license: '',
          licenseEdition: undefined,
          previewStatus: undefined,
          wrongEdition: false
        },
        state
      )
    );
  }
```

server/sonar-web/src/main/js/apps/coding-rules/components/RuleDetailsMeta.tsx
gregaubert: Ah I didn't see this, I'll fix it

PR #3104, opened 2018-02-27
SONAR-10345 Ensure that no orphaned deliveries remains when deleting …

gjambet:
…webhooks.


server/sonar-db-dao/src/main/java/org/sonar/db/webhook/WebhookDao.java
```diff
@@ -28,13 +29,16 @@
...
 import static com.google.common.base.Preconditions.checkState;
+import static java.util.Collections.emptyList;
 
 public class WebhookDao implements Dao {
 
   private final System2 system2;
+  private final WebhookDeliveryDao webhookDeliveryDao;
 
-  public WebhookDao(System2 system2) {
+  public WebhookDao(System2 system2, WebhookDeliveryDao webhookDeliveryDao) {
```
simonbrandhof: In our current state a DAO should not depend on another DAO. Callers of delete() methods should also call WebhookDeliveryDao#delete() methods.

simonbrandhof:
Be careful with dogfooding. Let's discuss that offline.

PR #3107, opened 2018-02-27
Use char notation to represent a character to improve performance.

reudismam:


janos-ss:
I challenge the usefulness of this request. This looks like a pure pedantic change, with insignificant expected benefits if any at all, depending on the JVM implementation. Do correct me if I'm wrong. I close this now until further notice.

PR #3116, opened 2018-03-05
SONAR-10450 Add a 'submitter' column in Background Tasks page

najibdarma:


simonbrandhof:
Apparently the PR is not correctly rebased with master. It should not contain so many commits and should not remove some Java files.

PR #3122, opened 2018-03-06
SONAR-10461 load branch even if project key in report is wrong

sns-seb:


server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/step/LoadReportAnalysisMetadataHolderStep.java
```diff
@@ -88,9 +88,10 @@ public void execute() {
 
     loadMetadata(reportMetadata);
     Organization organization = loadOrganization(reportMetadata);
-    loadProject(reportMetadata, organization);
+    Runnable projectValidation = loadProject(reportMetadata, organization);
     loadQualityProfiles(reportMetadata, organization);
     branchLoader.load(reportMetadata);
+    projectValidation.run();
```
simonbrandhof: Why not calling `loadQualityProfiles(reportMetadata, organization);branchLoader.load(reportMetadata);` inside `loadProject()` ? That would avoid this non-conventional Runnable to reuse local variables.

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/step/LoadReportAnalysisMetadataHolderStep.java
```diff
@@ -99,17 +100,33 @@ private void loadMetadata(ScannerReport.Metadata reportMetadata) {
...
+      throw MessageException.of(format(
+        "Compute Engine task component key is null. Project with UUID %s must have been deleted since report was uploaded. Can not proceed.",
+        ceTask.getComponentUuid()));
     }
+    ComponentDto dto = toProject(reportProjectKey);
     analysisMetadata.setProject(new Project(dto.uuid(), dto.getDbKey(), dto.name()));
+    return () -> {
+      if (!componentKey.equals(reportProjectKey)) {
+        throw MessageException.of(format(
+          "ProjectKey in report (%s) is not consistent with projectKey under which the report as been submitted (%s)",
```
janos-ss: `s/as been/has been/` (an annoying typo that was there in the original too...)

server/sonar-server/src/main/java/org/sonar/server/computation/task/projectanalysis/step/LoadReportAnalysisMetadataHolderStep.java
janos-ss: I agree with Simon. The `loadProject` method does validation in two stages. First when called, and second with the returned `Runnable`. This is really hacky, there's gotta be a better way to write it.

sns-seb:
> I agree with Simon. The loadProject method does validation in two stages. First when called, and second with the returned Runnable. This is really hacky, there's gotta be a better way to write it.

Feel free to suggest. I didn't find any way which wouldn't create a huge code mix and/or make even more complexe code

PR #3123, opened 2018-03-06
SONAR-10175 delete project assiociations when deleting Quality Gate

sns-seb:


simonbrandhof:
As discussed offline, the existing corrupted data should also be fixed.

PR #3126, opened 2018-03-07
SONAR-10450 Add a 'submitter' columnin Background Task page

najibdarma:


stas-vilchik:
Hi @najibdarma! Thanks for the contribution, the feature was implemented in SonarQube 7.2.

PR #3132, opened 2018-03-08
SONAR-10465 Remove clickable tags on Marketplace's list of Plugins

pascal-mugnier-sonarsource:


server/sonar-web/src/main/js/apps/marketplace/components/PluginDescription.tsx
```diff
@@ -27,24 +27,14 @@ interface Props {
 }
 
 export default class PluginDescription extends React.PureComponent<Props> {
-  handleCategoryClick = (e: React.SyntheticEvent<HTMLAnchorElement>) => {
-    e.preventDefault();
-    this.props.updateQuery({ search: this.props.plugin.category });
```
stas-vilchik: `updateQuery` can now be removed from the props. Also check the parent components, they might not need this prop as well.

stas-vilchik: and actually this component should become functional

PR #3137, opened 2018-03-08
SONAR-9206 Enhance /api/components/search_projects response with organizations

gjambet:


sonar-ws/src/main/protobuf/ws-components.proto
```diff
@@ -128,3 +124,10 @@ message Component {
   }
 }
 
+message Organization {
```
teryk: I have the feeling Organization should be in the ws-common.proto to have a common representation of an organization

server/sonar-server/src/main/java/org/sonar/server/component/ws/SearchProjectsAction.java
```diff
@@ -323,6 +324,15 @@ private SearchProjectsWsResponse buildResponse(SearchProjectsRequest request, Se
           .forEach(response::addComponents);
         return response;
       })
+      .map(response -> {
```
teryk: I have the feeling it should optional, at least I don't feel comfortable to add organization with a more global perspective

gjambet:
after discussions w Stass & Teryk : suggestions taken into account, merging as good on tech side

PR #3155, opened 2018-03-14
Fix minor safety issues in built-in patches

janos-ss:


scripts/patches/debug_ce.sh
```diff
@@ -6,5 +6,6 @@ source scripts/property_utils.sh
 
 SQ_HOME=$1
 
-echo "enabling debug on compute engine, listening on port 5005"
-set_property sonar.ce.javaAdditionalOpts -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 "$SQ_HOME/conf/sonar.properties"
+port=5005
```
sns-seb: in my patches, I use 5006 for ce and 5005 for web.
Probably because web always starts before ce (and existing even before ce existed).

scripts/patches/README.md
```diff
@@ -34,8 +34,9 @@ Example scripts
 
     SQ_HOME=$1
 
-    echo "enabling debug in conf/sonar.properties, listening on port 5005"
-    set_property sonar.web.javaAdditionalOpts -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 "$SQ_HOME/conf/sonar.properties"
+    port=5005
+    echo "enabling debug in conf/sonar.properties, listening on port $port"
+    set_property sonar.web.javaAdditionalOpts -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=$port "$SQ_HOME/conf/sonar.properties"
```
sns-seb: no quotes around `sonar.web.javaAdditionalOpts` and `-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=$port` ?

I start to do it because vi colors argument when quotes are present and doesn't when they are missing.
Probably, it's would also be relevant in this case because of `$port`

scripts/patches/debug_ce.sh
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
```
sns-seb: is it on any use if file is not intended to be called directly?

(I'm also wondering about the executable file permission)

scripts/patches/debug_ce.sh
janos-ss: I switched it ;-)

scripts/patches/README.md
janos-ss: A space in `$port` is a highly unlikely mistake. But since the parameter is quite long, it's not obvious at a glance that it's safe to not quote, so I added the quotes now.

scripts/patches/debug_ce.sh
janos-ss: The missing shebang suggests the script is source-d. But it expects `$1` to be present. It's not sourced, it's invoked like this in `call_patches`:

    "$script" "$sq_home"

To disambiguate how it will be interpreted, it really should have a shebang.

The current invocation pattern requires the executable permission, and I think it's good like this. Note that patches are not limited to Bash, they could be something else too (awk, python, perl), with appropriate shebang.

sns-seb: sourced => no shebang, not sourced => sheband

this makes sense.

However, all our sourced .sh files in `scripts/` have shebangs and are executable.

janos-ss: It would be good remove the executable bit from the scripts that are intended to be sourced. The shebang might be still good to keep, to signal to the reader the required compatibility level. If it's not there, I will wonder if it's ok to use Bash syntax (sourced by a Bash script) or if I should stick to `/bin/sh` syntax (sourced by `/bin/sh` script).

sns-seb: sounds good to me

PR #3163, opened 2018-03-22
SONAR-10285 WS org/enable_support is now only available on SonarCloud

julienlancelot:


tests/src/test/java/org/sonarqube/pageobjects/Navigation.java
```diff
@@ -59,7 +59,7 @@ public static Navigation create(Orchestrator orchestrator) {
     clearStorage(d -> d.getLocalStorage().clear());
     clearStorage(d -> d.getSessionStorage().clear());
     clearStorage(d -> clearBrowserLocalStorage());
-    return Selenide.open("/", Navigation.class);
+    return Selenide.open("/projects", Navigation.class);
```
simonbrandhof: why this change?

simonbrandhof:
I suggest to do the same with all web services api/organizations. WDYT?

julienlancelot:
It's not possible to disable all org ws as api/organization/search is used by front end, so we can disable all except this last one

julienlancelot:
As discussed all org ws except search will be removed

PR #3172, opened 2018-05-22
Hardening 6.7

gregaubert:

### DIFF
--- a/server/sonar-web/src/main/js/components/issue/components/IssueAssign.js
+++ b/server/sonar-web/src/main/js/components/issue/components/IssueAssign.js
@@ -52,13 +52,13 @@ export default class IssueAssign extends React.PureComponent {
             <Avatar
               className="little-spacer-right"
               hash={issue.assigneeAvatar}
-              name={issue.assigneeName}
+              name={issue.assigneeName || issue.assignee}
               size={16}
             />
           </span>
         )}
         <span className="issue-meta-label">
-          {issue.assignee ? issue.assigneeName : translate('unassigned')}
+          {issue.assignee ? issue.assigneeName || issue.assignee : translate('unassigned')}
         </span>
       </span>
     );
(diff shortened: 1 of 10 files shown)


server/sonar-web/src/main/js/components/issue/components/IssueCommentLine.js
```diff
@@ -85,10 +85,10 @@ export default class IssueCommentLine extends React.PureComponent {
           <Avatar
             className="little-spacer-right"
             hash={comment.authorAvatar}
-            name={comment.authorName}
+            name={comment.authorName || comment.author}
```
pascal-mugnier-sonarsource: Maybe this should be done to the "title" attribute of the parent div. If so, we would have 3 repetitions of the same code so we might need to do `const name = comment.authorName || comment.author` first

server/sonar-web/src/main/js/components/issue/components/IssueAssign.js
```diff
@@ -52,13 +52,13 @@ export default class IssueAssign extends React.PureComponent {
...
               hash={issue.assigneeAvatar}
-              name={issue.assigneeName}
+              name={issue.assigneeName || issue.assignee}
               size={16}
             />
           </span>
         )}
         <span className="issue-meta-label">
-          {issue.assignee ? issue.assigneeName : translate('unassigned')}
+          {issue.assignee ? issue.assigneeName || issue.assignee : translate('unassigned')}
```
pascal-mugnier-sonarsource: I understand it does the job but it looks weird to me... Maybe doing `const name = issue.assigneeName || issue.assignee;` first would make it more subtle. That would also avoid a code repetition. No big deal tho.

gregaubert: I won't fix this one, because of that type undefined that I told you about.

PR #3183, opened 2018-10-03
Put current userid in response header

flopma:
Currently, the current userid can be logged in the access log file provided by Sonarqube.
But some enterprise are using reverse proxies, and use the access logs from the reverse proxies instead. This provides the gain that a unified reverse proxy log setup can be put in place and used with Logstash for example to inject the logs into ElasticSearch.
So be able to do that, that information have to be made available on the response header.


benoit-sns:
Hi,

We have updated our contributions requirements and plan to be more responsive to external PR in the future.
Please be aware that we are not actively looking for feature contributions. The truth is that it's extremely difficult for someone outside SonarSource to comply with our roadmap and expectations. Therefore, we typically only accept minor cosmetic changes and typo fixes.
If you would like to see a new feature, like this one, please create a new thread in the forum ["Suggest new features"](https://community.sonarsource.com/c/suggestions/features).

Thank you for you contribution anyway!

PR #3186, opened 2018-11-14
SONAR-11475 Check if url is relative before redirecting

gregaubert:


server/sonar-web/src/main/js/apps/maintenance/main-view.js
```diff
@@ -90,6 +90,14 @@ export default Marionette.ItemView.extend({
     setInterval(() => {
       window.location = this.options.returnTo || getBaseUrl();
     }, 2500);
+
+    setInterval(() => {
+      let returnTo = this.options.returnTo;
+      if (!returnTo || !isRelativeUrl(returnTo)) {
```
stas-vilchik: Same comment: extract this to a function.

PR #3187, opened 2018-11-16
Data coming from Cobertura report cannot be analysed properly

mayito:
Hi,

On Windows 10 64 bits, Sonarqube is not able to parse a Cobertura report because the XML file contains the source tag which is structured like this : 

```
<sources>
    <source>
       X:\path\to\my\folder
    </source>
</sources>
```

To work properly, this tag must be structure like this :

```
<sources>
    <source>X:\path\to\my\folder</source>
</sources>
```

Is there a way to make the report analysis a bit more robust ?

Regards,

Thierry BRIZZI.


henryju:
@dbmeneses can you take care of this please?

dbmeneses:
@mayito I understand the problem you have, but is this actually a P/R? Could you please fix it if that's the case?

mayito:
@dbmeneses I've found a turn around by correcting the Cobertura report after generation to match the format awaited by Sonarqube.

Though, Sonarqube scanner should be a bit more robust and the logged errors more expressive.

How I can register a feature request ? Since the GitHub project seems only focused on P/R and developpers communication.

Regards.

simonbrandhof:
Closing this PR which aims to merge the branch-7.0 in master!!

PR #3194, opened 2019-01-29
Add properties to change the initial admin username and password

BramCeulemans:
Because of security reasons, that I will explain later, this change adds two properties to the application to change the default username and password for SonarQube (admin/admin, by default).

Why do we need this? Currently it is possible to change the Authentication to SSO with OAuth2 already using the properties. However when you do this, it still leaves the admin/admin credentials default.

Mainly this is done to make integrating into the docker image easier, because for example in docker-compose is would be quite nasty to do post-startup database changes and this allows for users to set the defaults themselves.

Also, here is an example of what could happen if someone forgot to change the default password when having SSO enabled: https://nakedsecurity.sophos.com/2018/05/14/2-million-lines-of-source-code-left-exposed-by-phone-company-ee/

### DIFF
--- a/server/sonar-ce/src/test/java/org/sonar/ce/container/ComputeEngineContainerImplTest.java
+++ b/server/sonar-ce/src/test/java/org/sonar/ce/container/ComputeEngineContainerImplTest.java
@@ -19,12 +19,6 @@
  */
 package org.sonar.ce.container;
 
-import java.io.File;
-import java.io.IOException;
-import java.util.Date;
-import java.util.Locale;
-import java.util.Properties;
-import java.util.stream.Collectors;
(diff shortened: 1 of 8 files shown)


BramCeulemans:
I don't know, but why is the Travis failing? Is it because this is a pull request that it does not have access to secrets?

julienlancelot:
Thanks a lot for this contribution @BramCeulemans !
We are aware that we need to increase the security of the admin user, and we're working on this.
You can follow our progress here : https://jira.sonarsource.com/browse/MMF-1352

I'm closing this PR as we don't know yet how we'l implement this, but tanks for being concerned by this topic !

julienlancelot:
I can't give you any ETA on that point, but it's in our backlog.

PR #3195, opened 2019-01-31
Increase number of supported webhooks

alexandruivan:
Please ensure your pull request adheres to the following guidelines: 

- [ ] Please explain your motives to contribute this change: Currently, if you have more than 10 jenkins instances you cannot use the same sonarqube instance as the number of the webhooks limit you.

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/webhook/ws/CreateAction.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/webhook/ws/CreateAction.java
@@ -59,7 +59,7 @@ import static org.sonarqube.ws.Webhooks.CreateWsResponse.newBuilder;
 
 public class CreateAction implements WebhooksWsAction {
 
-  private static final int MAX_NUMBER_OF_WEBHOOKS = 10;
+  private static final int MAX_NUMBER_OF_WEBHOOKS = 100;
 
   private final DbClient dbClient;
   private final UserSession userSession;


julienlancelot:
Hi @alexandruivan,

Could you please explain why have you done this change ?
Which difficulties have you encountered ?

Thanks

simonbrandhof:
@alexandruivan I'm not a Jenkins expert, so could you please explain in which case you need to notify multiple Jenkins servers (even 2)?

alexandruivan:
In order to use the Jenkins pipeline step waitForQualityGate, a webhook from Sonarqube needs to be set to point at the jenkins instance. Documentation can be found here: https://jenkins.io/doc/pipeline/steps/sonar/
Here is the jenkins plugin: https://github.com/jenkinsci/sonarqube-plugin

alexandruivan:
> @alexandruivan I'm not a Jenkins expert, so could you please explain in which case you need to notify multiple Jenkins servers (even 2)?

@simonbrandhof In case you don't have a centralized Jenkins instance, where you have multiple teams having their own Jenkins, you need to set multiple webhooks to point to each Jenkins instance.

simonbrandhof:
I suppose that a project is hosted in a single Jenkins instance. In that case the webhook should be configured in SonarQube on the project level, but not globally. Makes sense?

alexandruivan:
It makes sense. I didn't know about the option per project, I thought there are 10 webhooks and that's it.

simonbrandhof:
Glad that it helps you!

PR #3199, opened 2019-03-02
[TRANSPARENCY] Disable default server info collect by SonarSource

begarco:
By default, SonarSource collects server information about all SonarQube instances. 

This collect can be disabled by setting the property `sonar.telemetry.enabled` to `false`. 

But by defaut, this is enabled and SonarQube servers send server information to **https://telemetry.sonarsource.com/sonarqube** avery 6 hours. Collected information are those provided by `/api/system/info`.

No information is openly provided to users about telemetry feature and its default configuration.

It would be more *transparent* and *ethic* towards users and the community to set this feature to `false` by default, or at least warn them about **telemetry** before they launch a SoanrQube server.


benoit-sns:
Hi,

For the several reasons mentioned on the [related thread on the community forum](https://community.sonarsource.com/t/transparency-disable-default-server-info-collect-by-sonarsource/7424/9), I will close this PR.

Thank you!

PR #3205, opened 2019-04-12
Build a correct path when no message

Ripolin:
Just fixing a little bug when using a custom context path (ex: https//www.example.org/sonarqube). The generated path is not good when there is no additional message.

Thix bug is present on branches 6.7 and master too.

### DIFF
--- a/server/sonar-server/src/main/java/org/sonar/server/authentication/AuthenticationError.java
+++ b/server/sonar-server/src/main/java/org/sonar/server/authentication/AuthenticationError.java
@@ -57,7 +57,7 @@ final class AuthenticationError {
   private static String getPath(AuthenticationException e, String contextPath) {
     String publicMessage = e.getPublicMessage();
     if (publicMessage == null || publicMessage.isEmpty()) {
-      return UNAUTHORIZED_PATH;
+      return contextPath + UNAUTHORIZED_PATH;
     }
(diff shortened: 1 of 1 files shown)


henryju:
@julienlancelot seems related to https://jira.sonarsource.com/browse/SONAR-8413, can you handle this PR?

julienlancelot:
@Ripolin Could you please explain me in which use case you have encounter this bug ? So that I can create a JIRA ticket. Thanks

Ripolin:
@julienlancelot  My Sonarqube server is installed behind a NGINX HTTP reverse proxy which redirect all requests based on a custom path: example https://mydomain.example.org/sonarqube. I fill correctly 
 the ```sonar.core.serverBaseURL```  parameter and install SAMLv2 plugin but I make a mistake when defining user's roles configuration in my IdP. So Sonarqube normally redirect me on an error page but instead of  https://mydomain.example.org/sonarqube/sessions/unauthorized, it redirects me to  the https://mydomain.example.org/sessions/unauthorized url and generate 404 error page.

julienlancelot:
Thanks, I've created a ticket to take this fix into account : https://jira.sonarsource.com/browse/SONAR-11993

julienlancelot:
I've fixed this issue in this commit https://github.com/SonarSource/sonar-enterprise/commit/a51820b2bc2fc751ada5103d019288b01204029b.

I'm closing this PR, but thanks @Ripolin for raising this issue !

PR #3211, opened 2019-05-31
Replace 3 regex with 1 in DaoUtils.java

farnulfo:
Perfomance and factorising issue.
Use one regex instead of 3 by factorising the matched expression with captured group zero.

First PR, please be kind :-)

For the record: while working on https://community.sonarsource.com/t/new-rule-to-check-if-first-argument-of-string-replaceall-is-really-a-regexp/10085 I was searching if there is some possible use in Sonar.
Found this where each of 3 ".replaceAll" can be replaced by ".replace" (because first argument is not a regex) but then the 3 ".replace" can be replaced by one replaceAll with this time a true regex. :-)

PR #3213, opened 2019-06-17
Simple test to check if travis CI for SonarQUBE is really working for external PR

farnulfo:
This PR https://github.com/SonarSource/sonarqube/pull/3211 failed on Travis CI (https://travis-ci.org/SonarSource/sonarqube/builds/546984454) but I don't know why.

Here is simple text file added that shouldn't break Travis CI build.

### DIFF
--- a/test_travis_ci.md
+++ b/test_travis_ci.md
@@ -1,0 +1,1 @@
+Simple modification to test SonarQUBE travis CI


farnulfo:
Hi @henryju @simonbrandhof  : there is something wrong with SonarQUBE config for Travis.
This PR shouldn't had failed (just a md file added).

simonbrandhof:
The PR fails because master currently fails! This case should never occur, but that's only the theory...

farnulfo:
Travis CI seems fine now. Closing the PR

PR #3217, opened 2019-07-05
SONAR-12260 - Fix invalid MSSQL delete query

stefan-gheorghe-leica:
SONAR-12260 - Fix invalid MSSQL delete query for project_measures

* Original query "delete from project_measures pm where component_uuid=?" was failing on MSSQL v14 with the error "Incorrect syntax near 'pm'.", basically for delete, the pm alias requires an extra pm after the delete keyword. However, for the purpose of this query no alias is needed, therefore refactored without the pm alias which
* Tests already exist in DeleteFileMeasuresTest.java
* Checked also the similar queries from DeletePersonMeasures.java, these are working fine
* Checked all "Delete ... from" queries, there's none using aliases besides the above mentioned ones


sns-seb:
Hi,

Thanks for the fix @stefan-gheorghe-leica.

However, this ticket has already been fixed internally so I won't be merging this contribution. 
The merge of this internal change is coming soon.

PR #3233, opened 2020-01-09
SONAR-10485: Allow configuration of small changeset sizes

mc1arke:
SONAR-9352 introduced the concept of small changesets to allow the Quality Gate to ignore certain failing conditions where only a small number of lines had changed. This is useful for some scenarios where changes to a single line in code that already contained technical debt would not be feasible to make without large refactoring which would prevent making small changes quickly. However, this allows technical debt to creep into a code-base over time, such as new lines not having any tests to cover them, or code smells being introduced as part of minor changes.

This change introduces a new property to the UI of `sonar.smallchangeset.maxlines`, set at a Global level and allowed to be overridden per project. Any changeset that changes less than this number of lines will continue to have certain metrics excluded from Quality Gate checks, but any change that changes more than this number of lines will have all their metrics checked. To retain backwards-compatibility with the current change-set filtering, the default value treats and changeset with less than 20 lines as a small changeset. Setting this value to `0` would effectivley force all changesets to have all metrics checked, so require all changes to fix any issues that exist on the lines being changed, and not to introduce any new issues.

### DIFF
--- a/server/sonar-ce-task-projectanalysis/src/main/java/org/sonar/ce/task/projectanalysis/step/SmallChangesetQualityGateSpecialCase.java
+++ b/server/sonar-ce-task-projectanalysis/src/main/java/org/sonar/ce/task/projectanalysis/step/SmallChangesetQualityGateSpecialCase.java
@@ -21,8 +21,11 @@ package org.sonar.ce.task.projectanalysis.step;
 
 import java.util.Collection;
 import javax.annotation.Nullable;
+
+import org.sonar.api.CoreProperties;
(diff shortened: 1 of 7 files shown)


julienlancelot:
Thanks @mc1arke for this contribution. 

Unfortunately, I'm not convinced by the solution you've chosen to fix this issue. Indeed, adding a new setting to specify any possible value as the number of changed lines to be considered by Quality Gate can have some unexpected impact, and then it will be hard to help users investigate on what is happening.

I would recommend you to raise another topic on the community to raise awareness on the subject, in order to find the best solution to implement this improvement.

PR #3234, opened 2020-02-06
Add config parameter to increase elasticsearch index.max_result_window

tuomari:
When fetching over 10000 issues through api ( /api/issues/search?p=21&ps=-1 ), elasticsearch fails with exception `Result window is too large, from + size must be less than or equal to: [10000] but was [10500].` Increasing `index.max_result_window` will allow fetching of all issues for a project. 

For example IntelliJ Idea sonarqube plugin (https://github.com/sonar-intellij-plugin/sonar-intellij-plugin) will fail if there are over 10000 issues in a project..

### DIFF
--- a/server/sonar-main/src/main/java/org/sonar/application/es/EsSettings.java
+++ b/server/sonar-main/src/main/java/org/sonar/application/es/EsSettings.java
@@ -157,5 +157,11 @@ public class EsSettings {
     if (props.value("sonar.search.javaAdditionalOpts", "").contains("-D" + ALLOW_MMAP + "=false")) {
       builder.put(ALLOW_MMAP, "false");
     }
+
+    // Elasticsearch max results. This might need increasing when there are a lot of issues.
+    String maxResults = props.value("sonar.search.maxResults", "");
+    if (!maxResults.isEmpty()) {
+      builder.put("index.max_result_window", maxResults);
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


sns-seb:
Hello,

Increasing `index.max_result_window` is not fixing the problem, only pushing it a bit further away, at the cost of memory (see `index.max_result_window` at https://www.elastic.co/guide/en/elasticsearch/reference/6.8/index-modules.html#dynamic-index-settings) and bad performance.

WS api/issues/search should rather be fixed to use ES's scroll API.

Anyway, it's questionable whether it makes sense to display more than 10k issues per project at all without putting in any extra filter (issue type at least), and even more in an IDE.

inoreip:
Hi

Can we at least make sonarqube honour any settings in the elasticsearch.yml file on startup. It seems strange you cannot tailor the search index behaviour to a particular deployment. Even if these properties are defined in SonarQube.properties file.

vanderby:
While this PR does make this ES setting a first class citizen, it is not needed. If you enable ES's http port by setting sonar.search.httpPort you can then make a PUT request to change any ES setting. A more detailed walkthrough can be found at: https://github.com/vanderby/SonarQube-AzureAppService/issues/40#issuecomment-624414812

Depending on how your setup with SQ is, you may want to disable the ES http port after the change so you don't keep the endpoint open to the public internet.

pierre-guillot-sonarsource:
Hey @jacek-poreda-sonarsource, could you handle this one, please?

jacek-poreda-sonarsource:
Hello @tuomari ,

Thanks for the contribution, but as @sns-seb mentioned, it is not fixing root of the problem, just postponing it further.

Also with the current state of Elasticsearch API I'm unsure if scroll api should be used here either, hence I'm closing that PR.

PR #3235, opened 2020-02-13
Do not warn if the file is excluded already

tasomaniac:
This change is only about excessive warnings in the console. We have thousands of these comments in our project because we have a certain folder that is shared between projects.

Even when I exclude the unnecessary files with Sonar, this continues to Log them.

### DIFF
--- a/sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/filesystem/FileIndexer.java
+++ b/sonar-scanner-engine/src/main/java/org/sonar/scanner/scan/filesystem/FileIndexer.java
@@ -98,14 +98,6 @@ public class FileIndexer {
     throws IOException {
     // get case of real file without resolving link
     Path realAbsoluteFile = sourceFile.toRealPath(LinkOption.NOFOLLOW_LINKS).toAbsolutePath().normalize();
-    if (!realAbsoluteFile.startsWith(project.getBaseDir())) {
-      LOG.warn("File '{}' is ignored. It is not located in project basedir '{}'.", realAbsoluteFile.toAbsolutePath(), project.getBaseDir());
-      return;
-    }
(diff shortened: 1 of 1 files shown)


julienlancelot:
@tasomaniac  Could you please elaborate more on the issue you're encountering by opening a thread in the Community forum : https://community.sonarsource.com/c/help/sq/10 ?
Thanks a lot, it will help us to understand if your issue is specific to your context, or can be shared by other people.
In the meantime, I'll close this PR.

tasomaniac:
I'm not sure if I will invest time on that. If you look at the code, it is kind of obvious. That's why I wanted to just open the PR.

If you look at the code, the warning is the first thing that is done. The warnings says that the file is ignored. Why would you really need to put this warning if the user already excluded the files explicitly?

dbmeneses:
Makes sense. Were you able to validate the change? There are no tests changed.

tasomaniac:
Not really, unfortunately. I did not expect that the project has tests for log behavior. Should I add tests?

dbmeneses:
Yes. See `FileSystemMediumTest#ignore_files_outside_module_basedir`. You could add a similar test in that class with the exclusion.

dbmeneses:
Ticket created: https://jira.sonarsource.com/browse/SONAR-13094

tasomaniac:
Thanks for pointing that test. I will add soon. I will also try to reproduce the behavior with tests first.

dbmeneses:
I've added a test and as soon as tests pass I'll merge.
Thanks for the contribution.

tasomaniac:
I couldn't find the merge. Can you give link to the new tests? Thanks.

dbmeneses:
It should be sync'ed to this repo by tomorrow

PR #3237, opened 2020-03-05
Add a CockroachDB dialect

rafiss:
This dialect allows users to run SonarQube with CockroachDB as a
backend. In general, CockroachDB is compatible with Postgres. Certain
settings and statements differ however. In particular, CockroachDB does
not currently support the `backslash_quote` setting, so this dialect
does not attempt to set it. Also, this dialect uses a JDBC URL prefixed
with `cockroachdb`.

### DIFF
--- a/server/sonar-db-core/src/main/java/org/sonar/db/dialect/CockroachDB.java
+++ b/server/sonar-db-core/src/main/java/org/sonar/db/dialect/CockroachDB.java
@@ -1,0 +1,73 @@
+/*
+ * SonarQube
+ * Copyright (C) 2009-2020 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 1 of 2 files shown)


pierre-guillot-sonarsource:
@simonbrandhof Any opinion on this ?

simonbrandhof:
Thank you very much for the contribution.

Supporting a new database vendor requires:
* to be sure that the need exists. Reverting an unused feature is never a good experience.
* to automate testing in the delivery workflow (integration tests).
* to train the SonarQube development team for an optimal maintenance.
* to train the SonarSource support and consulting teams for an optimal customer experience.

I'm afraid that CockroachDB is not eligible to an official support for the time being. I suggest to reopen the topic later.

Thanks for your understanding.

rafiss:
@simonbrandhof Thanks for the explanation! That makes sense. Just to provide some background on this change: we at Cockroach Labs got a support issue from a user who was trying to use SonarQube with a CockroachDB backend. See https://github.com/cockroachdb/cockroach/issues/45774

simonbrandhof:
Good to know, thanks for the follow-up.

PR #3243, opened 2020-07-09
Fix SAML check feature using Tomcat Proxy configuration (https://tomc…

Ripolin:
Hi,

To me, correction of ticket [SONAR-13328](https://jira.sonarsource.com/browse/SONAR-13328) seems buggy. It only checks HTTP header ```X-Forwarded-Proto``` but behind a reverse proxy hostname is frequently changed too and header ```X-Forwarded-Host``` could be sent.

If Sonarqube is installed behind a reverse-proxy, I think it's better to customize Tomcat than hack incoming HTTP request  (https://tomcat.apache.org/tomcat-9.0-doc/proxy-howto.html).

So in this PR I add 3 new properties to configure RP :
* ```sonar.web.proxyName``` : reverse proxy hostname
* ```sonar.web.proxyPort``` : reverse proxy port
* ```sonar.web.scheme``` : reverse proxy incoming scheme

When this properties are filled, Tomcat will return information about the original HTTP(S) request when ```getRequestURL()``` is called (by [SamlResponse](https://github.com/onelogin/java-saml/blob/26a29dad1a8603d47a9393980d9e7822064399f2/core/src/main/java/com/onelogin/saml2/authn/SamlResponse.java#L108) for example).

### DIFF
--- a/server/sonar-auth-saml/src/main/java/org/sonar/auth/saml/SamlIdentityProvider.java
+++ b/server/sonar-auth-saml/src/main/java/org/sonar/auth/saml/SamlIdentityProvider.java
@@ -23,19 +23,6 @@ import com.onelogin.saml2.Auth;
 import com.onelogin.saml2.exception.SettingsException;
 import com.onelogin.saml2.settings.Saml2Settings;
 import com.onelogin.saml2.settings.SettingsBuilder;
-import java.io.IOException;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.regex.Pattern;
-import javax.annotation.CheckForNull;
-import javax.annotation.Nullable;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletRequestWrapper;
-import javax.servlet.http.HttpServletResponse;
 import org.sonar.api.server.ServerSide;
 import org.sonar.api.server.authentication.Display;
 import org.sonar.api.server.authentication.OAuth2IdentityProvider;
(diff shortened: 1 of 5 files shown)


pierre-guillot-sonarsource:
Hey @michal-duda-sonarsource can you have a look at this please?

julienlancelot:
Hi,

Could you please open a community thread in https://community.sonarsource.com/c/help/sq/10 and explain exactly why it's not possible to use the reverse proxy?

Thanks

Ripolin:
https://community.sonarsource.com/t/correction-of-sonar-13328-dont-work-when-hostname-is-changed/27838

pierre-guillot-sonarsource:
ping @michal-duda-sonarsource, could you please conclude on this PR?

wouter-admiraal-sonarsource:
@pierre-guillot-sonarsource Do you need any help with this?

pierre-guillot-sonarsource:
Hey Wouter, thanks for the ping. I missed the `assigned` notification. I actually need help on this indeed, as the setup to reproduce this behaviour is not easy and I don't know really well the SAML auth. Still, I think this change request could make sense and needs to be investigated. I suggest that we track this as a Jira ticket and add it to the potential hardening scope for 9.0. Make sense to you @wouter-admiraal-sonarsource?

wouter-admiraal-sonarsource:
That makes a lot of sense 👍 . Ticket created, and added to the 9.0 scope: https://jira.sonarsource.com/browse/SONAR-15051

@Ripolin thank you very much for your contribution and patience. We will try and have a look at this in the coming weeks, and see if we can ship it as part of 9.0. You can track our progress by checking the above ticket.

I will close this PR for now.

Ripolin:
@pierre-guillot-sonarsource To me it's not an problem about SAML auth, but how Sonarqube is  proxyfied.

PR #3245, opened 2020-07-10
Branch 7.9

oliverbcv:
hi folks

I run my sonaquber with docker and i've mount the conf volume with all the parameters LDAP connect .
However when i try to connect i cant .
i get the error below:
Authentication failed

So i would want to run sonarqube in debug mode . But i didn't find anything on how i can overwrite the variable in the sonarqube conf .


pierre-guillot-sonarsource:
Hi, please browse the community forum for answer, and if you found nothing, create a new topic in the community forum. https://community.sonarsource.com/

PR #3248, opened 2020-09-14
Add Gitea ALM integration

lafriks:
Add support for Gitea (https://gitea.io/) ALM integration.

Any feedback would be nice.

### DIFF
--- a/sonar-ws/src/main/java/org/sonarqube/ws/client/almsettings/AlmSettingsService.java
+++ b/sonar-ws/src/main/java/org/sonarqube/ws/client/almsettings/AlmSettingsService.java
@@ -112,6 +112,22 @@ public class AlmSettingsService extends BaseService {
   }
 
   /**
+   * This is a POST request.
+   * @see <a href="https://next.sonarqube.com/sonarqube/web_api/api/alm_settings/create_gitea">Further information about this action online (including a response example)</a>
+   * @since 8.5
+   */
+  public void createGitea(CreateGiteaRequest request) {
+    call(
+      new PostRequest(path("create_gitea"))
+        .setParam("appId", request.getAppId())
+        .setParam("key", request.getKey())
+        .setParam("url", request.getUrl())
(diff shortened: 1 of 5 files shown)


pierre-guillot-sonarsource:
Hi, thanks for your contribution. I see that you were inspired by the GitHub integration, could you please specify in what extends they are similar? I'm not familiar at all with Gitea.

Also, the PR decoration mechanism is close-source, so accepting this PR as-is would only result in creating some endpoint, but the actual PR decoration mechanism would not be implemented.

lafriks:
@pierre-guillot-sonarsource yes, Gitea has very similar API to GitHub rest API (endpoints/json structure etc). I was looking as possibility to implement this as plugin but there is no way to extend this through plugins. If there was ALM integration definition in SonarQube core. Decorations would be possible to implement using plugin as otherwise there is currently no other way I could find as even ClientDb can not be extended to add this. Especially ALM enum is limiting factor for this.

pierre-guillot-sonarsource:
Currently, is there is something stopping you to create a GitHub integration, and define as API endpoint your Gitea url?

lafriks:
They are not really exact. For example GitHub organizations would be: `GET https://api.github.com/users/lafriks/orgs` vs `GET https://<gitea-host>/api/v1/users/lafriks/orgs` and there are quite some more small differences so they will not work out of the box anyway.

lafriks:
@pierre-guillot-sonarsource I could most probably implement decorations myself for Gitea as plugin if #3249 got merged as currently than enum is blocking any way to extend this through plugins.

pierre-guillot-sonarsource:
Hi, thanks for your suggestion. Unfortunately adding Gitea to the list of supported ALM would require much more work than that (including the web UI). Your best option to make this happen is to push as a feature request on our [community forum](https://community.sonarsource.com/)

tomroffe:
As a paying customer... I would very much like to use an ALM intergration for gitea.

PR #3252, opened 2020-10-25
Implemented writeTimeoutMs

seaman333:
In some cases when Sonar's _Analysis Report_ is big enough, the upload process is failing with a _Session timeout_ error. This happens on big multimodal projects when analysis reports more than 1Mb. Given that nowadays=, many IT people working from home and the connection bandwidth varies a lot, in many cases, the default 10-sec write timeout inherited from OkHttp is not enough.

### DIFF
--- a/sonar-ws/src/main/java/org/sonarqube/ws/client/BaseRequest.java
+++ b/sonar-ws/src/main/java/org/sonarqube/ws/client/BaseRequest.java
@@ -50,7 +50,8 @@ abstract class BaseRequest<SELF extends BaseRequest> implements WsRequest {
   private final DefaultParameters parameters = new DefaultParameters();
   private final DefaultHeaders headers = new DefaultHeaders();
   private OptionalInt timeOutInMs = OptionalInt.empty();
-
+  private OptionalInt writeTimeOutInMs = OptionalInt.empty();
+  
   BaseRequest(String path) {
     this.path = path;
   }
@@ -72,6 +73,16 @@ abstract class BaseRequest<SELF extends BaseRequest> implements WsRequest {
 
   public SELF setTimeOutInMs(int timeOutInMs) {
     this.timeOutInMs = OptionalInt.of(timeOutInMs);
+    return (SELF) this;
+  }
+
+  @Override
+  public OptionalInt getWriteTimeOutInMs() {
+    return writeTimeOutInMs;
+  }
+
+  public SELF setWriteTimeOutInMs(int writeTimeOutInMs) {
+    this.writeTimeOutInMs = OptionalInt.of(writeTimeOutInMs);
     return (SELF) this;
   }
 
--- a/sonar-ws/src/main/java/org/sonarqube/ws/client/HttpConnector.java
+++ b/sonar-ws/src/main/java/org/sonarqube/ws/client/HttpConnector.java
@@ -55,6 +55,7 @@ public class HttpConnector implements WsConnector {
 
   public static final int DEFAULT_CONNECT_TIMEOUT_MILLISECONDS = 30_000;
   public static final int DEFAULT_READ_TIMEOUT_MILLISECONDS = 60_000;
+  public static final int DEFAULT_WRITE_TIMEOUT_MILLISECONDS = 240_000;
 
   /**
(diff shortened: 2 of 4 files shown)


seaman333:
Related conversations:
https://community.sonarsource.com/t/socket-time-out/26976/6
https://community.sonarsource.com/t/sonarqube-issue-failed-to-delete-temp-folder-directorynotemptyexception/16345/10
https://stackoverflow.com/questions/51259765/sonarqube-scanner-error-directorynotemptyexception/64534566#64534566

pierre-guillot-sonarsource:
Hi, to clarify a bit your issue here, you mean that SonarQube users are running full analysis locally?

seaman333:
> 
> 
> Hi, to clarify a bit your issue here, you mean that SonarQube users are running full analysis locally?

Hi, when developers are using maven as a build tool, sonar-maven-plugin connects to the SQ instance, caches locally sonar-clients-shaded.jar from sonar server, and all the reports (coverage, find bugs, etc.) generated locally on the host where maven executed: It can be developer machine or build server. Then, the report being zipped and uploaded to SQ server <- problem with timeout happens there.

pierre-guillot-sonarsource:
Thanks for the reply. To clarify further, why would developers run local analysis? I'm trying to understand your use case. Usually, SonarQube scanner run with a QA, on a builder machine, that is supposed to have a stable connection.

seaman333:
> Thanks for the reply. To clarify further, why would developers run local analysis? I'm trying to understand your use case. Usually, SonarQube scanner run with a QA, on a builder machine, that is supposed to have a stable connection.

Hi, the connection is stable either for local or build server analysis, but the connectivity throughput might be an issue when the organization (relevant to our use case) has a multi-cloud infrastructure with only one SQ instance. Also, our developers and 3rd party code suppliers running _mvn verify_ (involves SQ analysis as well) prior to git push.

pierre-guillot-sonarsource:
> connectivity throughput might be an issue

Write timeout won't help you in that case, as the write timeout is raised when the client did not receive any `ack` within 10sec from the server. That means a connection "cut" of 10s, which is more a sign of unstable connection rather than bandwidth limit.

Did you try to increase the read timeout on the scanner? The property is `sonar.ws.timeout`. [Documentation is here](https://docs.sonarqube.org/latest/analysis/analysis-parameters/)

seaman333:
> > connectivity throughput might be an issue
> 
> Write timeout won't help you in that case, as the write timeout is raised when the client did not receive any `ack` within 10sec from the server. That means a connection "cut" of 10s, which is more a sign of unstable connection rather than bandwidth limit.
> 
> Did you try to increase the read timeout on the scanner? The property is `sonar.ws.timeout`. [Documentation is here](https://docs.sonarqube.org/latest/analysis/analysis-parameters/)

Sure I have checked first the `sonar.ws.timeout` property. According to the ws-client sources this timeout param only applies to readTimeout (GET requests of OkHttp lib). The SQ Analysis Report ws-client publishes via the POST request (relevant writeTimeout within okHTTP lib), and this not been implemented within ws-client. My patch adds this writeTimeout feature for POST requests. I actually tested implementation on my environment prior to the current pull request and it helped with the Socket Timeout error.

seaman333:
Tho OkHttp documentation contains an example when low writeTimeout leads to the SocketTimeoutException:
https://www.baeldung.com/okhttp-timeouts#write

intelgentil:
Hi seaman333 i have the same probleme in our big project . i get this error "Timeout Socket Closed" in jenkins. i'm using maven plugin sonar-maven-plugin version 3.7.0.1746 . can you tell me if your correction worked well in your case. i want to use it .
Thanks

dbmeneses:
Thanks for the contribution @seaman333.
There is a balance between waiting too long for a time out that is inevitable and not waiting enough time in very slow connections. With that in mind, we decided to:
* Only change the timeout in this specific problematic request (POST of the scanner report), since it's the only one reported and we don't want to risk unwanted side effects in all other requests
* Use 30s instead of 10s

We'll do the change on our side, which should be included in v8.8: https://jira.sonarsource.com/browse/SONAR-14527

cforce:
Hey guys, where its this new property documented? How can i configure it on maven sonar client or is this only server side and howto configure there?

PR #3254, opened 2020-12-07
Support space in //NOSONAR

Sebelino:
Currently, SonarQube will suppress issues reported on lines ending with a `//NOSONAR` comment, but does not affect lines ending with `// NOSONAR`.

It looks like this makes it impossible to make use of this feature together with code formatters that expect there to be a space in the middle. In particular, [Google's Java Style formatter](https://github.com/google/google-java-format) always wants there to be a space after the `//`, and deliberately provides no way to configure this behavior.


colin-mueller-sonarsource:
Hey there.

NOSONAR is implemented at the individual language analyzer level -- not in SonarQube's core. What you've changed is a plugin that gets used for testing.

Let me recommend that you start a discussion on https://community.sonarsource.com/ before submitting a PR!

PR #3255, opened 2020-12-17
Update install-server.md

cyclops26:
Updated installation documentation to still show an example of a "customized" web host IP without the example being a non-RFC-1918 address to avoid any users accidentally referencing and using that IP address since per best practice security you would never want to be running a server or the service directly on the edge.

SonarSource Team,
Feel free to delete if you feel it is unnecessary.


emretanriverdi (approved):

jacek-poreda-sonarsource:
Hey @cyclops26,
Thanks for contribution! Your commit will be soon synced to this repo.

PR #3256, opened 2021-02-15
Expose `revision` in PostProjectAnalysisTaskTester

codescale:
Hi,

it would be very helpful to have the `revision` property accessible via the PostProjectAnalysisTaskTester.

Thanks :)

### DIFF
--- a/sonar-plugin-api/src/main/java/org/sonar/api/ce/posttask/PostProjectAnalysisTaskTester.java
+++ b/sonar-plugin-api/src/main/java/org/sonar/api/ce/posttask/PostProjectAnalysisTaskTester.java
@@ -112,6 +112,8 @@ public class PostProjectAnalysisTaskTester {
   private Branch branch;
   private ScannerContext scannerContext;
   private String analysisUuid;
+  @Nullable
+  private String revision;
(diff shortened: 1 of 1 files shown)


pierre-guillot-sonarsource:
Hi, thanks for your contribution. Could you elaborate a bit about why you would like this `revision` to be accessible? What's the use case behind your request?

codescale:
Sure. I have developed a `PostProjectAnalysisTask` that gets the revision from the given context and uses it to send the analysis result to our SCM system. To be able to test the behavior I'd like to use the `PostProjectAnalysisTaskTester` instead of mocking everything by myself.

pierre-guillot-sonarsource:
> Sure. I have developed a `PostProjectAnalysisTask` that gets the revision from the given context and uses it to send the analysis result to our SCM system. To be able to test the behavior I'd like to use the `PostProjectAnalysisTaskTester` instead of mocking everything by myself.

thanks.

Hey @dbmeneses , any opinion here?

dbmeneses (approved):

dbmeneses:
Thanks. I'll merge it and it should show up in this repo within 24h.

PR #3260, opened 2021-04-04
SONAR-10485: Add minimum changeset size configuration for QG conditions

mc1arke:
SONAR-9352 previous introduced conditions into the Quality Gate checks to ignore changesets that did not meet certain standards when those changesets did not change 20 or more lines. This bypass of conditions was hard-codes into Sonarqube Quality Gate checks so could not be disabled by an administrator, and effectively allowed technical debt to be introduced to a code-base through multiple small changesets that didnt contain appropriate coverage or duplicated existing lines.

This change alters the Quality Gate conditions to introduce a minimum number of lines the changeset much reach before that condition is viewed as effective, so that Sonarqube administrators can decide which conditions they may want to ignore, and what size the changeset is allowed to get up to before that condition starts being enforced.

To support this change, a new `minimum_effective_lines` database column has been added to the `quality_gate_conditions` table, with a migration setting this value to `20` for any metric that was previously checked in `SmallChangesetQualityGateSpecialCase`, and `0` for any other metric to ensure backwards compatibility. The default Quality Gate created on a new Sonarqube instance has also been altered to set the `NEW_COVERAGE` and `NEW_DUPLICATED_LINES_DENSITY` metrics to have a condition threshold of 20 lines for consistency with historical operations, however the other metrics that were historically skipped are not being set in the default quality gate since there were no historical default values for them.

The `create_condition` and `update_condition` web service endpoints have been altered to have an additional `minimumEffectiveLines` parameter on the request and response, with this being set as mandatory to ensure anyone using these endpoints doesn't accidentally break their instance configuration be overwriting customised configuration with default values.

Finally, the web interface has been altered to show the new condition limit fields in the Quality Gate configuration table, and allow editing of them through the conditions modal. The message that is shown on the scan overview page if any conditions have been skipped has also been altered to remove the reference to the hard-coded limits that previously existed, and make it clear that a configurable limit has been hit.

image
image
image
image

Please review our [contribution guidelines](https://github.com/SonarSource/sonarqube/blob/master/contributing.md).

And please ensure your pull request adheres to the following guidelines: 

- [x] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [x] Use the following formatting style: [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset#code-style)
- [x] Provide a unit test for any code you changed
[...]

PR #3261, opened 2021-04-05
SONAR-11283: Add ability to exclude uncoverable lines in condition check

mc1arke:
When committing changes that touched lines that could not have any tests performed against them - such as README files - these edited lines were included in the line count used to check whether the Quality Gate conditions should be applied, and therefore artificially caused the threshold on the condition to be met, so failed the Quality Gate if other code that could be tested but wasn't feasible to test was also included in the changeset.

This change alters the Quality Gate conditions to introduce a flag that controls what metric to use when calculating the number of changed lines that is then used when comparing against the threshold in each condition, thereby allowing Sonarqube administrators to decide which conditions they may want to ignore if multiple lines are changes in non-code files, but only a limited number of executable lines are changed, or whether all edited lines should count towards the threshold. For backwards compatibility this new flag defaults to false so treats all edited lines as counting towards the threshold.

To support this change, a new `only_coverable_lines` database column has been added to the `quality_gate_conditions` table, with a migration setting this value to `false` for all existing conditions to ensure backwards compatibility. The default Quality Gate created on a new Sonarqube instance also sets this field to `false` for consistency with historical setups.

The `create_condition` and `update_condition` web service endpoints have been altered to have an additional `onlyIncludeCoverableLines` parameter on the request and response, with this being set as mandatory to ensure anyone using these endpoints doesn't accidentally break their instance configuration be overwriting customised configuration with default values.

Finally, the web interface has been altered to show the new condition  fields in the Quality Gate configuration table, and allow editing of them through the conditions modal.

image
image


mc1arke:
Note: this includes the commit from #3260 for allowing configuration of the minimum changeset size for each condition in the quality gate.

pierre-guillot-sonarsource:
Hey @dbmeneses, seems definitely related to https://jira.sonarsource.com/browse/SONAR-10485 that you are working on

wouter-admiraal-sonarsource:
Hi @mc1arke ,

First of all, that's some impressive PR you prepared there. Unfortunately, after internal discussion, we will not accept it. It's a shame that your other PR was rejected a year ago, because that is indeed the direction we're taking now. But at the time, it wasn't at all clear what we wanted to do with this feature.

As we mention in our [contribution guidelines](https://github.com/SonarSource/sonarqube/blob/master/docs/contributing.md), we very rarely accept new features from pull requests. And unfortunately we won't accept your PR either, because:

- We think this approach is too fine-grained. The risk that people mis-configure their Quality Gates is too high
- This is would complicate other plans we have for the Quality Gate feature in the 9.x series
- We want to go for the minimal step to achieve this, to lessen the maintenance cost (and the best solution is similar to the one you suggested in your other PR)

For the above reasons, we cannot accept your work. I'm sorry, as I see that you did put a lot of effort in this.

Best regards.

PR #3268, opened 2021-07-10
Limit gitlab group sync to user's groups

remyj38:
The /groups Gitlab API endpoint list all groups related to the Gitlab users, not only the groups on which the user is a member.

For example, if a user is a member of the projet `foo/bar`, the `foo` group will be returned with the current query even if he's not a member of the group.

With group sync enabled, the user will be added on Sonar groups even though he's not member of the Gitlab group.

The query should include the `min_access_level` set to 10 (guest access) to limit the returned list.

### DIFF
--- a/server/sonar-auth-common/src/main/java/org/sonar/auth/OAuthRestClient.java
+++ b/server/sonar-auth-common/src/main/java/org/sonar/auth/OAuthRestClient.java
@@ -64,7 +64,7 @@ public class OAuthRestClient {
 
   public static <E> List<E> executePaginatedRequest(String request, OAuth20Service scribe, OAuth2AccessToken accessToken, Function<String, List<E>> function) {
     List<E> result = new ArrayList<>();
-    readPage(result, scribe, accessToken, request + "?per_page=" + DEFAULT_PAGE_SIZE, function);
+    readPage(result, scribe, accessToken, request + ((request.contains("?")) ? "&" : "?") + "per_page=" + DEFAULT_PAGE_SIZE, function);
     return result;
   }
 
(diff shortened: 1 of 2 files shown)


jacek-poreda-sonarsource:
Hey @remyj38 ,

Thanks for the contribution and raising this issue. The fix you have created seems to be appropriate, but we need to asses the impact of this change within a team. For that I've create a Jira ticket to track this on our side with target version for 9.1.

Ticket: [SONAR-15171](https://jira.sonarsource.com/browse/SONAR-15171)

For now I'm closing this PR.

pierre-guillot-sonarsource:
Hey @remyj38 , I just fix SONAR-15171 and it will be available with SQ 9.1, soon to be released (like, next week).

Guimove:
Hello,

Is there any possibility to have an option to enable/disable this fix ? 

Let me explain, we have more than 500 developers, they are separated in subgroups like `org_root/business_unit/subgroup`

The rights for the users are not all set at the `org_root/business_unit/subgroup ` level, it can be on a subgroup or a project level.

Before this change, it was working well because all users of  `org_root/business_unit/subgroup` can access all project report on sonarqube.

Just after the 9.1 update, we faced an incident, to resolve this we had to recreate all subgroups under   `org_root/business_unit/subgroup` recursively.

Now we have another issue, it's for a gitlab project that allows a gitlab group to access it and not a user, this access seems not working with Sonarqube access sync.

Is there any change to have an option to enable the old way of sync that was working like a charm for us use case ? 
Thanks,

pierre-guillot-sonarsource:
Hello @Guimove , Could you please report this on our [community forum](https://community.sonarsource.com/)? thanks :)

PR #3281, opened 2021-10-20
add option to enable group sync at parent level

Guimove:
Hello,

This PR is regarding an issue we got since version 9.1.
The issue was introduced by this PR ==> https://github.com/SonarSource/sonarqube/pull/3268 (https://jira.sonarsource.com/browse/SONAR-15171)

Let me explain, we have more than 500 developers, they are separated in subgroups like `org_root/business_unit/subgroup`

The rights for the users are not all set at the `org_root/business_unit/subgroup ` level, it can be on a subgroup or a project level.

Before this change, it was working well because all users of  `org_root/business_unit/subgroup` can access all project report on sonarqube.

Just after the 9.1 update, we faced an incident, to resolve this we had to recreate all subgroups under   `org_root/business_unit/subgroup` recursively.

Now we have another issue, it's for a gitlab project that allows a gitlab group to access it and not a user, this access seems not working with Sonarqube access sync.

Is there any change to have an option to enable the old way of sync that was working like a charm for us use case ?

### DIFF
--- a/server/sonar-auth-gitlab/src/main/java/org/sonar/auth/gitlab/GitLabRestClient.java
+++ b/server/sonar-auth-gitlab/src/main/java/org/sonar/auth/gitlab/GitLabRestClient.java
@@ -47,9 +47,9 @@ public class GitLabRestClient {
 
   List<GsonGroup> getGroups(OAuth20Service scribe, OAuth2AccessToken accessToken) {
     if (settings.syncUserGroupsParentLevel()) {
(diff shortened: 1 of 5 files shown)


jacek-poreda-sonarsource:
Hello @Guimove , thanks for your contribution, but this change has functional impact on SQ (new property). We need to assess in the team if that's what we want to do with the issue you are encountering.

I have created a bug ticket [here](https://jira.sonarsource.com/browse/SONAR-15674), which you can track. We plan to tackle this one in 9.3 version.

PR #3304, opened 2022-09-03
Reset %errorlevel% after checks in batch scripts

felipebz:
When trying to start SonarQube 9.6.1 on Windows using the SonarStart.bat nothing happens:

image

**Cause:**

This issue happens when the executables `jps.exe` or `java.exe` are not present on PATH. Examples:

```
C:\>where jps
INFO: Could not find files for the given pattern(s)

C:\>where java
INFO: Could not find files for the given pattern(s)
```

The `jps.exe` command is part of the Java installation but it isn't added on the PATH by default on Windows and there's no indication on the SonarQube docs that it's required.

When the `where` command fails in the batch script the `%errorlevel%` is set to `1` and, even if all checks succeeded, the subroutines of the batch files still fails because when they end the `%errorlevel%` is different than 0.

**FIX:**

The proposed change simply replaces the `goto:eof` on :check_if_sonar_is_running and :set_java_exe by `exit /b 0`, so these subroutines will return the error code `0` in case of success instead of returning the error code returned by the `where` command.

**Result:**

After the changes the SonarQube server starts normally:

image

### DIFF
--- a/sonar-application/src/main/assembly/bin/windows-x86-64/StartSonar.bat
+++ b/sonar-application/src/main/assembly/bin/windows-x86-64/StartSonar.bat
@@ -63,7 +63,7 @@ goto:eof
         echo ERROR: Another instance of the SonarQube application is already running with PID %SQ_PROCESS%
         exit /b 1
     )
-    goto:eof
+    exit /b 0
 
 endlocal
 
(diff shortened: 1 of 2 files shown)


klaudio-sinani-sonarsource:
Hell @felipebz! Thank you for taking the time to contribute!

Hey @dimitris-kavvathas-sonarsource @zipeng-wu-sonarsource! I believe this one might look familiar to you, feel free to have a look when you have time 😄

dimitris-kavvathas-sonarsource:
Hi @felipebz ! Thank you for sharing your solution with us! 

I'll close this PR and cherry-pick your commit to our internal repo. Your changes will be synced on this repo soon.

PR #3308, opened 2022-11-17
Optimized build, bumped Gradle to 6.9.3 and Gradle Shadow plugin to 6.1.0

sniezek:
In this PR I optimized the Gradle build using Gradle Enterprise, which was a part of my onboarding at the Gradle company; I chose your project because I think it's valuable and I saw that it uses Gradle.

I additionally bumped Gradle to 6.9.3, which is the latest 6.x release, and bumped the Gradle Shadow plugin to 6.1.0 which is the latest version to be used with Gradle 6.x.

Here is the build scan of the `clean build` command run on the master branch (note that I excluded tests from the build, as well as one module that did not always work on my local setup due to some npm issue):
https://scans.gradle.com/s/lr34ucq3gipt4/timeline?sort=longest
and here is the build scan of the `build` command run on the master branch, right after the `clean build` that was run previously:
https://scans.gradle.com/s/h7fx3a64dsz44/timeline?sort=longest
it took 12.628s.

Here's the build scan of the `clean build` command run on this PR's branch:
https://scans.gradle.com/s/2xosfbuqxcq2y/timeline?sort=longest
and here's the scan of the `build` command run on that branch right after the `clean build` that was run previously:
https://scans.gradle.com/s/d3jhfhx2pgvxm/timeline?sort=longest
it took 3.174s.

The rework that I did allowed the `zip` and `downloadLicenses` tasks to use the [incremental build feature](https://docs.gradle.org/current/userguide/performance.html#incremental_build). See also [this](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.bundling.Zip.html#org.gradle.api.tasks.bundling.Zip:filter(java.lang.Class)) and [this](https://stackoverflow.com/questions/33248064/gradle-zip-task-replace-entire-file-content-while-being-copied/33248692#33248692). [Here](https://docs.gradle.org/current/userguide/common_caching_problems.html#avoid_changing_inputs_external_to_your_build) the `doFirst` replacement is explained.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ plugins {
   // Ordered alphabetically
   id 'com.github.hierynomus.license' version '0.15.0'
   id "com.github.hierynomus.license-report" version "0.15.0" apply false
-  id 'com.github.johnrengelman.shadow' version '5.2.0' apply false
+  id 'com.github.johnrengelman.shadow' version '6.1.0' apply false
   id 'com.google.protobuf' version '0.8.19' apply false
   id 'com.jfrog.artifactory' version '4.24.23'
   id "de.undercouch.download" version "5.0.1" apply false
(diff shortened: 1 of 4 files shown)


aurelien-poscia-sonarsource:
@sniezek thanks a lot for your contribution! Your changes definitely look very interesting to us. For technical reasons, we can't merge the PR as is. However, we are looking at it and we will probably cherry pick your commits.

Thanks again!

PR #3311, opened 2022-12-29
To pull request

margarida2000:
Hello,

I’m writing to suggest that in a next version of sonarqube, you offer an extra analysis parameter, that disables or enables (the default could be enable) the final step of execution of sonarqube, specifically, when sonarqube pushes the results to the server to be presented as a report.

I’m asking these as the company I work for has now other needs, and want to be able to apply sonarqube to a new tool designed by us, to evaluate commits from all developers of a project and extract metrics from the sonarqube analysis, that throughout the time will give us more information. I notice that with the end of the analysis, the sonarqube scanner writes down the results to a given directory, and sonarqube just reads them from there and interpret to a report structure, that then pushes to the server. But, as we already have the results locally, we don’t need the publishing the report phase, as this would carry unnecessary overhead. And that’s why we want this new feature.

It is a small change that would allow your product to be applied to even more cases.

Thanks!

### DIFF
--- a/PullRequest.txt
+++ b/PullRequest.txt
@@ -1,0 +1,12 @@
+Hello,
+
+I’m writing to suggest that in a next version of sonarqube, you offer an extra analysis parameter, that disables or
+enables the final step of execution of sonarqube, specifically, when sonarqube pushes the results to the server to
+be presented as a report.
+
+I’m asking these as the company I work for has now other needs, and want to be able to apply sonarqube to a new tool 
+designed by us, to evaluate commits from all developers of a project and extract metrics from the sonarqube analysis,
+that throughout the time will give us more information. I notice that with the end of the analysis, the sonarqube
+scanner writes down the results to a given directory, and sonarqube just reads them from there and interpret to a
+report structure, that then pushes to the server. But, as we already have the results locally, we don’t need the
+publishing the report phase, as this would carry unnecessary overhead. And that’s why we want this new feature.
--- a/sonar-scanner-engine/src/main/java/org/sonar/scanner/report/ReportPublisher.java
+++ b/sonar-scanner-engine/src/main/java/org/sonar/scanner/report/ReportPublisher.java
@@ -240,7 +240,7 @@ public class ReportPublisher implements Startable {
       LOG.info("Analysis report uploaded in " + (stopTime - startTime) + "ms");
     }
     */
-    return branchName;
+    return ""; //
   }
 
   void prepareAndDumpMetadata(String taskId) {
--- a/sonar-scanner-engine/src/main/java/org/sonar/scanner/report/ReportPublisher.java
+++ b/sonar-scanner-engine/src/main/java/org/sonar/scanner/report/ReportPublisher.java
@@ -50,7 +50,7 @@ import org.sonar.scanner.protocol.output.ScannerReportWriter;
 import org.sonar.scanner.scan.ScanProperties;
 import org.sonar.scanner.scan.branch.BranchConfiguration;
 import org.sonar.scanner.scan.branch.BranchType;
-import org.sonarqube.ws.Ce;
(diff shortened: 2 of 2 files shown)


wouter-admiraal-sonarsource:
Hi @margarida2000 ,

First: your changes are simply commenting lines, which, although it might fit your use-case, would break it for many others.

Second: pushing the report does not prevent you from achieving what you want. Perhaps there is some (very little) overhead, yes, but you will always need a running SonarQube instance anyway (because of the analyzers). And I don't see how this change will fit into our overall product strategy.

For these reasons, I will close this PR.

margarida2000:
Hello,

Firstly, thanks for replying!

But, what do you intend with "you will always need a running SonarQube instance anyway (because of the analyzers)"? I mean, if SonarLint can do a local analysis, this should not be impossible. Couldn't I previously download the analysers?

wouter-admiraal-sonarsource:
Indeed, SonarLint embeds many analyzers. But the scanner does not.

To give some more context: the scanner has a completely independent lifecycle, and we want it to be as backward-compatible as possible. To achieve this -- and to save people from having to update all their CIs every time a new version of SonarQube is released -- we build the scanner to be a "wrapper"* around the analyzers. It downloads all analyzers directly from your SonarQube server and executes them locally. This way, as soon as you update SonarQube, all your CIs immediately benefit from the latest and greatest analyzer features, without any further action on your side.

Please note that many of our analyzers are open-source. We even have some "competitors" that use our analyzers in their products ([Codacy](https://docs.codacy.com/getting-started/supported-languages-and-tools/) for example, or [CodeClimate](https://docs.codeclimate.com/docs/list-of-engines)). Depending on your use case, you could execute the analyzers directly (note this is not a use-case we support).

\* this is selling it short, as the scanner does a lot more than just execute analyzers.

margarida2000:
I do understand what you are saying about the need of connecting to SonarQube when running an instance of Sonar Scanner, so to get stuff, such as the analyzers. But these get requests are done before the analysis and what I intended with this pull request is only to ask for a way that I could turn off the pushing the results to the server phase, i.e. the step done after the analysis. Because, for my case (as I explained before, I intended to aggregate these results of SonarQube with results from other tools such as Jira), I am not interest in what is presented there, I will have my own UI. 

I'm being persistent about this issue, because until now you only said that this is a matter of (very little) overhead, but didn't say that what I'm trying to achieve is impracticable.

wouter-admiraal-sonarsource:
Sorry, let me rephrase: sending the report to SonarQube is "a little overhead". Of course, running and maintaining a SonarQube instance is a different story.

If my understanding is correct, you want to benefit from the analyzers we have, without using SonarQube or its Scanner helpers. To illustrate very roughly, this is the architecture we have:

```
SonarQube  <-> Scanner <-> Analyzers
```

My understanding is that you are only interested in the "Analyzers" part. This, you can achieve (at least, for all our open-source analyzers). You would need to download the analyzers yourself (usually directly from GitHub; example for [SonarJS](https://github.com/SonarSource/SonarJS)) and execute them individually (this is what others like CodeClimate and Coday do; to my knowledge, they don't use SonarQube or any scanners). Note that, although they are executable, this is not what they are intended for; so it will take some figuring out (even I wouldn't know how to do this; but I know it's possible, and you have access to the source code, so 🤷‍♂️ ). You could also fork one of our scanners (they're all open source; example for [Maven](https://github.com/sonarsource/sonar-scanner-maven)), and simply update it to fit your needs.

Another approach is to still run a SonarQube server but to fetch all results via its Web API. This might actually be an easier approach in the long run.

In any case, your use case is not something we would support. I don't see how it would fit into our mid-to-long-term strategy, and you can still achieve what you want with our existing products (via the Web API). Our products are mostly open source, so I do believe you can achieve what you want. But I'm not sure how practical it would be.

PR #3313, opened 2023-01-06
[NO JIRA] Fix non-compliant L10N key/value definition

jycr:
With previous declaration in `core.properits`:

```text
metric_domain.Integration Tests=Integration Tests
```

... following Java code:

```java
resourceBundle.getString("metric_domain.Integration")
```

... returns:

```text
Tests=Integration Tests
```

And following Java code:

```java
resourceBundle.getString("metric_domain.Integration Tests")
```

... throw exception:

```text
java.util.MissingResourceException: Can't find resource for bundle java.util.PropertyResourceBundle, key metric_domain.Integration Tests
```

-----

I don't know if it's the right fix in your context.

Perhaps following fix would be better:

```text
metric_domain.Integration\ Tests=Integration Tests
```

With this declaration in core.properties, following Java code:

```java
resourceBundle.getString("metric_domain.Integration Tests")
```

... returns:

```text
Integration Tests
```

### DIFF
--- a/sonar-core/src/main/resources/org/sonar/l10n/core.properties
+++ b/sonar-core/src/main/resources/org/sonar/l10n/core.properties
@@ -2230,7 +2230,7 @@ severity.INFO.description=Neither a bug nor a quality flaw. Just a finding.
 
 metric_domain.Size=Size
 metric_domain.Tests=Tests
-metric_domain.Integration Tests=Integration Tests
+metric_domain.Integration=Tests\=Integration Tests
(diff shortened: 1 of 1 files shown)


guillaume-peoch-sonarsource (commented):
Thanks again for your input @jycr ! This is very appreciated 😃 

For this case, I believe we are not using`metric_domain.Integration...` property anymore.
Si I'll therefore remove it on my side.

Cheers

PR #3322, opened 2023-03-31
BUILD-2798 Implement suggested workaround for false positive in Travis CI

elrzn:
We currently have an issue in which the `after_failure` body gets executed even after the scripts have run successfully (rc=0).

A support ticket has been opened and while they're looking into it, a workaround has been provided.

The response is quoted below:

Thank you for reaching out and sorry for the hassle.

The issue encountered has been raised with our engineering team and they are looking into it. In the meantime, I will be keeping this ticket in open status and get back to you as soon as we have more information available.

Additionally, as a workaround, can you please try adding the TRAVIS_TEST_RESULT environment variable as a check (source: https://docs.travis-ci.com/user/environment-variables/#:~:text=is%20enabled.-,TRAVIS_TEST_RESULT,-%3A%200%20if) within your .travis.yml after_success/after_failure phase, e.g.:

```
after_success:
  - "[[ $TRAVIS_TEST_RESULT = 0 ]] && echo passed"

after_failure:
  - "[[ $TRAVIS_TEST_RESULT = 1 ]] && echo failed"
```


tomverin:
@elrzn this PR should be made on sonar-enterprise as we never write directly to the sonarqube repository

PR #3323, opened 2023-04-07
Update BatchIndex.java

tisonv:
Because of unknown reason, the /lib/scanner folder was not extracted from the windows archive several times on our server. /batch/index fails without notifying the user.
The SonarQube Azure DevOps task will fail with a cryptic error : ##[error]ERROR: Failed bootstrap index response:
ERROR: Failed bootstrap index response:

### DIFF
--- a/server/sonar-webserver-webapi/src/main/java/org/sonar/server/batch/BatchIndex.java
+++ b/server/sonar-webserver-webapi/src/main/java/org/sonar/server/batch/BatchIndex.java
@@ -65,6 +65,8 @@ public class BatchIndex implements Startable {
           }
         }
       }
+    } else {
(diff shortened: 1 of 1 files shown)


tisonv:
The `prepare_fs()` in `server\sonar-webserver-webapi\src\test\java\org\sonar\server\batch\BatchIndexTest.java` prevents me from testing the exception.
Do I need to add a new test file in In `server\sonar-webserver-webapi\src\test\java\org\sonar\server\batch\` then ? What would be its name ?

Sorry not familiar at all with Java

pierre-guillot-gh:
Bump @wojciech-wajerowicz-sonarsource , can we make this move forward or drop it?

tisonv:
Sorry I couldn't find time to complete it properly.

PR #3325, opened 2023-06-13
adding updatedAt field in web api of rule

Agasar-CN:
The /rules/search interface document of the Soanr9 version shows that the updatedAt field is added to the f field, but the value cannot be obtained from the interface.
I now fill in this field to make its function consistent with the interface document.


aurelien-poscia-sonarsource:
Hi @Agasar-CN , thanks for your contribution and identifying this missing information. For us to be able to take your contribution into consideration, could you please add a corresponding test? Also, can you target the master branch?

Another option if you want to report this problem, is to do it via our [community portal](https://community.sonarsource.com/).

aurelien-poscia-sonarsource (changes requested):
i @Agasar-CN , thanks for your contribution and identifying this missing information. For us to be able to take your contribution into consideration, could you please add a corresponding test? Also, can you target the master branch?

Another option if you want to report this problem, is to do it via our [community portal](https://community.sonarsource.com/).

Agasar-CN:
I have changed the target branch to master and added a SearchActionTest unit test class.

Agasar-CN:
Do I need to make any other modifications?

PR #3330, opened 2023-06-28
Update the api documentation for `api/components/search_projects`.

mbyrnepr2:
The current [documentation](https://next.sonarqube.com/sonarqube/web_api/api/components?internal=true) for this endpoint mentions filtering on language using `language`:
_To filter on language keys use the language key:
to filter on a single language you can use 'language = java'
to filter on several languages you must use 'language IN (java, js)'_

This seems to work only when `languages` is used as in:
`/api/components/search_projects?filter=languages=py` or `/api/components/search_projects?filter=languages IN (py, js)`


zipengwu:
Hi @mbyrnepr2, thank you very much for spotting this error. I will cherry-pick your contribution into our repository. I will close this PR for now.

PR #3331, opened 2023-08-09
Allow overriding PIDFILE.

ticpu:
Currently, the PID file is written in the /bin folder, security-wise, this is not a good pratice as this directory should be read-only.

Allowing change to the PIDFILE via environment allows placing it in the correct location such as /run/sonarqube/sonarqube.pid


CloneOfAlex (approved):
Looks good to me.

zipengwu:
Thanks for raising the security concern. We decided to solve it differently by providing a PIDDIR environment viable. You can track the status [here](https://sonarsource.atlassian.net/browse/SONAR-20310).

PR #3342, opened 2023-12-12
Update logback from 1.3.11 to 1.3.14

hdeadman:
Using latest 1.3.14 will clean up a vulnerability finding that doesn't apply to SonarQube because it doesn't use the logback receiver but might as well make it go away.  https://mvnrepository.com/artifact/ch.qos.logback/logback-core/1.3.11

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -284,7 +284,7 @@ subprojects {
       dependency 'org.sonarsource.text:sonar-text-plugin:2.7.1.1388'
 
       // please keep this list alphabetically ordered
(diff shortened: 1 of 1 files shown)


aurelien-poscia-sonarsource:
Hi @hdeadman, thank you for your contribution. We will not accept it as we have scheduled a 3rd party dependency upgrade sprint next week, where we will review and update all dependencies. As part of it, we will upgrade logback and probably to version 1.4.xxx.

PR #3344, opened 2024-01-03
Introduce toolchain to compile with Java 17

KengoTODA:
Currently, Gradle buildscript forces developers to install Java 17+. This process can be handled by the [JVM toolchain](https://docs.gradle.org/7.5.1/userguide/toolchains.html), suggested in this PR, and it helps developers set up the development environment more easily.

JVM toolchain tries to find existing JDK, so this change will not affect existing development/build environments.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,8 +2,6 @@ import groovy.json.JsonOutput
 import groovy.text.SimpleTemplateEngine
 import org.sonar.build.BlackBoxTest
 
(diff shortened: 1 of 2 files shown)


matteo-mara-sonarsource:
Thanks for this PR @KengoTODA, we are considering it to be merged and we will start the cherry-picking procedure so that it gets properly propagated to the sonarqube repository.

PR #3350, opened 2024-01-24
NO-JIRA Make Gson usage more type-safe

Marcono1234:
Changes:
- Uses `Gson.fromJson(..., TypeToken<T>)` instead of `fromJson(..., Type)`
The overload with `TypeToken` argument was added in Gson 2.10 and provides more type-safety because the return type is bound by the `TypeToken`, whereas for the `fromJson(..., Type)` overloads the return type is unbounded, so you can accidentally specify incorrect types.
This is actually the case for `GsonEmail.java`, `GsonTeam.java` and `GsonGroup.java` which specified `Collection` as type to parse but actually expected a `List`. This just happened to work because Gson [creates an `ArrayList`](https://github.com/google/gson/blob/gson-parent-2.10.1/gson/src/main/java/com/google/gson/internal/ConstructorConstructor.java#L318-L322) when asked to create a `Collection`, but that is an implementation detail.
- Replaces usage of the deprecated `JsonParser()` constructor

:warning: I have not tested or built these changes locally because the Gradle build started by downloading a 600MB Elasticsearch Linux artifact (on a Windows machine), so I did not consider the effort to set up a development environment for this worth it.
Sorry for this; feel free to use the changes of this pull request and adapt them as necessary.

### DIFF
--- a/server/sonar-alm-client/src/main/java/org/sonar/alm/client/github/GithubApplicationClientImpl.java
+++ b/server/sonar-alm-client/src/main/java/org/sonar/alm/client/github/GithubApplicationClientImpl.java
@@ -73,9 +73,12 @@ public class GithubApplicationClientImpl implements GithubApplicationClient {
   protected static final String WRITE_PERMISSION_NAME = "write";
   protected static final String READ_PERMISSION_NAME = "read";
   protected static final String FAILED_TO_REQUEST_BEGIN_MSG = "Failed to request ";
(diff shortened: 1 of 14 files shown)


aurelien-poscia-sonarsource:
Hello @Marcono1234 ,

Thank you for your contribution, which we decided to keep with a few edits!

For technical reasons, we can't accept contributions to this public repository. So I've cherry-picked your contribution, did a few edits and it is now part of the SonarQube code base (your name as an author was preserved :) ). I will now close this PR.

Aurélien

PR #3356, opened 2024-02-16
SONAR-21589 Do not require 'api' scope for GitLab authentication

am97:
SonarQube 9.9.4 breaks GitLab authentication when group sync is disabled and the token has only `read_user` scope, as reported in SONAR-21589. (This used to work fine on v9.9.3)

The first commit on this PR backports 8f24b6c to branch-9.9 (otherwise the build doesn't work)

The second commit fixes the scope. It's a partial revert of c6027f9

The third commit fixes a problem introduced by 4e7453d: an API query was made even if group sync was disabled

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -69,22 +69,28 @@ allprojects {
     def artifactoryPassword = System.env.'ARTIFACTORY_PRIVATE_PASSWORD' ?: (project.hasProperty('artifactoryPassword') ? project.getProperty('artifactoryPassword') : '')
     def artifactoryUrl = System.getenv('ARTIFACTORY_URL') ?: (project.hasProperty('artifactoryUrl') ? project.getProperty('artifactoryUrl') : '')
 
-    if (artifactoryUrl == '') {
-      throw new GradleException('Invalid artifactoryUrl')
-    }
-
-    maven {
-        if (artifactoryUsername && artifactoryPassword) {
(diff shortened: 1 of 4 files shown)


aurelien-poscia-sonarsource:
See comment in https://github.com/SonarSource/sonarqube/pull/3357; once it is fixed for 10.5, we will decide whether we also want to support "read_api" for 9.9.5

behniafb:
Can you give an update, whether it is fixed for 9.9 version or not ? @aurelien-poscia-sonarsource

aurelien-poscia-sonarsource:
Hey @behniafb, yes and all details can be found [here](https://sonarsource.atlassian.net/browse/SONAR-21589)

behniafb:
> Hey @behniafb, yes and all details can be found [here](https://sonarsource.atlassian.net/browse/SONAR-21589)

Thanks. 
I've read through the issue and there was mentioned that versions prior to 10.4 were using correct api scope `read_api`.
But it's not! I'm using version 9.9 and the issue exists there too. 
I know that the solution is to update, but I wanted to know wether I'm thinking wrong or the issue description is wrong.

aurelien-poscia-sonarsource:
Hi @behniafb ,

Did you upgrade to version 9.9.5? it should be fixed there.

Cheers,
Aurélien

PR #3357, opened 2024-02-20
[master] SONAR-21589 Do not require 'api' scope for GitLab authentication

am97:
This is a port of #3356 for the `master` branch. It fixes GitLab authentication when group sync is disabled and the token has only `read_user` scope, as reported in SONAR-21589.


aurelien-poscia-sonarsource:
Hi @am97, thanks a lot for your great contribution! We will however not accept as we plan to spend time on the topic before 10.5. We would also like to not require the "api" scope, but "read_api" instead. But we also need to remain compatible with instances using an application with the "api" scope. So there is a bit more work involved.

PR #3360, opened 2024-04-27
fix: RuleCreator these missing rule template keys should display in the error message

qxo:
these missing rule template keys should display in the error message

so we can easy to fix the problem,


antoine-vinot-sonarsource:
Hi @qxo!

Thank you for your contribution and willingness to improve the product!

Your change makes sense, and it would be a nice improvement. 

I've created [a ticket](https://sonarsource.atlassian.net/browse/SONAR-22191) on our side to keep track of it, as it still needs to add a few tests and minor adjustments.

I'm closing this Pull Request for now. 

Best

PR #3361, opened 2024-05-24
Validate SonarCloud webhook signature fail was caused by the JsonWriter build request payload

ahasgx:
as a developer, i validate sonarcloud webhook signature fail was caused by the JsonWriter build request payload.

org.sonar.server.webhook.WebhookPayloadFactoryImpl#create build the WebhookPayload with the JsonWriter which in sonar-plugin-api.jar, JsonWriter  based on Gson's JsonWriter  and set the htmlSafe is true, 
so it's escape some chars if the  payload contains the chars which in the array HTML_SAFE_REPLACEMENT_CHARS, and signature the payload with the secret, then post the body by the callback url, 

But the payload I received was un-escaped so that i validate SonarCloud webhook signature always fail, (the payload is changed).

for example: payload Object has a key-value is  {"key": "key=val"},  by the JsonWriter to json, sonar server sign the payload is {"key": "key\u003dval"}, i received the payload is  {"key": "key=val"}.

in the document https://docs.sonarsource.com/sonarcloud/advanced-setup/webhooks/#securing-your-webhooks,
i think at least the JsonWriter conver to json that escape some special chars maybe validate sonarcloud webhook signature fail need to describe clearly .

### DIFF
--- a/server/sonar-server-common/src/test/java/org/sonar/server/webhook/WebhookPayloadFactoryImplTest.java
+++ b/server/sonar-server-common/src/test/java/org/sonar/server/webhook/WebhookPayloadFactoryImplTest.java
@@ -20,6 +20,9 @@
 package org.sonar.server.webhook;
 
 import com.google.common.collect.ImmutableMap;
+
+import java.io.StringWriter;
+import java.io.Writer;
 import java.util.Map;
 import javax.annotation.Nullable;
 import org.junit.Before;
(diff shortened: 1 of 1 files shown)


lukasz-jarocki-sonarsource:
Hi there,

Thanks for spotting the issue and this contribution! Before we proceed, could you clarify whether the issue you are having is with SonarCloud or SonarQube? Because you mention SonarCloud in your description but this pull request targets SonarQube product.

ahasgx:
> Hi there,
> 
> Thanks for spotting the issue and this contribution! Before we proceed, could you clarify whether the issue you are having is with SonarCloud or SonarQube? Because you mention SonarCloud in your description but this pull request targets SonarQube product.

Sorry, I thought it was the same organization.

ahasgx:
ignore

PR #3368, opened 2024-10-23
Fix build cache issues and improve task configuration

mondrejko:
- Identified and corrected tasks missing input and output values, improving cache efficiency.
- Added missing path sensitivity configuration to prevent cache misses from differences in absolute file paths.
- Added explicit log messages for tasks that are uncacheable.
- Resolved overlapping task outputs issue: generatePackageInfo task now outputs to a build directory, with a finalizer task copying results to the source directory.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -521,47 +521,69 @@ subprojects {
   }
 
   task generatePackageInfo {
-    doLast {
-      def allPathsContainingJavaFiles = [] as Set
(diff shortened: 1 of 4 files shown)


lukasz-jarocki-sonarsource:
Apologies for the delay but your commit has been approved and merged to the upstream repository. It will be visible in the public repository in 24 hours.

Parts of the commit had to be removed as they were not compatible with our private code.

PR #3372, opened 2024-11-06
Release action - dry run

lukasz-jarocki-sonarsource:
1. This is the first step. To just run it with the same settings as from sonar-enterprise with a dry run.
2. The second step will be to add releasability checks - I understand this is optional as we said that "we will figure out the Jira process later" i.e.
3. The next step will be to run it with appropriate versioning with a dry run.
4. And the next step (In December) - to remove dry run.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,19 @@
+name: Release
+# This workflow is triggered when publishing a GitHub release
+# yamllint disable-line rule:truthy
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  release:
+    permissions:
+      id-token: write
+      contents: write
+    uses: SonarSource/gh-action_release/.github/workflows/main.yaml@v5
+    with:
+      slackChannel: squad-sq-platform-build-notification
+      publishToBinaries: true
+      mavenCentralSync: true
+      dryRun: true


.github/workflows/release.yml
```diff
@@ -0,0 +1,19 @@
+name: Release
+# This workflow is triggered when publishing a GitHub release
+# yamllint disable-line rule:truthy
```
sebastienvermeille, line 3: not sure you need this comment if you are not making use of validation tools such as pre-commit or any yaml linters (up to you)

sebastienvermeille (approved):
LGTM

alain-kermis-sonarsource (approved):
👍

.github/workflows/release.yml
lukasz-jarocki-sonarsource: Ok I will remove it

lukasz-jarocki-sonarsource:
Let me run the public git sync locally to see if something will break

lukasz-jarocki-sonarsource:
There is no easy way to run the synchronization script locally with local (not remote) repositories as arguments. And using local repositories is required to test it.
From my investigation it looks like it should work, as long as the corresponding commit is on the public_master https://github.com/SonarSource/sonar-enterprise/pull/12232/files

Otherwise very likely we will get the message 
`Error:  Latest reference to public master (4232221a1a65039dee2e8c5bd2288231e5250fe8) is not HEAD of branch public_master. Previous run of synchonization script left an inconsistent state
`

matteo-mara-sonarsource:
Merging to the public repo after analyzing all the pros and cons.

No synchronization issues are expected; the GitHub workflows folder is not synchronized from sonar-enterprise.

lukasz-jarocki-sonarsource:
It turns out after another talk with dev-infra is that we can't release from public repository.

PR #3377, opened 2024-12-24
DO NOT MERGE Adding shibboleth for OpenSaml dependency

matteo-mara-sonarsource:


lukasz-jarocki-sonarsource:
What is the status of this PR? It is approved, let's merge?

PR #3378, opened 2025-01-08
NOJIRA: Fix CVE-2024-52316 for LTS version

marvin-w:
Part of 

Closes https://nvd.nist.gov/vuln/detail/CVE-2024-52317

Newer versions are not affected as far as I can tell but please cross-check as well.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -325,7 +325,7 @@ subprojects {
         entry 'log4j-api'
         entry 'log4j-to-slf4j'
       }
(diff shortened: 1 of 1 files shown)


marvin-w:
When can we get this out? It's security related...

matteo-mara-sonarsource:
Thanks a lot @marvin-w for this contribution.
I can confirm that SonarQube is not affected by this CVE since we don't use custom Jakarta Authentication. 
That is the reason why we did not patched tomcat to 9.0.96 when the CVE was raised on the codebase by our tools.

PR #3380, opened 2025-02-18
Make sonar startup script insensitive to the terminal width

necto:
Surprizingly, even when `ps` output is piped into another command, such as `grep` in this case, it is still truncated based on the terminal width. Depending on how wide your terminal you get different results from running the same command: 
- Wide-enough terminal:
```
./sonar.sh start
/usr/bin/java
Starting SonarQube...
Started SonarQube.
```
- Not-wide-enough terminal (width = 263 columns):
```
./sonar.sh start
/usr/bin/java
Starting SonarQube...
Removed stale pid file: ./SonarQube.pid
Failed to start SonarQube.
```

To fix that, I propose to run `ps` with `-ww` option, see the [man page](https://man7.org/linux/man-pages/man1/ps.1.html) Tested on `ps` version 4.0.3 and 3.3.17


pierre-guillot-gh:
Hey, thanks for having a look at this. Please open your PR on the private repo. Also, the SQ team does not exist anymore, you can find the owner by looking at the CODEOWNER file.

PR #3383, opened 2025-03-23
Pollute the CI platfrom cache with arbitrary data

ericmorand:
Would you approve this PR checks, it would likely pollute the Cirrus CI cache with some arbitrary data.

I don't recommend it.


david-cho-lerat-sonarsource:
> Would you approve this PR checks, it would likely pollute the Cirrus CI cache with some arbitrary data.
> 
> I don't recommend it.

A little early for April Fools ;)

PR #3384, opened 2025-04-24
fix: fix logging for missing deprecated rules

iTrooz:
Hello ! This PR is a small fix for the following incorrectly formatted string

I am not familiar with Sonarqube internals, so I don't know if `SingleDeprecatedRuleKey::getOldRuleKeyAsRuleKey()` was the string you originally wanted to output. Let me know if I should change it

### DIFF
--- a/server/sonar-webserver-core/src/main/java/org/sonar/server/rule/registration/RulesRegistrationContext.java
+++ b/server/sonar-webserver-core/src/main/java/org/sonar/server/rule/registration/RulesRegistrationContext.java
@@ -83,9 +83,11 @@ class RulesRegistrationContext {
       String ruleUuid = entry.getKey();
       RuleDto rule = dbRulesByRuleUuid.get(ruleUuid);
       if (rule == null) {
-        LOG.warn("Could not retrieve rule with uuid %s referenced by a deprecated rule key. " +
-          "The following deprecated rule keys seem to be referencing a non-existing rule",
-          ruleUuid, entry.getValue());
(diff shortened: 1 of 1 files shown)


henryju:
Hi @iTrooz 

I agree there is a obvious mistake. The problem is I can't merge your contribution without the addition of a test case, since this branch is unfortunately not covered.
https://next.sonarqube.com/sonarqube/code?id=sonarqube&selected=sonarqube%3Aserver%2Fsonar-webserver-core%2Fsrc%2Fmain%2Fjava%2Forg%2Fsonar%2Fserver%2Frule%2Fregistration%2FRulesRegistrationContext.java&line=86

If you have the time to work on adding a test (possibly in `RulesRegistrantIT` but not sure), that would be much appreciated. Else let me know and I will simply create a ticket in our backlog to be tackled during the next hardening sprint.

iTrooz:
Hey ! Sorry, I don't have enough understanding about the codebase to add a test. Specifically, when the deprecated key is processed, I have no idea how to test [that a rule was not added in rulesByKey](https://github.com/SonarSource/sonarqube/blob/f726ef07e7e1836624489befab56ae7c59c3040d/server/sonar-webserver-core/src/main/java/org/sonar/server/rule/registration/RulesRegistrationContext.java#L90). So I could write a test that runs the code and add coverage, but it would not really check anything

It's better if I let you handle this

henryju:
I have created a [ticket](https://sonarsource.atlassian.net/browse/SONAR-24944) to be picked during the next hardening sprint. I am closing this PR as we will do the fix upstream.

